PR #4, opened 2022-01-26
feat(pagerduty): Move incidents to different escalation policy [OSD-9777]

georgettica:
this also
- adds README's for everything
- adds Makefile for testing and building

currently just wrote the testing framework for this and the first command in the list

next steps are
- [x] add the actual API
- [x] fill in the types with what is minimally needed for this action
- [x] use PD creds to run a test on this code
- [x] update README with changes
- [x] ? add Makefile / automated way to test this ?


georgettica:
related to OSD-9777

pkg/pagerduty/pagerduty_test.go
```diff
@@ -0,0 +1,151 @@
...
+// http.DefaultClient, to avoid other packages tweaks to http.DefaultClient
+// causing issues with our HTTP calls. This also allows us to tweak the
+// transport values to be more resilient without making changes to the
+// http.DefaultClient.
+//
+// Keep this unexported so consumers of the package can't make changes to it.
+//var defaultHTTPClient HTTPClient = newDefaultHTTPClient()
+
+//
+// func testMethod(t *testing.T, r *http.Request, want string) {
```
Tessg22, line 145: Can we remove the unused function?

pkg/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil
+	}
+	incident.EscalationPolicy = escalationPolicy
+
+	o := []sdk.ManageIncidentsOptions{
+		{
+			ID: incident.ID,
+			EscalationPolicy: &sdk.APIReference{
+				Type: "escalation_policy_reference",
+				ID:   incident.EscalationPolicy.ID,
```
Tessg22, line 73: Doesn't `incident.EscalationPolicy = escalationPolicy` on the line 66 do the same thing?

pkg/pagerduty/pagerduty_test.go
georgettica: Yes, I'll add a comment on where I pulled them from

pkg/pagerduty/pagerduty.go
georgettica: Nope, line 66 populates the strict that is used here, they are not connected

Tessg22: Right!

Tessg22:
Otherwise lgtm. Should we get one more eyes of someone who didn't touch the code?

georgettica:
Nope, we can do that in the next feature

pkg/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,100 @@
...
+		perr := sdk.APIError{}
+		if errors.As(err, &perr) {
+			switch perr.StatusCode {
+			case http.StatusUnauthorized:
+				return InvalidTokenErr{Err: err}
+			case http.StatusBadRequest:
+				isErrAnInvalidInputErr := perr.APIError.Valid &&
+					perr.APIError.ErrorObject.Code == InvalidInputParamsErrorCode
+				if isErrAnInvalidInputErr {
+					return InvalidInputParamsErr{Err: err}
```
ramonbutter, line 92: A 400 is always a client-side bad request. Do we really need to create the boolean isErrAnInvalidInputErr? On any http.StatusBadRequest we shouldn't retry, so the logic is the same whether the InputParameter are invalid or not

pkg/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,100 @@
...
+			},
+		},
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), pagerDutyTimeout)
+	defer cancel()
+	_, err := p.c.ManageIncidentsWithContext(ctx, p.userEmail, o)
+
+	if err != nil {
+		perr := sdk.APIError{}
```
ramonbutter, line 83: better put this error handling into its own function, since it is going to be used by other client interactions with PD, no?

pkg/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,100 @@
...
+	resp := PagerDuty{
+		c:         client,
+		userEmail: user.Email,
+	}
+
+	return resp, nil
+}
+
+// MoveToEscalationPolicy will move the alerts EscalationPolicy to the new EscalationPolicy
+func (p PagerDuty) MoveToEscalationPolicy(incident Incident, escalationPolicy EscalationPolicy) error {
```
ramonbutter, line 62: Why do we need this again?

pkg/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,100 @@
...
+
+// New will create a PagerDuty struct with all of the required fields
+func New(client *sdk.Client) (PagerDuty, error) {
+	opts := sdk.GetCurrentUserOptions{
+		Includes: []string{},
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), pagerDutyTimeout)
+	defer cancel()
+	user, err := client.GetCurrentUserWithContext(ctx, opts)
+	if err != nil {
```
ramonbutter, line 42: put this into an own function.

pkg/pagerduty/pagerduty.go
```diff
@@ -0,0 +1,100 @@
...
+	"net/http"
+	"time"
+
+	pkgerrors "github.com/pkg/errors"
+
+	sdk "github.com/PagerDuty/go-pagerduty"
+)
+
+const (
+	InvalidInputParamsErrorCode = 2001
```
ramonbutter, line 15: I am not a big fan of redefining magic constants that are used by other code. Can we not import this? Or do we really need this?

pkg/pagerduty/types.go
```diff
@@ -0,0 +1,15 @@
+package pagerduty
+
+// Incident represents a PagerDuty Incident
+type Incident struct {
```
ramonbutter, line 4: Do we really want to redefine the pagerduty-go types? For me it doesn't make any sense.
We can just pick it up from here: https://github.com/PagerDuty/go-pagerduty/blob/433447fbcc6577b0faec04b3c613d04e51763ad4/incident.go#L75
As we will need to maintain the duplicate definition.
Wrapping the communication to Pagerduty makes sense to me, as we are going to mock the functions for testing. But that doesn't apply to the data types or? 
However, if I am thinking about it now even wrapping the pagerduty-go library doesn't make too much sense IMO. Just defining an interface of the PD SDK in our code would simply also do its job and we would be able to mock and catch any call that is done to the library during runtime.

pkg/pagerduty/pagerduty.go
georgettica: I am not retrying, it's just so I can wrap the PD APIError with something I can pass to the user && validate.

extracting to a variable was to try and make the if check smaller

pkg/pagerduty/pagerduty.go
georgettica: I think this is premature optimization, the next one along can do that :)

pkg/pagerduty/pagerduty.go
georgettica: this is what the card required, didn't think too much on why though

pkg/pagerduty/pagerduty.go
georgettica: that sounds right, will do

pkg/pagerduty/pagerduty.go
georgettica: this is the error code that was pulled from the PD api, I can explain where I got it if that helps

pkg/pagerduty/types.go
georgettica: I think that working this way ensures that ANY change we want to make, if it's in the datastructure or the api calls can happen without changing the code outside this function

pkg/pagerduty/pagerduty.go
ramonbutter: sure, I was just wondering why you are untangling this specific error when there is no different logic in place to handle this case. Especially as you return [here]( https://github.com/openshift/configuration-anomaly-detection/pull/4/files/73105185ca69ce1667ab3caca50dd82effad072d#diff-ff48eb69c8e0cf64f068d910259cb2b2aa429edae92bde70980c64d2e290a85fR96) a  generic err object ... But sure at the moment, the result is not used anyways.

pkg/pagerduty/pagerduty.go
ramonbutter: kk , ah maybe for the case its SRE actionable we move the incident from automation escalation policy immediately to SRE escalation policy

pkg/pagerduty/pagerduty.go
ramonbutter: yes at least a link or so added as a comment sounds good to me

pkg/pagerduty/types.go
ramonbutter: got it: For the case, we update the SDK we can simply change our wrapper function. However, not sure whether it is worth it for the data structures. In gcp-project-operator, we also simply used the SDK types.

georgettica:
not merging as now I don't have a CLI I can check against, so the last commit is untested

pkg/pagerduty/pagerduty.go
georgettica: it makes sense if the CLI wants to do something prettier on the error, and it makes testing easier (I don't need to know in the test the SDK types.

this makes the test be more or less what the library user will take

pkg/pagerduty/pagerduty.go
georgettica: oh, that might make sense

pkg/pagerduty/pagerduty.go
georgettica: I cannot provide a link as I found the error during running the command manually. and the API page doesn't note this at all

pkg/pagerduty/types.go
georgettica: I know gcp-project-operator doesn't do that. but by doing it my way the code outside the pagerduty package (even the tests) doesn't use the sdk types (aside of creating the client)

pkg/pagerduty/pagerduty.go
Tessg22: [https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTYz-errors](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTYz-errors) search for code '2001'

pkg/pagerduty/pagerduty.go
georgettica: done @Tessg22

Tessg22: Where? I don't see the file updated

PR #5, opened 2022-02-02
feat(ack): add new endpoint [OSD-9777]

georgettica:
- extract common functions into private functions
- make the AcknowledgeIncident endpoint smaller
- reduce the Incident struct size as it was not necessary


georgettica:
related to OSD-9777

PR #8, opened 2022-02-02
feat(assign): add new endpoint

georgettica:
- add more test targets
- add pagerduty-cli usage in `pkg/README.md`
- remove github.com/pkg/errors in favor of `fmt`
- reduce testing responses so it's cleaner
- testing now uses the same name `Should` for the expect commands

related to [OSD-9777](https://issues.redhat.com/browse/OSD-9777)


pkg/pagerduty/pagerduty.go
```diff
@@ -62,7 +62,38 @@ func (p PagerDuty) MoveToEscalationPolicy(incident Incident, escalationPolicy Es
...
+// AssignToCurrentUser will assign the incident to the current logged in user
+func (p PagerDuty) AssignToCurrentUser(incident Incident) error {
+	err := p.AssignToUser(incident, p.currentUser)
+	if err != nil {
+		return fmt.Errorf("could not assign to the current user: %w", err)
+	}
+	return nil
+}
+
+// AssignToUser will assign the incident to the curr
```
georgettica, line 79:
```suggestion
// AssignToUser will assign the incident to the current user
```

pkg/pagerduty/pagerduty_test.go
```diff
@@ -97,6 +100,112 @@ var _ = Describe("Pagerduty", func() {
...
+				//Arrange
+				mux.HandleFunc("/incidents", func(w http.ResponseWriter, r *http.Request) {
+					Expect(r.Method).Should(Equal("PUT"))
+					_, err := w.Write([]byte(`{}`))
+					Expect(err).ShouldNot(HaveOccurred())
+				})
+				// Act
+				err := p.AssignToUser(incident, user)
+				// Assert
+				// Assert
```
georgettica, line 157:
```suggestion
```

pkg/README.md
```diff
@@ -8,13 +8,40 @@ this will hold all of the integrations / code that is to be used by the `cadctl`
...
+```shell
+INCIDENT_ID= # the ID pulled from the create command
+pd incident:resolve --ids ${INCIDENT_ID}
+```
+
 ## Development
 
 add a new endpoint to cadctl
 ```
 make bin/cobra
```
ramonbutter, line 42: this really works everywhere?

pkg/pagerduty/pagerduty.go
```diff
@@ -62,7 +62,38 @@ func (p PagerDuty) MoveToEscalationPolicy(incident Incident, escalationPolicy Es
 	}
 	err := p.manageIncident(o)
 	if errors.Is(err, FailedToUpdateIncidentError{}) {
-		return pkgerrors.Wrap(err, "could not update the escalation policy")
+		return fmt.Errorf("could not update the escalation policy: %w", err)
+	}
+	return err
```
ramonbutter, line 67: wouldn't you want to add "could not update the escalation policy on any error" for the case an error is returned by manageIncident. Or to ask differently, is there a case where an error is returned but the update was nevertheless successful?
```suggestion
	err := p.manageIncident(o)
	if err != nil {
		return fmt.Errorf("could not update the escalation policy: %w", err)
	}
	return nil
```

ramonbutter: we could maybe think about the following function, just to save some lines:

```suggestion
	return onErrorAddDetails(err, "could not update the escalation policy:")
```

and the onErrorAddDetails would return nil for the case err == nil

pkg/pagerduty/pagerduty.go
```diff
@@ -62,7 +62,38 @@ func (p PagerDuty) MoveToEscalationPolicy(incident Incident, escalationPolicy Es
...
+					Assignee: sdk.APIObject{
+						Type: "user_reference",
+						ID:   user.ID,
+					},
+				},
+			},
+		},
+	}
+	err := p.manageIncident(o)
+	if errors.Is(err, FailedToUpdateIncidentError{}) {
```
ramonbutter, line 95: again like above, I would probably keep it generic with:


```suggestion
	if err != nil {
```
As InvalidTokenError and InvalidInputParameterError have the same effect, meaning `"could not assign to user"` And you should also state that in their error messages.

pkg/pagerduty/pagerduty.go
```diff
@@ -77,7 +108,7 @@ func (p PagerDuty) AcknowledgeIncident(incident Incident) error {
 	}
 	err := p.manageIncident(o)
 	if errors.Is(err, FailedToUpdateIncidentError{}) {
```
ramonbutter, line 110: same as above

pkg/pagerduty/pagerduty.go
```diff
@@ -62,7 +62,38 @@ func (p PagerDuty) MoveToEscalationPolicy(incident Incident, escalationPolicy Es
...
-		return pkgerrors.Wrap(err, "could not update the escalation policy")
+		return fmt.Errorf("could not update the escalation policy: %w", err)
+	}
+	return err
+}
+
+// AssignToCurrentUser will assign the incident to the current logged in user
+func (p PagerDuty) AssignToCurrentUser(incident Incident) error {
```
ramonbutter, line 71: not sure right now, but isnt the preferred way to write?:

`func (p *Pagerduty)`

pkg/README.md
georgettica: Yes, I used it already 2 times

pkg/pagerduty/pagerduty.go
georgettica: No, we use interfaces only if we modify the data.

I always use the non pointer of I can help it

pkg/README.md
georgettica: @ramonbutter I didn't see what like this was referring to before.

anyways you can't run a makefile from a different folder

pkg/pagerduty/pagerduty.go
georgettica: no need to wrap this, saw your reasoning and raised it.
if we need to handle the `UnknownUpdateIncidentError`, it can be easily done

pkg/pagerduty/pagerduty.go
```diff
@@ -36,14 +35,15 @@ func NewWithToken(authToken string) (PagerDuty, error) {
...
-		return PagerDuty{}, pkgerrors.Wrap(err, "could not retrieve the current user")
+		return PagerDuty{}, fmt.Errorf("could not retrieve the current user: %w", err)
 	}
 
+	user := toLocalUser(sdkUser)
 	resp := PagerDuty{
-		c:         client,
-		userEmail: user.Email,
+		c:           client,
+		currentUser: user,
```
georgettica, line 46: @ramonbutter this is the one-way conversion I have as of now. is that fine for now?

pkg/README.md
georgettica: this assumes you are in the root of the git folder, will change if this is unclear

ramonbutter: yea somehow I thought this makefile command was pointing to a path.
I tried it on my computer as well, but am receiving:
```
make bin/cobra                    
GOBIN=/home/ramon/git/configuration-anomaly-detection/bin go install github.com/spf13/cobra/cobra@v1.3.0
package github.com/spf13/cobra/cobra@v1.3.0: can only use path@version syntax with 'go get'
make: *** [Makefile:3: bin/cobra] Error 1

```

pkg/pagerduty/pagerduty.go
ramonbutter: it's ok, you can define any data type you want. I always wanted to write as little code as possible. But I also see that in some cases there is also a gain in redefining things with limited scope to decouple the business logic from the external SDK.

pkg/README.md
georgettica: what is your `go version`?

ramonbutter: go version go1.15.15 linux/amd64

georgettica: this doesn't reproduce on my machine :(

ramonbutter: nvm

georgettica: now it does. if you run the `make bin/cobra` target again and it should warn you

georgettica:
/approve
/lgtm
?

required-version.go
```diff
@@ -0,0 +1,8 @@
+//go:build !go1.16
```
rafael-azevedo, line 1: This information can be put into the go.mod to stop building unless it has the correct version of go.

pkg/pagerduty/pagerduty.go
```diff
@@ -61,10 +61,41 @@ func (p PagerDuty) MoveToEscalationPolicy(incident Incident, escalationPolicy Es
...
+}
+
+// AssignToUser will assign the incident to the curr
+func (p PagerDuty) AssignToUser(incident Incident, user User) error {
+	o := []sdk.ManageIncidentsOptions{
+		{
+			ID: incident.ID,
+			Assignments: []sdk.Assignee{
+				{
+					Assignee: sdk.APIObject{
```
rafael-azevedo, line 86: This formatting could be a little more compact so we.

Makefile
```diff
@@ -18,9 +20,27 @@ build:
...
 
+.PHONY: cadctl-install-local-force
+cadctl-install-local-force: 
+	rm cadctl/cadctl >/dev/null 2>&1  || true
+	make cadctl-install-local
+
+
+# this required-version.go checks that you are at least at go1.16
+# being at go1.16 allows you to go install
+.PHONY: verify-gobin
```
rafael-azevedo, line 38: If we have this in the go.mod we no longer need this command.

georgettica: We do, that was the solutionfor @ramonbutter's issue

required-version.go
georgettica: That is true, I'll test it when I'm back

pkg/pagerduty/pagerduty.go
georgettica: Yeah, but this will hide some stuff?
I'll try and compact it

pkg/pagerduty/pagerduty.go
```diff
@@ -23,8 +22,8 @@ const (
 type PagerDuty struct {
 	// c is the PagerDuty client
 	c *sdk.Client
-	// userEmail is the user that will run the commands
-	userEmail string
+	// currentUser is the current logged in user
```
rafael-azevedo, line 25: We don't need the currentUser. We will always be passing in Service Account Credentials

pkg/pagerduty/pagerduty.go
```diff
@@ -36,14 +35,15 @@ func NewWithToken(authToken string) (PagerDuty, error) {
...
 // New will create a PagerDuty struct with all of the required fields
 func New(client *sdk.Client) (PagerDuty, error) {
-	user, err := getCurrentUser(client)
+	sdkUser, err := getCurrentUser(client)
 	if err != nil {
-		return PagerDuty{}, pkgerrors.Wrap(err, "could not retrieve the current user")
+		return PagerDuty{}, fmt.Errorf("could not create a new client: %w", err)
 	}
 
+	user := toLocalUser(sdkUser)
```
rafael-azevedo, line 43: we don't need the local user anymore

pkg/pagerduty/pagerduty.go
```diff
@@ -61,10 +61,37 @@ func (p PagerDuty) MoveToEscalationPolicy(incident Incident, escalationPolicy Es
...
+func (p PagerDuty) AssignToCurrentUser(incident Incident) error {
+	err := p.AssignToUser(incident, p.currentUser)
+	if err != nil {
+		return fmt.Errorf("could not assign to the current user: %w", err)
+	}
+	return nil
+}
+
+// AssignToUser will assign the incident to the curr
+func (p PagerDuty) AssignToUser(incident Incident, user User) error {
```
rafael-azevedo, line 80: func (p PagerDuty) AssignToUser(incidentID ,userID, userEmail) error {}

pkg/pagerduty/pagerduty.go
```diff
@@ -88,7 +115,7 @@ func (p PagerDuty) AcknowledgeIncident(incident Incident) error {
 func (p PagerDuty) manageIncident(o []sdk.ManageIncidentsOptions) error {
 	ctx, cancel := context.WithTimeout(context.Background(), pagerDutyTimeout)
 	defer cancel()
-	_, err := p.c.ManageIncidentsWithContext(ctx, p.userEmail, o)
+	_, err := p.c.ManageIncidentsWithContext(ctx, p.currentUser.Email, o)
```
rafael-azevedo, line 118: since we no longer will have current user we can actually just user the p.UserEmail from the sdkClient.

pkg/pagerduty/pagerduty.go
```diff
@@ -124,7 +151,16 @@ func getCurrentUser(client *sdk.Client) (*sdk.User, error) {
...
 
-		return nil, pkgerrors.Wrap(err, "could not retrieve the current user")
+		return nil, fmt.Errorf("could not retrieve the current user: %w", err)
 	}
 	return user, nil
 }
+
+// toLacalUser trims down the User struct to only the fields we are actually using, and is internal so other packages won't follow this tool and use it
+// this is not in the types.go file as that shows the clean interface, and I don't want the `go-pagerduty` import to leak out of `pagerduty.go` and `pagerduty_test.go`
+func toLocalUser(user *sdk.User) User {
```
rafael-azevedo, line 161: No longer need this function.

pkg/pagerduty/types.go
```diff
@@ -11,3 +11,11 @@ type EscalationPolicy struct {
 	// ID is the unique ID of the EscalationPolicy
 	ID string
 }
+
```
rafael-azevedo, line 14: We will no longer need the types as we discussed.

Makefile
```diff
@@ -1,9 +1,11 @@
 # Binaries used in Makefile
-bin/cobra:
+GOENV := go
```
rafael-azevedo, line 2: you don't need the go env

Makefile
```diff
@@ -1,9 +1,11 @@
...
+
+bin/cobra: verify-gobin
 	GOBIN=$(PWD)/bin go install github.com/spf13/cobra/cobra@v1.3.0
 
-cadctl/cadctl: cadctl/**/*.go pkg/**/*.go go.mod go.sum
-	GOBIN=$(PWD)/cadctl go install $(PWD)/cadctl
+cadctl/cadctl: cadctl/**/*.go pkg/**/*.go go.mod go.sum *.go
+	GOBIN=$(PWD)/cadctl $(GOENV) install $(PWD)/cadctl
```
rafael-azevedo, line 8: go bin

PR #9, opened 2022-02-09
feat(note): add new endpoint

georgettica:
- change `cluster-missing` endpoint to include basic PD token stuff
  (rewritten 4 times already)
- errors were renamed/data changed for easier use
- makefile change to fix a build error
- moved to use `fmt.Fprint` instead of `w.Write` as it's cleaner and
   doesn't require to cast to a byte array
- refactored the common errors together

related to [OSD-9777](https://issues.redhat.com/browse/OSD-9777)

### DIFF
--- a/pkg/pagerduty/pagerduty.go
+++ b/pkg/pagerduty/pagerduty.go
@@ -99,6 +99,34 @@ func (p PagerDuty) AcknowledgeIncident(incidentID string) error {
 	return nil
 }
 
+// AddNote will add a note to an incident
+func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), pagerDutyTimeout)
+	defer cancel()
+	sdkNote := sdk.IncidentNote{
+		Content: noteContent,
+	}
+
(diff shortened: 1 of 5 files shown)


pkg/pagerduty/pagerduty.go
```diff
@@ -99,6 +99,34 @@ func (p PagerDuty) AcknowledgeIncident(incidentID string) error {
...
+		if errors.As(err, &sdkErr) {
+			err := commonErrorHandling(err, sdkErr)
+			if err != nil {
+				return err
+			}
+			switch sdkErr.StatusCode {
+			case http.StatusNotFound:
+				// this case can happen if the incidentID is not a valid incident (like a number prepended with zeroes)
+				return IncidentNotFoundErr{Err: err}
+			}
```
ramonbutter, line 123:
```suggestion
			if sdkErr.StatusCode == http.StatusNotFound {
				// this case can happen if the incidentID is not a valid incident (like a number prepended with zeroes)
				return IncidentNotFoundErr{Err: err}
			}
```

georgettica: Yeah, makes sense

georgettica:
/hold
I need to change this and commit

georgettica:
@ramonbutter another lgtm?

PR #10, opened 2022-02-14
[WIP] added cloudclient and ocmclient

ramonbutter:


pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+const (
+	// the ClusterDeployment label key we're expecting to be able to pull the ClusterID from
+	cdClusterIDLabelKey string = "api.openshift.com/id"
+)
+
+// CloudClient is the client used for handling AWS Cloud Handler requests
+type CloudClient struct {
+	// clusterDeployment is the ClusterDeployment for the target cluster (the one
+	// for which a login url is requested)
+	clusterDeployment *hivev1.ClusterDeployment
```
georgettica, line 36: Do we have to use the CD directly and not take what we exactly need?

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// awsClient is the awsClient that has the BACKPLANE user credentials.  This client
```
georgettica, line 38: It says backplane here, maybe a copy-paste thing

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+	// is used to create other clients in order to perform the necessary role chaining to
+	// access a customer's account via STS assume roles.
+	awsClient awsclient.AwsClient
+
+	// ocmClient is a client that can query OCM in order to get the AccountClaim
+	ocmClient ocm.Client
+
+	// awsClientBuilder is responsible for building the aws client in order to
+	// allow mocking during testing
+	awsClientBuilder awsclient.BuilderIface
```
georgettica, line 48: I'm not sure builderIface is the best name for this interface

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// region is the region in which we are building sts clients
+	region string
+}
+
+// NewCloudClient makes a new AWS CloudClient based on the ClusterDeployment and
+// handles initial configuration building for a single request.
+func NewCloudClient(cloudConfig *interface{}, ocmClient ocm.Client, cd *hivev1.ClusterDeployment) (*CloudClient, error) {
```
georgettica, line 56: How is the cloudConfig used?

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+	return cuCreds, nil
+}
+
+func (c *CloudClient) validateAccountClaim(ac *awsv1alpha1.AccountClaim) error {
+	supportRoleArn := ac.Spec.SupportRoleARN
+	if supportRoleArn == "" {
+		// if the supportRoleARN is not set, then we won't know which role inside of the customer
+		// AWS account to assume into. This is defined by the customer for STS clusters, and defined
+		// by the aws-account-operator on CCS and OSD accounts
+		return errors.Wrap(fmt.Errorf("AccountClaim is invalid"), "supportRoleARN is not present in the AccountClaim")
```
georgettica, line 193: Change this to fmt.Errorf

This way we can remove the `errors` package and not alias `ccerrors`

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+func (c *CloudClient) AssumeSupportRole() (*sts.Credentials, error) {
+
+	clusterID, err := extractClusterID(c.clusterDeployment)
+	if err != nil {
+		glog.Errorf("Could not extract ClusterID from ClusterDeployment: %s", err.Error())
+		return nil, err
+	}
+
+	glog.V(0).Infof("Getting AccountClaim for %s", clusterID)
+	ac, err := c.ocmClient.GetAWSAccountClaim(clusterID)
```
georgettica, line 139: If using glog.v(0) the v(0) can be removed

pkg/ocm/ocm_client.go
```diff
@@ -0,0 +1,87 @@
...
+	GetClusterDeployment(clusterID string) (*hivev1.ClusterDeployment, error)
+}
+
+type client struct {
+	Conn   *sdk.Connection
+	logger *sdk.GlogLogger
+}
+
+func NewOcmClient(ocmConfigFile string) (Client, error) {
+	if ocmconfig := os.Getenv("OCM_CONFIG"); ocmconfig == "" {
```
georgettica, line 24: Didn't check, but this means that sending an empty string and not having an envvar forces the env to be empty?

georgettica (commented):
This is not a full pass on the code but for me the biggest thing is this is not tested and tests give me confidence to add features above this quickly

pkg/cloudclient/aws/aws.go
ramonbutter: It's not used yet, it still uses hardcoded values, like the jumpRole, defaultRegion and credentialPosition. That's basically a placeholder. I mean at the moment it works on my machine.
I wanted to put the configuration injection into an own story. And rather try to test it like it is. Just to split the work a bit

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
+	awsclient "github.com/openshift/configuration-anomaly-detection/pkg/cloudclient/aws/client"
+	ccerrors "github.com/openshift/configuration-anomaly-detection/pkg/cloudclient/errors"
+	"github.com/openshift/configuration-anomaly-detection/pkg/ocm"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/pkg/errors"
+)
+
+const (
+	jumpRole      = "ba"
```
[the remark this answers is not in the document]
ramonbutter: no, it's a placeholder until we are able to get the configuration from a confimaps.

pkg/cloudclient/aws/aws.go
ramonbutter: good point, I will try to define a ClusterDeployment Type in the ocm-client with only the values that are required.

pkg/cloudclient/aws/aws.go
ramonbutter: Yep, probably it's a ClientFactory or so is a better name.

pkg/cloudclient/aws/aws.go
ramonbutter: I will adjust the log level for each statement later on.

pkg/ocm/ocm_client.go
ramonbutter: For the case OCM_CONFIG is not set, it will set the envVariable OCM_CONFIG with the configuration position that was specified in the function argument. And yes if you don't provide a configPath, and don't have the env set will leave everything empty, and the connection will fail

georgettica: Then should the error also catch that case?

cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -38,6 +41,35 @@ var ClusterMissingCmd = &cobra.Command{
 			return fmt.Errorf("could not start client: %w", err)
 		}
 
+		// ---------------------------------------
+		// will be fetched from PD
+		clusterID := "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```
georgettica, line 46: Just so we are clear on the API, this will be the external id

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+	defaultRegion = "us-east-1" // TODO extract to own config
+	credentialDir = "~/.aws"    // TODO extract to own config
+)
+
+const (
+	// the ClusterDeployment label key we're expecting to be able to pull the ClusterID from
+	cdClusterIDLabelKey string = "api.openshift.com/id"
```
georgettica, line 29: Is this something that can be pulled from the API package?
If not we should ask it to be/create a PR

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+// CloudClient is the client used for handling AWS Cloud Handler requests
+type CloudClient struct {
+	// clusterDeployment is the ClusterDeployment for the target cluster (the one
+	// for which a login url is requested)
+	clusterDeployment *hivev1.ClusterDeployment
+
+	// awsClient is the awsClient that has the BACKPLANE user credentials.  This client
+	// is used to create other clients in order to perform the necessary role chaining to
+	// access a customer's account via STS assume roles.
+	awsClient awsclient.AwsClient
```
georgettica, line 41: Why call the struct awsclient?
It's in a separate package

pkg/cloudclient/aws/aws.go
georgettica: Also maybe the issue is the struct and the interface have a similar name

@rafael-azevedo had thoughts on where the interfaces should live

pkg/cloudclient/aws/aws.go
georgettica: You can also create an empty struct instead of an empty interface

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+
+func (c *CloudClient) getAwsClientBuilder() awsclient.BuilderIface {
+	if c.awsClientBuilder != nil {
+		return c.awsClientBuilder
+	}
+	return &awsclient.Builder{}
+}
+
+func getAWSCredentialsFromDirectory(dir string) (string, string, error) {
+	accessKeyIDFilename := "aws_access_key_id"
```
georgettica, line 96: Shouldn't condys be at the top of the package?

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+		return "", "", fmt.Errorf("%s cannot be read from %s", accessKeyIDFilename, filepath)
+	}
+	secretKeyBytes, err := ioutil.ReadFile(path.Join(filepath, secretAccessKeyIDFilename))
+	if err != nil {
+		glog.Error(err)
+		return "", "", fmt.Errorf("%s cannot be read from %s", secretAccessKeyIDFilename, filepath)
+	}
+	accessKeyID := strings.TrimRight(string(accessKeyBytes), "\n")
+	secretKeyID := strings.TrimRight(string(secretKeyBytes), "\n")
+	return accessKeyID, secretKeyID, nil
```
georgettica, line 114: Wrap these into a struct so they are more meaningful?
Also makes the return value clearer and reduces flipping them around

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+
+	return &CloudClient{
+		clusterDeployment: cd,
+		awsClient:         awsClient,
+		ocmClient:         ocmClient,
+		region:            region,
+	}, nil
+}
+
+// The singleton client defined here is used to build the initial cloud client for a request.
```
georgettica, line 84: Comments don't adhere to the godoc format I think

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+		glog.Error(err)
+		return "", "", fmt.Errorf("%s cannot be read from %s", secretAccessKeyIDFilename, filepath)
+	}
+	accessKeyID := strings.TrimRight(string(accessKeyBytes), "\n")
+	secretKeyID := strings.TrimRight(string(secretKeyBytes), "\n")
+	return accessKeyID, secretKeyID, nil
+}
+
+// extracts a cluster ID from a given clusterdeployment
+func extractClusterID(cd *hivev1.ClusterDeployment) (string, error) {
```
georgettica, line 118: Helper functions can be lower in code, the top is where the public things are

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+
+	}
+	return clusterID, nil
+}
+
+// abstracted functionality in order to do the assume-role chaining in order to assume the support role
+// credentials in a given customer cluster. This function will get the accountclaim, assume the jump role,
+// and then from that assumed role make another assume_role call into the customer's account
+func (c *CloudClient) AssumeSupportRole() (*sts.Credentials, error) {
+
```
georgettica, line 131: Do we have to return the STS type back?
I know it's double converting but still asking

pkg/ocm/connection.go
```diff
@@ -0,0 +1,95 @@
...
+	return &ConnectionBuilder{}
+}
+
+// Build uses the information stored in the builder to create a new OCM connection.
+func (b *ConnectionBuilder) Build() (result *sdk.Connection, err error) {
+	if b.cfg == nil {
+		// Load the configuration file:
+		b.cfg, err = ocmcfg.Load()
+		if err != nil {
+			err = fmt.Errorf("Failed to load config file: %v", err)
```
georgettica, line 30: Change %v to %w

pkg/cloudclient/aws/aws.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func getAWSCredentialsFromDirectory(dir string) (string, string, error) {
+	accessKeyIDFilename := "aws_access_key_id"
+	secretAccessKeyIDFilename := "aws_secret_access_key"
+
+	// ensure that the directory has no trailing slash
+	filepath := strings.TrimSuffix(dir, "/")
+
+	accessKeyBytes, err := ioutil.ReadFile(path.Join(filepath, accessKeyIDFilename))
```
georgettica, line 102: Two statements in the same line, please split

pkg/cloudclient/aws/client/client.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// New is used when we already know the secrets and region, without any need to do any lookup.
+func (b *Builder) New(accessID, accessSecret, token, region string) (AwsClient, error) {
+	awsConfig := &aws.Config{
+		Region:                        aws.String(region),
+		CredentialsChainVerboseErrors: aws.Bool(true),
+	}
+	if token == "" {
+		os.Setenv("AWS_ACCESS_KEY_ID", accessID)
```
georgettica, line 46: Does this if statement make sense? I'm not sure

pkg/ocm/ocm_client.go
```diff
@@ -0,0 +1,87 @@
...
+	GetAWSAccountClaim(clusterID string) (*awsv1alpha1.AccountClaim, error)
+	GetClusterDeployment(clusterID string) (*hivev1.ClusterDeployment, error)
+}
+
+type client struct {
+	Conn   *sdk.Connection
+	logger *sdk.GlogLogger
+}
+
+func NewOcmClient(ocmConfigFile string) (Client, error) {
```
georgettica, line 23: You can simplify a but here so the if statement is shorter

georgettica (commented):
No real simplifying needed in the end, just separate commands into there own lines

pkg/ocm/ocm_client.go
ramonbutter: a what?

pkg/cloudclient/aws/aws.go
ramonbutter: interesting, but then when not providing the same struct type in the function call it will break no? 
From what I understood, you can pass any value to a parameter that implements an empty interface. 
Any type matches an empty interface

pkg/cloudclient/aws/aws.go
ramonbutter: not sure what's the problem... 
But I guess the fact that there is 3x awsClient, once used for the struct, the package, and the Interface. Yes, that can be improved.
Since we need to create several clients on-demand with different credentials, it makes sense to pack the client inside of its own package.
This variable holds the client to the origin of the chain. We may want to change that to the target, meaning the customer account in the end. Then this would become the client for executing commands in the customer AWS account

pkg/cloudclient/aws/aws.go
ramonbutter: I think it's ok for now since the function itself AssumeSupportRole is very STS specific. 
Later on, when we may want to add more clients for other cloud providers, we can try to unify the GetCredential returns.

pkg/cloudclient/aws/aws.go
georgettica: Yeah, just I don't like seeing a pointer to an empty Interface

pkg/cloudclient/aws/client/client.go
ramonbutter: yes I am wondering as well probably we can just use here the 
`credentials.NewStaticCredentials`
function as well

pkg/ocm/ocm_client.go
georgettica: I meant, change the if statements to use os.LookupEnv() which returns a boolean
and split the declaration of the envvar to the if statement

georgettica:
related to OSD-9836

pkg/cloudclient/aws/aws.go
ramonbutter: unfortunately no, its defined within gitlab in the uhc-clusters-service, so you would need to have vpn in order to get it.

cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -38,6 +41,35 @@ var ClusterMissingCmd = &cobra.Command{
...
+
+		// requires credentials to be at a specific dir (~/.ocm.json)
+		ocmClient, err := ocm.NewOcmClient("")
```
georgettica, line 49: maybe rename the function to `NewOcmClientFromPath`
so it's clear the "" is a filepath

pkg/cloudclient/aws/aws.go
georgettica: I meant the package can be called awsClient, the struct in the awsClient can be called Client and the field could be client or awsClient

pkg/cloudclient/aws/client/client.go
rafael-azevedo: We don't need to store the creds in ENV.

pkg/cloudclient/aws/client/client.go
```diff
@@ -0,0 +1,64 @@
+package awsclient
```
rafael-azevedo, line 1: I'd like to see this file be pkg/aws/aws.go

pkg/cloudclient/aws/client/client.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Client is a representation of the AWS Client
+type Client struct {
+	stsClient stsiface.STSAPI
+	AwsClient
+}
+
+// BuilderIface is a representation of the interface needed to create a builder
+type BuilderIface interface {
```
rafael-azevedo, line 30: We can remove the builder interface for now

shibumi:
Closing this as well, because it's not needed anymore

PR #12, opened 2022-02-21
feat(webhook): add webhook parsing

georgettica:
currently WIP as this is not the direction we are going to, but soon it'll be correct

* initial dockerfile: required for webhook stuff
* add test cleanup
* add 2 endpoints
  * sender - to send a webhook template
  * reciever - a webserver to recieve the webhook


georgettica:
related to OSD-10061

georgettica:
this is blocked on #16 #15 as they are helpers to make CHGM alerts parse correctly

pkg/pagerduty/reciever/reciever.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return "", UnmarshalErr{Err: err}
+	}
+	//incidentID := p.Incident.IncidentKey
+
+	return "", nil
+}
+
+func readPayloadFile(payloadFilePath string, reader fileReader) ([]byte, error) {
+	fileNotFoundErr := FileNotFoundErr{FilePath: payloadFilePath}
```
ramonbutter, line 34: Did you check whether that in this case (meaning payloadFilePath==""),  doesn't already fall into the PathError of reader.ReadFile?

georgettica: maybe?
I'll try and remove the check and see if this works still

ramonbutter: Yes, I would check that because if this is the case you are basically defining work twice since it should be also done by the reader library.

pkg/pagerduty/reciever/reciever.go
```diff
@@ -0,0 +1,55 @@
...
+type fileReader interface {
+	ReadFile(name string) ([]byte, error)
+}
+
+// ExtractExternalIDFromPayload will retrieve the payloadPath and return the externalID
+func ExtractExternalIDFromPayload(payloadFilePath string, reader fileReader) (string, error) {
+	data, err := readPayloadFile(payloadFilePath, reader)
+	if err != nil {
+		return "", fmt.Errorf("could not read the payloadFile: %w", err)
+	}
```
ramonbutter, line 21: and yes I would use the reader lib directly here, as you are not adding logic on top of it in the readPayloadFile.
So I would simply do something like.
```suggestion
	data, err := reader.ReadFile(payloadFilePath)
	if err != nil {
		return "", fmt.Errorf("could not read the payloadFile: %w", err)
	}
```

georgettica: i'll see what I can do

georgettica: I remember why that was.
we wanted to add comments to pull the payload from a file (which can change) so I extracted it to a function that can be converted/deleted once we have a different/better implementation

georgettica:
@ramonbutter added a commit for readPayloadFile, please review now

pkg/pagerduty/reciever/reciever.go
ramonbutter: not sure if I understand this right. But from what I understood, wouldn't you then rather define the function as follows?:
```

  func ExtractExternalIDFromPayload(payload []data){
           [..]
  }
  
  func ExtractExternalIDFromFile(filepath: string) {
    data, err := reader.ReadFile(payloadFilePath)
    if err != nil {
          return "", fmt.Errorf("could not read the payloadFile: %w", err)
     } 
    return ExtractExternalIDFromPayload(data)
  }
```
But anyway, may just be taste

georgettica: agreed this would be good, just didn't expose the API as I didn't see the usage of this yet, testing didn't require me a different API and we can extract this API later if we need it

georgettica: added a comment regarding this so I can address it when I need to

georgettica:
Thanks! Signing off 🐱

PR #13, opened 2022-02-22
OSD-9836: adding sts client

ramonbutter:


georgettica:
related to OSD-9836

pkg/aws/aws.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/aws/aws-sdk-go/service/sts/stsiface"
+)
+
+type Client interface {
```
rafael-azevedo, line 11: remember we don't actually need the interface here .. since it will live where the business logic goes

pkg/aws/aws.go
```diff
@@ -0,0 +1,46 @@
...
+		return AwsClient{}, err
+	}
+	return AwsClient{
+		Region:    *aws.String(region),
+		StsClient: sts.New(s),
+	}, nil
+}
+
+// AssumeRole is a function to wrap the AWS STS AssumeRole command
+func (a AwsClient) AssumeRole(roleARN string) (AwsClient, error) {
```
rafael-azevedo, line 37: @ramonbutter I think actually we can pass in a region here so that we have the ability of assuming a role in a different region then the client should we need it. We can do a check in the func to see if the passed in region is empty use the client region.

pkg/aws/aws.go
ramonbutter: in theory, yes. You should implement BL towards an abstraction instead of an implementation. This would allow you to replace the dependency. 
however not sure how practical this approach actually is, for this case, as probably will use clients in multiple pkg (or files)s and there is definitely just one SDK implementation. I am just a bit careful when applying unessesary design patterns with the cost of more code.
[here](https://github.com/openshift/aws-account-operator/blob/1fd83f85a9e90aa43f657833fdfeaad72778ea87/pkg/awsclient/client.go#L58) we didn't do it, and the [aws-sdk-go](https://github.com/aws/aws-sdk-go/blob/main/service/sts/stsiface/interface.go) doesn't do it as well. They are providing the interface to be able to mock their part out and enable easier testing of your code on top of it, without the need to define it on our end. I will leave it out for now as it was anyway not used, and we can discuss it later on

PR #14, opened 2022-02-23
OSD-10058: EC2 client and list Instances functionality

ramonbutter:


ramonbutter:
wow +30k loc. 
Most of it is part of the ec2 interface mock.
Maybe we really have to redefine the ec2 interface only with the functions we need and create our mocks on that.
WDYT?

cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -28,6 +28,22 @@ var ClusterMissingCmd = &cobra.Command{
 	Use:   "cluster-missing",
 	Short: "Will remediate the cluster-missing alert",
 	RunE: func(cmd *cobra.Command, args []string) error {
+		/*
```
rafael-azevedo, line 31: We don't need to store these comments

rafael-azevedo (commented):
It all looks good to me except the commented out section in the main

PR #15, opened 2022-02-24
feat(pagerduty): add getalerts endpoint

georgettica:
- add errors
- add private parser to local object
- put the alert type into 'types.go' (can move back to 'pagerduty.go' if
   wanted)
- general un-nesting of errors
- fix CONTRIBUTING.md

### DIFF
--- a/pkg/pagerduty/pagerduty.go
+++ b/pkg/pagerduty/pagerduty.go
@@ -50,7 +50,7 @@ func New(client *sdk.Client) (PagerDuty, error) {
 	return resp, nil
 }
 
-// MoveToEscalationPolicy will move the alerts EscalationPolicy to the new EscalationPolicy
+// MoveToEscalationPolicy will move the incident's EscalationPolicy to the new EscalationPolicy
 func (p PagerDuty) MoveToEscalationPolicy(incidentID string, escalationPolicyID string) error {
 	o := []sdk.ManageIncidentsOptions{
 		{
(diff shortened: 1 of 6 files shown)


georgettica:
/hold
as I want to add a description and a better commit message

but feel free to look at the code

georgettica:
look at OSD-10158

pkg/pagerduty/types.go
```diff
@@ -0,0 +1,6 @@
+package pagerduty
+
+type Alert struct {
+	ID   string
+	Body map[string]interface{}
```
ramonbutter, line 5: this field is going to change and become more written out in the future?

pkg/pagerduty/pagerduty.go
```diff
@@ -126,6 +126,38 @@ func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
...
+func (p PagerDuty) GetAlerts(incidentID string) ([]Alert, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), pagerDutyTimeout)
+	defer cancel()
+
+	o := sdk.ListIncidentAlertsOptions{}
+
+	alerts, err := p.c.ListIncidentAlertsWithContext(ctx, incidentID, o)
+	if err != nil {
+		sdkErr := sdk.APIError{}
+		if errors.As(err, &sdkErr) {
```
ramonbutter, line 139: To get rid of one level of nestign, write the following here:

```
if !error.As(err, &sdkErr){
 return nil, UnknownGetAlertsError{Err: err}
}
```

pkg/pagerduty/types.go
georgettica: don't think so, but this is no longer something that is a single field

if you want I can chuch it to pagerduty.go

pkg/pagerduty/pagerduty.go
```diff
@@ -126,6 +126,38 @@ func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
...
+			if sdkErr.StatusCode == http.StatusNotFound {
+				// this case can happen if the incidentID is not a valid incident (like a number prepended with zeroes)
+				return nil, IncidentNotFoundErr{Err: err}
+			}
+		}
+		return nil, UnknownGetAlertsError{Err: err}
+	}
+
+	res := []Alert{}
+	for i := range alerts.Alerts {
```
ramonbutter, line 153: whats about? :
```
for i, alert := range alerts.Alerts {
	localAlert := toLocalAlert(alert)
	res = append(res, localAlert)
}
```

pkg/pagerduty/pagerduty_test.go
```diff
@@ -284,6 +284,81 @@ var _ = Describe("Pagerduty", func() {
 			})
 		})
 	})
+	Describe("GetAlerts", func() {
```
ramonbutter, line 287: Just as a Note for future work: 
we are testing here only for the error return values, 
there is no testing of an expected actual array of alerts

pkg/pagerduty/pagerduty.go
ramonbutter: this is probably something that you can also put into commonErrHandler? Not sure right now

pkg/pagerduty/pagerduty.go
georgettica: I think there was something with `gosec` that didn't want to use this format
@NautiluX might remember the card

pkg/pagerduty/pagerduty.go
georgettica: did something similar, resolving this and please reopen a different comment on the new structure if it's still not good

pkg/pagerduty/pagerduty_test.go
georgettica: adding these test cases, thanks for pointing it out!

pkg/pagerduty/pagerduty.go
ramonbutter: yes, I think we had the issue here:
https://github.com/openshift/route-monitor-operator/commit/25e839a441cccfb1bdcad1e176c9713000096b84
However, in this case is a bit different IMO.
We are creating a new struct in toLocalAlert(alert) and appending this.
So we are not using the memory address of the iterator element.

georgettica: let's leave it as is? maybe add a note in a different card regarding this

ramonbutter: we would have been able to save one line by utilizing the elegant syntax :) But yes its nitpicking

georgettica: checked and gosec doesn't complain. so I'll leave the card but use your syntax

pkg/pagerduty/types.go
rafael-azevedo, line 5: I am not sure we want to use a map[string]interface{}  I would like to discuss this usage here at standup. I think we need a more defined struct. Even if it means marshalling the data temporarily into map[string]interface{}, then doing type checking to construct our own body with known type

georgettica: This was done to separate the getalerts from #16 , in an ideal world I would save the external struct here and marshal it later

georgettica: added a comment to explain why the body is not a normal struct

PR #16, opened 2022-02-24
feat(pagerduty): translate alert to externalid

georgettica:
- use mapstructure to convert between alert.Body to struct
   - use gopkg.in/yaml to convert the 'notes' field to struct

### DIFF
--- a/pkg/pagerduty/pagerduty.go
+++ b/pkg/pagerduty/pagerduty.go
@@ -5,9 +5,11 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"reflect"
 	"time"
 
 	sdk "github.com/PagerDuty/go-pagerduty"
+	"gopkg.in/yaml.v2"
 )
 
 const (
@@ -124,6 +126,78 @@ func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
 		return UnknownAddIncidentNoteError{Err: err}
 	}
 	return nil
+}
+
+// ExtractIDFromCHGM extracts from an Alert body until an external ID
+// the function is a member function but doesn't use any of the other funcs / types in the PagerDuty struct
+func (_ PagerDuty) ExtractIDFromCHGM(data map[string]interface{}) (string, error) {
+	var err error
+
+	externalBody, err := extractNotesFromBody(data)
+	if err != nil {
+		return "", fmt.Errorf("cannot marshal externalCHGMAlertBody: %w", err)
+	}
+
+	if externalBody == "" {
+		return "", AlertBodyDoesNotHaveNotesFieldErr{}
+	}
+
+	internalBody := internalCHGMAlertBody{}
+	err = yaml.Unmarshal([]byte(externalBody), &internalBody)
+	if err != nil {
+		return "", NotesParseErr{Err: err}
+	}
+
+	if internalBody.ClusterID == "" {
+		return "", AlertNotesDoesNotHaveClusterIDFieldErr{}
+	}
+
+	return internalBody.ClusterID, nil
+}
+
+// extractNotesFromBody will extract from map[string]interface{} the '.details.notes' while doing type checks
+// this is better than a third party as it is better maintained and required less dependencies
+func extractNotesFromBody(body map[string]interface{}) (string, error) {
+	var ok bool
+	_, ok = body["details"]
+	if !ok {
+		return "", AlertBodyExternalParseErr{FailedProperty: ".details"}
+	}
+
+	details, ok := body["details"].(map[string]interface{})
+	if !ok {
+		err := AlertBodyExternalCastErr{
+			FailedProperty:     ".details",
+			ExpectedType:       "map[string]interface{}",
+			ActualType:         reflect.TypeOf(body["details"]).String(),
+			ActualBodyResource: fmt.Sprintf("%v", body["details"]),
+		}
+		return "", err
+	}
+
+	notesInterface, ok := details["notes"]
+	if !ok {
+		return "", AlertBodyExternalParseErr{FailedProperty: ".details.notes"}
+	}
+
+	notes, ok := notesInterface.(string)
+	if !ok {
+		err := AlertBodyExternalCastErr{
+			FailedProperty:     ".details.notes",
+			ExpectedType:       "string",
+			ActualType:         reflect.TypeOf(details["notes"]).String(),
+			ActualBodyResource: fmt.Sprintf("%v", details["notes"]),
+		}
+		return "", err
+	}
+
+	return notes, nil
+}
+
+// internalCHGMAlertBody is a struct for manipulating CHGM from the note until the ClusterID
+type internalCHGMAlertBody struct {
+	// ClusterID in the ExternalId that the notes holds
+	ClusterID string `yaml:"cluster_id"`
 }
 
 // manageIncident will run the API call to PagerDuty for updating the incident, and handle the error codes that arise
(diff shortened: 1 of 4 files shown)


georgettica:
related to OSD-10081

georgettica:
look at OSD-10158

pkg/pagerduty/pagerduty.go
```diff
@@ -126,6 +128,50 @@ func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
...
+
+	var err error
+	externalBody := externalCHGMAlertBody{}
+
+	err = mapstructure.Decode(data, &externalBody)
+	if err != nil {
+		return "", fmt.Errorf("cannot marshal externalCHGMAlertBody: %w", err)
+	}
+
+	internalRawData := externalBody.Details.Notes
```
ramonbutter, line 143: All in all, this looks ok to me. However, I am generally not sure about the definition of these structures for unmarshalling as there are just used in one place.
I would maybe just go the way over json/yaml unstructured here. But its probably just taste, so feel free to go ahead anyway: 

```
details, ok := data["details"].((map[string]interface{}))
if !ok{
    return nil, UnexpectedFormatErr{field="details", data=data}
}
notes := make(map[string]interface{})
err := yaml.Unmarshal([]byte(details), &notes)
if !err{
    return nil, UnexpectedFormatErr{field="Notes", data=details}
}
clusterid, ok :=  notes["ClusterId"]
if err != nil {
    return nil, UnexpectedFormatErr{field="ClusterId", data=notes}
}
return clusterid, nil

```

pkg/pagerduty/pagerduty.go
```diff
@@ -126,6 +128,50 @@ func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
...
+
+	return internalBody.ClusterID, nil
+}
+
+// externalCHGMAlertBody is a struct for manipulating CHGM body until the note object
+type externalCHGMAlertBody struct {
+	// Details are the alert details
+	Details struct {
+		// Notes are the user given notes on the cluster
+		Notes string `mapstructure:"notes"`
```
ramonbutter, line 165: this type is exported by the external lib?

georgettica: yeah, the `mapstructure` helps the external lib comprehend what is going on inside the struct

pkg/pagerduty/pagerduty.go
georgettica: I agree they are used only in one place, bit as @rafael-azevedo mentioned and I support that, marshalling to map[string]interface{} is bad

ramonbutter: why is it bad? Isn't a standard format to map a string key to different values:
 
`key : value with type of {string, int, float, map[string]interface{}}`

georgettica: not really. if I can get something more concrete here I will choose to do so. this is why I am willing to push the third parties in. having types creates certainty.

anyways we should move this offline

pkg/pagerduty/pagerduty.go
```diff
@@ -126,6 +128,77 @@ func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
 	return nil
 }
 
+// ExtractExternalIDFromCGHMAlertBody extracts from an Alert body until an external ID
+// the function is a member function but doesn't use any of the types
+func (_ PagerDuty) ExtractExternalIDFromCGHMAlertBody(data map[string]interface{}) (string, error) {
```
rafael-azevedo, line 133: Can we come up with a Smaller function name ?

pkg/pagerduty/pagerduty.go
```diff
@@ -126,6 +128,77 @@ func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
...
+	}
+
+	notes, ok := notesInterface.(string)
+	if !ok {
+		return externalCHGMAlertBody{}, AlertBodyExternalCastErr{FailedProperty: ".details.notes", ExpectedType: "string", ActualType: reflect.TypeOf(notes).String()}
+	}
+
+	res := externalCHGMAlertBody{}
+	res.Details.Notes = notes
+	return res, nil
```
rafael-azevedo, line 184: cleaner here if you have
`return externalCHGMALertBody{details: notes}, nil `

georgettica: Sure

pkg/pagerduty/pagerduty.go
shibumi: What about ExtractExternalID? Or ExtractExternalCGHMAID? If we look on the receiver of the function, I think it becomes clear that this function is about alert bodys, but I never worked with PagerDuty. You might come up with a better idea :)

georgettica: nope, that's a good idea, naming is hard :/

pkg/pagerduty/pagerduty.go
georgettica: i had an anonymous struct in the `Details` field,  which made it hard to create the stuct there. so I removed the whole struct as I only need to extract the "notes" field

pkg/pagerduty/pagerduty.go
```diff
@@ -126,6 +128,77 @@ func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
...
+		return "", AlertBodyDoesNotHaveNotesFieldErr{}
+	}
+
+	internalBody := internalCHGMAlertBody{}
+	err = yaml.Unmarshal([]byte(internalRawData), &internalBody)
+	if err != nil {
+		return "", NotesParseErr{Err: err}
+	}
+
+	// TODO: verify if no ClusterID is missing (doesn't break the test case, but should be noted
```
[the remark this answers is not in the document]
georgettica: no, I think I'll keep it as this TODO is for the case we have another parser and it arrives here.

@rafael-azevedo noted that this won't happen as if this case would occur this function should fail but I wanted to note it somewhere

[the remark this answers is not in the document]
shibumi: @georgettica Does it make sense to open a github issue for this todo? If yes, I can open one.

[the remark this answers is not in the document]
georgettica: It doesn't really, but maybe we should track it with out internal ticketing
If you want I can show you the ropes there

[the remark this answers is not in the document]
georgettica: added an error, so it's fine for now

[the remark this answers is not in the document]
shibumi: lgtm

pkg/pagerduty/pagerduty.go
```diff
@@ -126,6 +128,76 @@ func (p PagerDuty) AddNote(incidentID string, noteContent string) error {
...
+	var ok bool
+	_, ok = body["details"]
+	if !ok {
+		return "", AlertBodyExternalParseErr{FailedProperty: ".details"}
+	}
+
+	details, ok := body["details"].(map[string]interface{})
+	if !ok {
+		err := AlertBodyExternalCastErr{
+			FailedProperty: ".details",
```
ramonbutter, line 170: What's about adding the body here as a parameter? When you print the error, you immediately got the input that has caused this exception

shibumi (approved):
/lgtm

georgettica:
@shibumi 
It's fir a forward slash In the from `/lgtm`

pkg/pagerduty/pagerduty.go
georgettica: done

georgettica:
@shibumi editing is not captured by the bot, so you need to write it correctly the first time

(or file an issue somewhere, not sure where this should go)

PR #18, opened 2022-03-01
OSD-10060 Add ability to list cloudtrail stoppedinstances events

ninabauer:
Adds ability to list out events from CloudTrail that are related to stopped ec2 instances.


georgettica:
related to OSD-10060

started doing this as it'll link to the right link by the bot

pkg/aws/aws.go
```diff
@@ -170,3 +188,28 @@ func (a *AwsClient) listInstancesWithFilter(filters []*ec2.Filter) ([]*ec2.Insta
 	}
 	return instances, nil
 }
+
+// ListStopInstancesEvents lists StopInstances events from CloudTrail
+func (a AwsClient) ListStopInstancesEvents() ([]*cloudtrail.Event, error) {
```
ramonbutter, line 193: It may make sense to define this function differently. At the moment we are able to list all events of all instances. However, I think we rather need to list events on stopped instances meaning:

```suggestion
func (a AwsClient) ListStopInstancesEvents(instanceID string) ([]*cloudtrail.Event, error) {
```

cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -28,14 +29,24 @@ var ClusterMissingCmd = &cobra.Command{
 	Use:   "cluster-missing",
 	Short: "Will remediate the cluster-missing alert",
 	RunE: func(cmd *cobra.Command, args []string) error {
+		CAD_AWS_FOLDER, ok := os.LookupEnv("CAD_AWS_FOLDER")
+		if !ok {
+			return fmt.Errorf("could not load CAD_AWS_FOLDER envvar")
+		}
+
+		_, err := aws.NewClientFrmFileCredentials(CAD_AWS_FOLDER, "")
```
georgettica, line 37: discussed somewhere else, but this library API is less usable than others (gave you a replacement snippet for this)

pkg/aws/aws.go
ramonbutter: however this strongly depends on what we are returning [here](https://github.com/openshift/configuration-anomaly-detection/blob/739c2b02d9c4d38bb0647b9b3a6b5413595783ea/pkg/aws/aws.go#L126)

pkg/aws/aws.go
```diff
@@ -65,10 +69,24 @@ func NewClient(accessID, accessSecret, token, region string) (AwsClient, error)
...
 	}
 
+	cloudtrailConfig := &aws.Config{
+		Region:      aws.String(region),
+		Credentials: credentials.NewStaticCredentials(accessID, accessSecret, token),
+		Retryer: client.DefaultRetryer{
+			NumMaxRetries:    maxRetries,
+			MinThrottleDelay: 2 * time.Second,
+		},
+	}
```
georgettica, line 79: @ramonbutter can we move all of the configs to one place?

georgettica (commented):
noticed stuff when this became a PR.
LMK if things are unclear

pkg/aws/aws.go
georgettica: wanted to make sure if this duplication makes sense or not

rafael-azevedo:
@ninabauer Everything looks good. I agree with the comments that we should using the same config when creating all the clients. I'd also like the function with instanceID `func (a AwsClient) ListStopInstancesEvents(instanceID string) ([]*cloudtrail.Event, error)`

pkg/aws/aws_test.go
```diff
@@ -53,7 +53,7 @@ var _ = Describe("Aws", func() {
 		})
 		When("the client fails for arbitary reason", func() {
 			It("the error is propagated and empty client returned", func() {
-				errOcc := errors.New("something happend")
+				errOcc := fmt.Errorf("something happend")
```
georgettica, line 56: extract this into a variable and use it in the three tests :)

pkg/aws/aws.go
```diff
@@ -26,12 +28,14 @@ const (
 
 //go:generate mockgen -destination mock/stsmock.go -package $GOPACKAGE github.com/aws/aws-sdk-go/service/sts/stsiface STSAPI
 //go:generate mockgen -destination mock/ec2mock.go -package $GOPACKAGE github.com/aws/aws-sdk-go/service/ec2/ec2iface EC2API
+//go:generate mockgen -destination mock/cloudtrailmock.go -package $GOPACKAGE github.com/aws/aws-sdk-go/service/cloudtrail/cloudtrailiface CloudTrailAPI
 
 // Client is a representation of the AWS Client
```
shibumi, line 33: Minor nit: // Client breaks the Go Doc strings, because the comment is not starting with AwsClient.
You might want to replace the doc strings with:

// AwsClient is a representation of the Amazon AWS Client.

shibumi: Mhh I just realized that we have a constructor called `NewClient` but we are creating a struct called `AwsClient` with it. I don't know if this is already in use, but it might makes sense to rename one of those for a cleaner documentation.
The Go syntax best practice guide usually suggests that a New<Object> constructor should match the struct name.
So, it should be either NewClient and Client or NewAwsClient and AwsClient:

https://go.dev/doc/effective_go#:~:text=func%20newfile(fd%20int%2C%20name%20string)%20*file%20%7B%20if%20fd%20%3C%200%20%7B%20return%20nil%20%7D%20f%20%3A%3D%20new(file)%20f.fd%20%3D%20fd%20f.name%20%3D%20name%20f.dirinfo%20%3D%20nil%20f.nepipe%20%3D%200%20return%20f%20%7D

That's really just a minor nit and a design decision and this should not block the PR merge process :)

pkg/aws/aws_test.go
```diff
@@ -50,9 +53,9 @@ var _ = Describe("Aws", func() {
 				Expect(c).ShouldNot(Equal(aws.AwsClient{}))
 			})
 		})
-		When("the client fails for arbitary reason", func() {
+		When("the client fails for arbitrary reason", func() {
 			It("the error is propagated and empty client returned", func() {
-				errOcc := errors.New("something happend")
+				errOcc = fmt.Errorf("something happened")
```
georgettica, line 58:
```suggestion
```

pkg/aws/aws_test.go
```diff
@@ -119,9 +122,9 @@ var _ = Describe("Aws", func() {
 				Expect(len(c)).Should(Equal(1))
 			})
 		})
-		When("the client fails with an arbitary error", func() {
+		When("the client fails with an arbitrary error", func() {
 			It("the error is propagated and nothing is returned", func() {
-				errOcc := errors.New("something happend")
+				errOcc = fmt.Errorf("something happened")
```
georgettica, line 127:
```suggestion
```

pkg/aws/aws_test.go
```diff
@@ -130,4 +133,65 @@ var _ = Describe("Aws", func() {
...
+			It("the values are returned and the cloudtrail api is called just once", func() {
+				mockCloudTrailClient.EXPECT().LookupEvents(gomock.Any()).Return(lookupEventOut, nil).Times(1)
+				c, err := client.ListStopInstancesEvents()
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(len(c)).Should(Equal(1))
+			})
+		})
+		When("the client fails with an arbitrary error", func() {
+			It("the error is propagated and nothing is returned", func() {
+				errOcc = fmt.Errorf("something happened")
```
georgettica, line 188:
```suggestion
```

georgettica (commented):
sorry for the late comments, didn't see I didn't publish

PR #19, opened 2022-03-01
docs(aws): add docs

georgettica:
- move pagerduty docs to it's own folder

### DIFF
--- a/pkg/aws/README.md
+++ b/pkg/aws/README.md
@@ -1,0 +1,36 @@
+# AWS Package
+
+Use the `aws.New` to create an AWS client, and use the functions it has to interact with AWS resources
+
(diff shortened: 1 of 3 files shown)


pkg/aws/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+### extract the external ID
+```
+I= #Incident ID
+pd rest:get -e=/incidents/$I/alerts | jq -r '.alerts[].body.details.notes' | yq .cluster_id
+```
+
+### login to the owner cluster
+using a private script, we can login to the cluster that owns the cluster we want to check (as it holds the aws account)
+-- DM privately to get the details
```
ramonbutter, line 23: what does that mean?

georgettica: replied privately

PR #22, opened 2022-03-03
OSD-10119: added GetSupportRoleARN in ocm client

ramonbutter:
Enabling ability to get support role ARN for sts cluster

### DIFF
--- a/pkg/ocm/ocm.go
+++ b/pkg/ocm/ocm.go
@@ -21,21 +21,35 @@ type ocmClient struct {
 // New will create a new ocm client by using the path to a config file
 // if no path is provided, it will assume it in the default path
 func New(ocmConfigFile string) (ocmClient, error) {
-
 	client := ocmClient{}
 	cfg, err := newConfigFromFile(ocmConfigFile)
 	if err != nil {
-		return client, fmt.Errorf("failed to load config file: %w", err) // ConfigError{path: ocmConfigFile, err:err}
+		return client, fmt.Errorf("failed to load config file: %w", err)
 	}
 	client.cfg = cfg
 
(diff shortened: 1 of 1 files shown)


pkg/ocm/ocm.go
```diff
@@ -38,6 +38,22 @@ func New(ocmConfigFile string) (ocmClient, error) {
 	return client, nil
 }
 
+// GetSupportRoleARN returns the support role ARN that allows the access to the cluster
+func (client ocmClient) GetSupportRoleARN(clusterID string) (string, error) {
```
georgettica, line 42: please test this

rafael-azevedo: @georgettica I think we should be fine without a unit test for this function since this really is no additional logic.

georgettica: I'll disagree, but ok

rafael-azevedo:
@ramonbutter Can you just put a description on the PR then we can merge tomorrow

pkg/ocm/ocm.go
fahlmant: What would a test look like? Its just a GET against an API and then returning an error if the field is blank, or the field. What logic is there to test?

georgettica: That is true, but I would like to cover all public APIs with a test, even if it doesn't make sense.

Just because of standard practices for all

shibumi:
> @ramonbutter Can you just put a description on the PR then we can merge tomorrow
@rafael-azevedo  wouldn't it make more sense to add a better description in the commit message? The commit message just says:"added GetSupportRoleARN in ocm client"

PR #23, opened 2022-03-04
[WIP]: add ability to send chgm servicelog in ocm

ramonbutter:
This PR enables the ocm client to send a cluster has gone missing SL.
It was tested manually. 
Some unit tests will follow soon, but feel free to comment already until then.


pkg/ocm/ocm.go
```diff
@@ -3,16 +3,24 @@ package ocm
...
 	sdk "github.com/openshift-online/ocm-sdk-go"
 
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	servicelog "github.com/openshift-online/ocm-sdk-go/servicelogs/v1"
 	awsv1alpha1 "github.com/openshift/aws-account-operator/pkg/apis/aws/v1alpha1"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 )
 
+var (
+	sl_clusterHasGoneMissing = "https://raw.githubusercontent.com/openshift/managed-notifications/master/osd/cluster_has_gone_missing.json"
```
georgettica, line 21: doesn't using `master` mean it might break in the future? isn't it better to have it pinned to a release / commit?

pkg/ocm/ocm.go
```diff
@@ -66,6 +74,23 @@ func (client ocmClient) GetAWSAccountClaim(clusterID string) (*awsv1alpha1.Accou
 	return ac, err
 }
 
+// GetClusterInfo returns cluster information from ocm by using either internal, external id or the cluster name 
+// Returns a v1.Cluster object or an error
+func (client ocmClient) GetClusterInfo(identifier string) (*v1.Cluster, error) {
+	q := fmt.Sprintf("(id like '%[1]s' or external_id like '%[1]s' or display_name like '%[1]s')", identifier)
+	resp, err := client.conn.ClustersMgmt().V1().Clusters().List().Search(q).Send()
```
georgettica, line 81: I think i've seen this code like 3 times already. not in this codebase but altogether.. might suggest adding an `ocm-sdk-go` util for this

georgettica (commented):
looks good
/lgtm
/approve
/hold
discuss privately if @rafael-azevedo should gate this or not

pkg/ocm/ocm.go
ramonbutter: yes thought of it as well. It makes definitely sense to evaluate it in the future. Pinning it to master will just take the latest template so far but makes it prone to failures. But I left it out for now

pkg/ocm/ocm.go
ramonbutter: yes we are going to encapsulate the ocm communication for our tests in an own interface and struct the next PR, to make it also testable

pkg/ocm/ocm.go
```diff
@@ -66,6 +74,23 @@ func (client ocmClient) GetAWSAccountClaim(clusterID string) (*awsv1alpha1.Accou
...
+	if err != nil || resp.Error() != nil || resp.Status() != http.StatusOK {
+		return nil, fmt.Errorf("received error while fetch ClusterInfo from ocm: %w with resp %#v", err, resp)
+	}
+	if resp.Total() > 1 {
+		return nil, fmt.Errorf("the provided cluster identifier is ambiguous: %s", identifier)
```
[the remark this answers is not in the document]
ramonbutter: Yep, makes sense. I wasn't aware of a one-liner to do so, so I skipped it but maybe we can just use a json dump into the error message or so

pkg/ocm/ocm.go
rafael-azevedo: @ramonbutter I think it will be better if we hard code the message instead of using the file from github. This will prevent issues if the messsages change or if github is down.

georgettica: You can use triple ticks to do that.
This would also mean that we will have to get mgmt approval for changing the SL here aswell

rafael-azevedo: I saw we put currently approved here and if it changes in the future we make a new PR with current wording.

rafael-azevedo:
@ramonbutter this looks great. If we can get the tests to pass and address hard coding the service log message

pkg/ocm/ocm.go
shibumi: Agree.. having the hardcoded service log in JSON in there is the better option

rafael-azevedo:
@ramonbutter @georgettica I am merging this PR , I will put a card in for a refactor which includes putting the service log into a datastructure and not using an external call to get it. This will also solve the go sec issue.

shibumi:
@rafael-azevedo I think @georgettica  meant adding the `//nosec` as comment in the code that triggers the linter

PR #24, opened 2022-03-07
chore(gosec): fix gosec warning

georgettica:
related to [OSD-10161](https://issues.redhat.com/browse/OSD-10161)

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/aws/aws.go
```diff
@@ -76,15 +77,18 @@ func NewClient(accessID, accessSecret, token, region string) (AwsClient, error)
 
 // NewClientFrmFileCredentials creates a new client by reading credentials from a file
 func NewClientFrmFileCredentials(dir string, region string) (AwsClient, error) {
-	filepath := strings.TrimSuffix(dir, "/")
+	dir = strings.TrimSuffix(dir, "/")
+	dir = filepath.Clean(dir)
```
ramonbutter, line 81: can we not leave this one out, and only do it once in 83?

ramonbutter: and line 80 may also be unrequired, as the 'path.Join' should take care of it how to concatenate the path appropriately. But I will test it

georgettica:
@ramonbutter this is a low prio ticket that wasn't pulled into the sprint
/hold
unless you want to do it regardless

ramonbutter:
It's ok I guess, you can merge it. My comments were just addressing minor details (that were part of my earlier commits) and I can have a look at it later on

georgettica:
Then I'll merge
/unhold

georgettica:
related to OSD-10161

PR #26, opened 2022-03-07
test(ocm): add first tests for the ocm pkg

shibumi:


test/ocm_test.json
```diff
@@ -0,0 +1,10 @@
+{
+    "access_token": "DUMMYVALUE",
```
georgettica, line 2: Maybe move this file.i to the package, as this is the file that is being tested (no need to create another folder structure IMO)

pkg/ocm/ocm_test.go
```diff
@@ -0,0 +1,44 @@
...
+	sdkcfg "github.com/openshift-online/ocm-cli/pkg/config"
+)
+
+var _ = Describe("OCM", func() {
+	Describe("When trying to load a configuration", func() {
+		var (
+			config         *sdkcfg.Config
+			configLocation string
+		)
+		JustBeforeEach(func() {
```
georgettica, line 15: I understand how it works, but maybe this should be explained for future developers

georgettica:
This is how ocm-cli does it..
Is that good?
https://github.com/openshift-online/ocm-cli/blob/main/tests/list_clusters_test.go

pkg/ocm/ocm_test.go
shibumi: Done

test/ocm_test.json
shibumi: I thought we follow the golang project structure: https://github.com/golang-standards/project-layout#test
They suggest creating a test directory, but if this is the only test file, you might be right and it would make sense to move the file to the location where it is actually used. 

@ramonbutter  suggested we could just generate the file during the test process. Do you have an opinion about that?

shibumi:
> This is how ocm-cli does it.. Is that good? https://github.com/openshift-online/ocm-cli/blob/main/tests/list_clusters_test.go

Monday evening @ramonbutter  and me tried to do it with interfaces, however due to a few issues I have with the mocking, I think I will try your webserver solution in a separate branch. Later, we could compare both test methods and see what works better for us.

test/ocm_test.json
georgettica: I wouldn't want to generate it, but you can use a in-memory fs for it (can send a link in a sec)

shibumi: Ah I think I know what you mean. We can use the new Go Embed library.

shibumi: Never mind, i think this might be the better solution: https://pkg.go.dev/testing/fstest#MapFS

georgettica: exactly https://github.com/openshift/configuration-anomaly-detection/blob/ab181d77a75051b1e8cfbe53b0cdb3389af0ebbc/pkg/pagerduty/pagerduty_test.go#L517-L523

shibumi: This doesn't  work for us.. the problem is that with MapFS I have to pass a file handle instead of a filepath with type string.
So, even when I pass the filepath to the file in the MapFS, the sdkcfg.Location() function call will fail, because it will search on the wrong filesystem.

TBH, I would prefer that upstream fixes this and implements a io.Reader interface for passing config files around.. this would be by far more testable for downstream projects like ours.

shibumi:
@georgettica This should be ready to review now..We have finished most test cases with mocking. The test coverage is at 83% now.

We should probably squash our commits, right?

georgettica:
If there is no good reason to keep the commits together than I would say yes, squash them

pkg/ocm/ocm_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package ocm
```
ramonbutter, line 1: this should be ocm_test

pkg/ocm/ocm_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestPagerduty(t *testing.T) {
```
ramonbutter, line 10: TestOcm

test/ocm_test.json
ramonbutter: that's annoying, but then leave it out.  And to explain my thoughts about generating the test config file (and cleaning them up running the test):
I am not a big fan of defining the config twice, once [here](https://github.com/openshift/configuration-anomaly-detection/pull/26/files#diff-f579fc1054cc8232c801eee714549d207c451899358068b5a4757154b8e9c7aaR29) and in the file. It's unclearer how the expectation is reasoned. Would only go commit stuff for complex test data (photos, binaries, etc.).  But this is just my taste, and am ok with it. Lets just continue and merge for the case this PR works

shibumi:
@ramonbutter all done

pkg/ocm/ocm_test.go
```diff
@@ -0,0 +1,147 @@
...
+		awsAccountClaim   *awsv1alpha1.AccountClaim
+		err               error
+		clustername       string
+		supportRoleArn    string
+	)
+	Describe("When fetching a ClusterDeployment", func() {
+		BeforeEach(func() {
+			mockCtrl = gomock.NewController(GinkgoT())
+			mocOCMConnection = mocks.NewMockocmHandlerIf(mockCtrl)
+			client = &ocm.OcmClient{Comm: mocOCMConnection}
```
georgettica, line 33: so this does seem to be worse.
you can have a `func New() ocm.OcmClient` that will take the connection / take an options struct that holds the connection struct

pkg/ocm/ocm.go
```diff
@@ -13,31 +13,51 @@ import (
...
+	if err != nil {
+		return "", err
+	}
+	return response.Body().Resources()[resourceKey], nil
+}
+
+// OcmClient provides a client for the Openshift Cluster Manager
+type OcmClient struct {
+	//conn *sdk.Connection
+	Cfg  *sdkcfg.Config
```
georgettica, line 36: if it's public it should have a public docstring?

also the name seems to be too short for the scope this field has

pkg/ocm/ocm_test.go
```diff
@@ -0,0 +1,147 @@
...
+			err = fmt.Errorf("some error")
+			clustername = "test-cluster"
+			clusterDeployment = &hivev1.ClusterDeployment{Spec: hivev1.ClusterDeploymentSpec{ClusterName: clustername}}
+		})
+		When("cluster ID is valid", func() {
+			It("should return a valid ClusterDeployment", func() {
+				cdstring, _ := json.Marshal(clusterDeployment)
+				mocOCMConnection.EXPECT().OcmGetResourceLive(clustername, "cluster_deployment").Return(string(cdstring), nil).Times(1)
```
georgettica, line 41: break this line down so it won't be too long?

georgettica (commented):
not a full pass, but may be a good iteration

pkg/ocm/ocm_test.go
shibumi: done

pkg/ocm/ocm_test.go
shibumi: makes sense

pkg/ocm/ocm.go
shibumi: I think if we use a constructor for this one. We can get rid of the public fields as well

ramonbutter: yes, we don't need it at all actually afaik. We can get rid of it.

pkg/ocm/ocm_test.go
shibumi: @georgettica  https://github.com/openshift/configuration-anomaly-detection/pull/26/commits/0277c38a13c8fce6100ff346f53dc2e9a10b3e5f

Is this what you had in your mind? I introduced a constructor, but I cannot make the OcmClient struct private, because we are running the tests in a separate package called `ocm_test`.

pkg/ocm/ocm.go
shibumi: I made the two fields private

pkg/ocm/ocm_test.go
```diff
@@ -0,0 +1,147 @@
...
+				Expect(err).ShouldNot(HaveOccurred())
+				Expect(clusterDeployment).Should(Equal(cd_out))
+			})
+		})
+		When("the cluster doesn't exist or any other error happends in the sdk", func() {
+			It("wrapps the error message and returns nil", func() {
+				mocOCMConnection.EXPECT().OcmGetResourceLive(clustername, "cluster_deployment").Return("", err).Times(1)
+				cd_out, err_out := client.GetClusterDeployment(clustername)
+				Expect(err_out).Should(HaveOccurred())
+				Expect(errors.Unwrap(err_out)).Should(Equal(err))
```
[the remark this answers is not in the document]
ramonbutter: yea I know... I am just hesitant with defining errors for pinning lines in tests. This should also be just fine. It basically says the error is going to be bubbled up with whatever error message. So would like to leave it now as it is

test/ocm_test.json
shibumi: Note: i have added a second config file with invalid json and another test case... I think the /test directory should be fine as suggested in the standard golang project structure repository.

pkg/ocm/ocm_test.go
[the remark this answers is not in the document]
shibumi: Defining errors would have the advantage that we could export them. In a library this would make sense, but we are writing a client.. so no idea if we really need exported error variables :thinking:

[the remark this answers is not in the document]
georgettica: I think errors.Unwrap is just one level down, so it assumes you did: `fmt.Errorf("something: %w", actual_err)` and not
```
fmt.Errorf("other: %w", fmt.Errorf("something: %w", actual_err))
```
which might happpen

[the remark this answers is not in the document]
georgettica: I think also that adding an error is not too much of an overhead, it makes the test clearer, code more explicit and we "lose" a few lines of error code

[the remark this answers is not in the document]
shibumi: sorry for the delay, @georgettica. I missed your notification. I will have a look on this, that we can merge this PR ASAP.
Consider it done until tomorrow morning.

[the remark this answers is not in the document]
georgettica: Yes it was.
Just trying to understand how we should change the test if we are moving away of testing the error that comes back

georgettica:
@shibumi do you know how to proceed here?
I am not sure we discussed it

As I see it, you can remove any and all custom errors that are not generic that ate blocking you.

Additionally I would like that if we do compare errors not torely on the wrapping structure (expect the error is one layer deep using unwrap)

shibumi:
@georgettica  I don't know what you mean with "Custom errors". Do you mean statements like:

```go
var ErrCustomFailure = errors.New("error in ...")
```

? We do not have any statement like this in our code. Same applies to structured errors like:

```go
type QueryError struct {
    Query string
    Err   error
}
```


This is how I would continue. I would try to remove the unwrap lines and just use errors.Is() if possible..

georgettica:
That is something I am fine with.
Continue with errors.is

And if you need to do errors.new than this is the discussion we had

shibumi:
@georgettica Do you mind having a second look? We use errors.Is now..

georgettica:
this solves my only comment
/approve
/lgtm
/hold

holding as to say if someone has something to note, please say now or forever hold your regrest

shibumi:
@georgettica I've added another customError as suggested by @ramonbutter  + fixed a spell issue.

shibumi:
@rafael-azevedo wanna have a final look, before we merge? If you are good with it, feel free to do a /unhold

PR #27, opened 2022-03-08
ci: add linting, code coverage and other useful checks

shibumi:
I don't know if this is useful, but I invested a few minutes and build a proper release pipeline + added tests with golangci.
If some of these steps are already covered by our jenkins pipeline I can remove them.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,26 @@
+codecov:
+  notify:
+    require_ci_to_pass: no
+
+coverage:
+  precision: 2
+  round: down
+  range: "20...100"
+
+  status:
+    project: no
+    patch: no
+    changes: no
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment:
+  layout: "reach,diff,flags,tree"
+  behavior: default
+  require_changes: no
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 /bin
+dist
+*.out
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,27 @@
+run:
+  # Our CI expects a vendor directory, hence we have to use -mod=readonly here as well.
+  modules-download-mode: readonly
+
+issues:
+  # on default, golangci-lint excludes gosec. Hence, we have to
+  # explicitly disable the exclude-use-default: https://github.com/golangci/golangci-lint/issues/1504
+  exclude-use-default: false
+
+# individual linter configs go here
+linters-settings:
+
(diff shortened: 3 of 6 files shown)


shibumi:
converting this to a WIP/draft... I want to make a few adjustments:

1. provide documentation for the pipeline
2. edit Makefile, so it is consistent with the pipeline. @georgettica  suggested that we could use goreleaser for building local binaries, too.

georgettica:
Odd, I can't add emojis to these comments.
But I agree and think this is a good way to go

.goreleaser.yaml
```diff
@@ -0,0 +1,71 @@
+project_name: cadctl
+builds:
+  - ldflags:
+      - "-s -w"
+      - "-extldflags=-zrelro"
```
georgettica, line 5: you had comments before on zrelro being a requirement for stuff, can you add it again?

shibumi: Oh yeah.. I don't know where they went

shibumi: done

shibumi:
@georgettica  what do we do with this PR here? Now, that we know that the GitHub pipeline won't work GoReleaser will not work as well (50% of the goreleaser config file only works in combination with github actions).

I would suggest we remove the goreleaser/github workflow part and just add the codecov hack files we need for code coverage.

I will try to have a look on prow (openshift/release). Would be cool if we could at least add `golangci-lint` as Linter for our Go code.

georgettica:
Sadly I agree.

I think we can also ask if we can add this GHA as it hardens our binaries, but we will see about that later on

shibumi:
@georgettica  well, the binary hardening can be added in the makefile as well. These are just go flags :)

georgettica:
Then this can be done aswell fairly easy

shibumi:
@georgettica Done. I have incorporated the changes from GoReleaser into the Makefile. Will add make targets for golangci-lint gosec etc next.

shibumi:
@georgettica this adds make targets for golangci-lint + gosec and removes the goreleaser files again.

Do we need a separate target for installing the lint/test dependencies? (running go install {gosec|golangci-lint})?

georgettica:
I think gosec was not officially noted as something we need (have a discussion on this still)

but I think golangci-lint is a good addition

georgettica (changes requested):
One we merge this we can retest the release code

georgettica:
@shibumi could you make all of these commands pass?
I want to merge this so the CI will verify our stuff :)

georgettica:
it seems go test also needs the vendor thingy XP

shibumi:
/retest


Triggering a retest because the job is pending for over 20min now..

PR #30, opened 2022-03-16
chore(lint): add misspell linter

georgettica:
this is super cool and it finds common misspells

PR #32, opened 2022-03-20
fix(typos): using typos

georgettica:
rust has a typos cli, which works better than misspell

this is what it found
https://github.com/crate-ci/typos


georgettica:
see #30

PR #35, opened 2022-03-31
chore: bump Go to 1.18

shibumi:
I don't know if this would make sense, because openshift/boilerplate is still on Go 1.17, but I bumped CAD on Go 1.18 and it builds.

Would it make sense to bump to Go 1.18 or do we wait?


georgettica:
I want to go 1.18, but we _should_ wait for the whole org

basically did something similar in RMO (moved to latest operator-sdk) and not it's like a sore thumb

not opposed, but not sure if to say yes to this

georgettica:
/retest
this error is not something in the code I think

shibumi:
@georgettica it's a timeout in the linter

georgettica:
let's give a huge timeout  `--timeout=5m` and add `--verbose`  to see where it fails

shibumi:
@georgettica  lol.. my local golangci-lint fails:
```
❯ golangci-lint run --help
panic: load embedded ruleguard rules: rules/rules.go:13: can't load fmt

goroutine 1 [running]:
github.com/go-critic/go-critic/checkers.init.22()
        github.com/go-critic/go-critic@v0.6.2/checkers/embedded_rules.go:46 +0x4b4
```

georgettica:
see - https://github.com/go-critic/go-critic/issues/1193

georgettica:
so let's stay on go1.17, as all the linters work there :)
once they are mature enough we'll migrate?

shibumi:
> so let's stay on go1.17, as all the linters work there :)
> once they are mature enough we'll migrate?

Sounds good. Do we leave this PR open or do we close it? I think it would make sense to close it, but document our problems somewhere... we might run into the same issues in a year when we try to upgrade to Go 1.18

shibumi:
This is from now on tracked in: OSD-10639

PR #36, opened 2022-03-31
fix(ocm): remove ocm-cli dependency

shibumi:
We should not depend on openshift-online/ocm-cli, because
this package was never intended as library. Let us
extract the functions we need and see how that goes.

This commit fixes: [OSD-10576](https://issues.redhat.com/browse/OSD-10576)


shibumi:
@georgettica  @rafael-azevedo This PR removes the ocm-cli dependency as discussed in https://issues.redhat.com/browse/OSD-10576

However, I am not happy with that PR yet, because it introduces several new dependencies to our project (one of the deps is external).

The current state just extracts all functions we use for the config handling from ocm-cli and puts them in our project.

Now, we should check what functionality we want/need and what not.

pkg/ocm/config.go
```diff
@@ -0,0 +1,157 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/golang/glog"
+	"github.com/mitchellh/go-homedir"
+	sdk "github.com/openshift-online/ocm-sdk-go"
+	"github.com/openshift/rosa/pkg/debug"
```
georgettica, line 13: I'm not sure you were supposed to take rosa/debug and Rosa/info

pkg/ocm/config.go
```diff
@@ -0,0 +1,157 @@
...
+	RefreshToken string   `json:"refresh_token,omitempty" doc:"Offline or refresh token."`
+	Scopes       []string `json:"scopes,omitempty" doc:"OpenID scope. If this option is used it will replace completely the default scopes. Can be repeated multiple times to specify multiple scopes."`
+	TokenURL     string   `json:"token_url,omitempty" doc:"OpenID token URL."`
+	URL          string   `json:"url,omitempty" doc:"URL of the API gateway. The value can be the complete URL or an alias. The valid aliases are 'production', 'staging' and 'integration'."`
+	User         string   `json:"user,omitempty" doc:"User name."`
+	Pager        string   `json:"pager,omitempty" doc:"Pager command, for example 'less'. If empty no pager will be used."`
+}
+
+// Connection creates a connection using this configuration.
+func (c *Config) Connection() (connection *sdk.Connection, err error) {
```
georgettica, line 37: Do we have tests for this? And if not we should card it (we have now business logic to verify

pkg/ocm/config.go
shibumi: Do we need these loggers and go-homedir even? Our code should only run on POSIX systems (Mac/Linux) so we could just read `$HOME` and that's it.

pkg/ocm/config.go
shibumi: Not yet.. the big question is if we really need this.

shibumi:
> Additionally, we should make this package a sub package to the root, this way we can be the source others import the code from

That's the question. Do we want to implement it in our project.. or do we want to implement this particular functionality on ocm-sdk-go?

pkg/ocm/config.go
georgettica: Yeah, I don't think we need it

shibumi: Ok, will remove all unnecessary stuff tomorrow.

rafael-azevedo:
@shibumi I've looked over the code and want to discuss. Lets have a quick sync when we are both back online tomorrow.

georgettica:
Just for context, this has been extracted from the ocm-cli code (which we were told not to use)

shibumi:
@ramonbutter pointed me to this code here: https://github.com/openshift-online/ocm-sdk-go/blob/ae3d77d559729baff1758859a7a104d2ffc70069/examples/load_config.go#L48

So loading configs should work with ocm-sdk-go alone as well.. we can probably drop all code and just use this instead + some logic for file reading etc

shibumi:
Tested this with modifying the clustermissing command:
```go
// ClusterMissingCmd represents the cluster-missing command
var ClusterMissingCmd = &cobra.Command{
	Use:   "cluster-missing",

		// in this case it's ok if the envvar is empty
		CAD_OCM_FILE_PATH := os.Getenv("CAD_OCM_FILE_PATH")
		_, err := ocm.New(CAD_OCM_FILE_PATH)
		if err != nil {
			return fmt.Errorf("could not create ocm client: %w", err)
		}

		return nil
	},
}
```

The OCM configuration gets loaded correctly. I think this PR is ready to go @georgettica, but maybe @rafael-azevedo  or @ramonbutter want to have a final look :)

georgettica:
/lgtm
/approve
/hold
As you wanted second 👀

pkg/ocm/ocm.go
```diff
@@ -23,24 +22,26 @@ var (
...
+			return client, err
+		}
 	}
-	client.cfg = cfg
 
-	conn, err := client.cfg.Connection()
+	client.conn, err = sdk.NewConnectionBuilder().Load(ocmConfigFile).Build()
 	if err != nil {
-		return client, fmt.Errorf("can't create connection: %w", err)
+		return client, fmt.Errorf("failed to load config file: %w", err)
```
shibumi, line 42: Do we want to change this error message? Failing to load the config is just one possible errors.. others can be problems with establishing the connection and more.

rafael-azevedo: I think a more generic error here would be better for sure.

rafael-azevedo:
@shibumi PR looks great. I want to see that error message improved before its approved

shibumi:
@rafael-azevedo fixed the error message for you. Feel free to merge :)

georgettica:
Not sure this solves what @rafael-azevedo 
But
/lgtm
/hold

shibumi:
@rafael-azevedo  feel free to merge if it's okay

PR #38, opened 2022-04-07
feat(cmd): annotate alert and send SL

shibumi:
This PR aims for [OSD-10779](https://issues.redhat.com/browse/OSD-10779)


cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -55,17 +56,34 @@ var ClusterMissingCmd = &cobra.Command{
...
+			return fmt.Errorf("could not verify all instances were running: %w", err)
+		}
+		fmt.Printf("yay! the result is %#v\n", res)
+
+		if res {
+			fmt.Println("The node shutdown was not the customer. Nothing to do.")
+			return nil
+		}
+
+		err = pd.AddNote(incidentID, "TODO_NOTE_CONTENT")
```
georgettica, line 75: some format I have in another script (can be modified)
```
PD_CHGM_NOTE='User-initiated shutdown of nodes in the cluster. Servicelog posted. Silencing alert.'
```
your flow doesn't only do the PD note, it does more.
I guess you can pull some of the logic out of my PR (how I did the client nesting and such) so you can do tyour whole flow in one func

shibumi: In our Ticket @rafael-azevedo  specified that we should attach the full service log on the alert. Do you know how I get that service log? I see that it's implemented in the ocm package, but we have no function yet to retrieve it :thinking:

georgettica: he meant to use the ocm command but:
- instead of pulling from GH, have it saved locally
- remove the code that pulls from GH (as now it will exist inside the codebase

cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -55,17 +56,33 @@ var ClusterMissingCmd = &cobra.Command{
...
 			return fmt.Errorf("could not start pagerdutyClient: %w", err)
 		}
 
+		res, err := chgm.AreAllInstancesRunning(externalClusterID)
+		if err != nil {
+			return fmt.Errorf("could not verify all instances were running: %w", err)
+		}
+
+		if res {
+			fmt.Println("The node shutdown was not the customer. Nothing to do.")
```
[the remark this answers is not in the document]
georgettica: Yeah, I think if this happens this is an odd alert, as everything is fine on the cluster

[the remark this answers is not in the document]
ramonbutter: hmm, I think we have to have a look over the definition of the function AreAllInstanceRunning. 
I would make the return value more focused on what we need at that point. And that is the answer to the question, where the CHGM is our fault or the fault of the customer
Meaning rather defining:
`isCHGMcustomerCaused(externalClusterID) bool`

cadctl/cmd/cluster-missing/cluster-missing.go
shibumi: Well, I can tackle this in my PR, too.

shibumi: I tackled this in a separate PR, because i think it's cleaner: https://github.com/openshift/configuration-anomaly-detection/pull/39

pkg/ocm/ocm.go
```diff
@@ -118,38 +118,40 @@ func (client Client) getClusterResource(clusterID string, resourceKey string) (s
...
 	}
 	return client.sendServiceLog(le, cluster)
 }
 
-// sendServiceLog allows to send a generic servicelog to a cluster
-func (client Client) sendServiceLog(le *servicelog.LogEntry, cluster *v1.Cluster) error {
+// sendServiceLog allows to send a generic servicelog to a cluster.
+// On success it will return the sent service log entry for further processing.
+func (client Client) sendServiceLog(le *servicelog.LogEntry, cluster *v1.Cluster) (*servicelog.LogEntry, error) {
 	builder := servicelog.NewLogEntry()
```
rafael-azevedo, line 138: This needs to be updated to our new logic with the struct and its methods right ? instead of using the builder directly.

rafael-azevedo, line 138: This needs to be updated to our new logic with the struct and its methods right ? instead of using the builder directly.

shibumi: This is correct. The PR hasn't been rebased yet.

georgettica:
100% 💚

shibumi:
I have rebased

PR #39, opened 2022-04-07
fix(ocm): embed service log for chgm

shibumi:
This PR embeds the Service Log into the cadctl binary


georgettica (approved):
/lgtm
/approve

shibumi:
@rafael-azevedo I talked with Ramon and we really came to the conclusion that embedding is not necessary. There is no vulnerability if we just set the SL URL as constant variable: https://securego.io/docs/rules/g107.html
I would argue, that we can trust GitHub as endpoint (if we cannot trust GitHub we shouldn't host our code there either :P)

georgettica:
I think another reason is it simplifies the code

shibumi:
@georgettica  it does remove one function if we embed it, but it also adds toil to us. If  the service log ever changes we have to change this at two different sources + have to update the code + have to deploy the new version of cadctl with the new embedded service log.

Considering all of this, I really think embedding the URL and changing it to constant should prevent Server-Side Request Forgery (SSRF) attacks and reduces toil for us.

ramonbutter:
hardcode one SL as a struct in the code may be ok. But what's about when we have to serve different SL in the future? Then it will become problematic and synchronization toil between managed notification and the code increases. In the long run, we would most probably revert this

pkg/ocm/ocm.go
```diff
@@ -17,8 +16,14 @@ import (
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 )
 
-var (
-	sl_clusterHasGoneMissing = "https://raw.githubusercontent.com/openshift/managed-notifications/master/osd/cluster_has_gone_missing.json"
+const (
```
rafael-azevedo, line 19: can we put this in a struct actually ?

shibumi: We can do a const out of this, but a struct cannot be const as far as I know :thinking: 
Or is `struct` a typo? :D

rafael-azevedo: Hmm that is true we would need a helper function to create the struct. Maybe you can add NewClusterMissingSL function

shibumi: I don't understand the benefit of having a struct there, yet. Can you explain this?
If we make a struct out of this the struct will be mutable again and then we will have the G107 security alert again, because the URL field will be not a const anymore.


Maybe we should decide first if we want to embed or not embed. This PR is embedding the URL, but embedding it comes with a price mentioned above.

shibumi:
@rafael-azevedo  reminder that we should test  this!

rafael-azevedo:
LGTM , I just want it to be tested @ramonbutter or @georgettica  Could you help Christian test this monday.

shibumi:
Update for @ramonbutter  @georgettica:

@rafael-azevedo  and me decided to embed it via the service log builder, that we do not need to marshal at all and can use generic structs to feed the builder.

shibumi:
@rafael-azevedo  we have found two issues with the ocm-sdk-go:

1. Our current ocm config load implementation does not work. (We never tested this properly it seems). The ocm-sdk-go code loads a different ocm config (YAML + different format) than us (JSON + the default one generated by ocm).
2. The Severity levels in the ocm-sdk-go are wrong. This has already been reported upstream and will be fixed soon without a specific ETA.

georgettica:
as all this waited for was a test.
we tested and this works
/lgtm
/approve

PR #41, opened 2022-04-11
fix(ocm): pull in OCM config file handling again

shibumi:
This commit pulls in the OCM config file handling again from the
ocm-cli repository. We do this, because the ocm-sdk-go provides
no way to load a JSON config file. They only handle YAML files.

### DIFF
--- a/pkg/ocm/ocm_config.go
+++ b/pkg/ocm/ocm_config.go
@@ -39,7 +39,6 @@ func Load() (cfg *Config, err error) {
 	_, err = os.Stat(file)
 	if os.IsNotExist(err) {
 		cfg = &Config{}
-		err = nil
 		return
 	}
 	if err != nil {
--- a/pkg/ocm/ocm_config.go
+++ b/pkg/ocm/ocm_config.go
@@ -7,11 +7,7 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/golang/glog"
-	"github.com/mitchellh/go-homedir"
 	sdk "github.com/openshift-online/ocm-sdk-go"
-	"github.com/openshift/rosa/pkg/debug"
-	"github.com/openshift/rosa/pkg/info"
 )
 
 // Config is the type used to store the configuration of the client.
@@ -77,7 +73,7 @@ func Location() (path string, err error) {
 	}
 
 	// Determine home directory to use for the legacy file path
-	home, err := homedir.Dir()
+	home, err := os.UserHomeDir()
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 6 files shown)


shibumi:
/hold

because I would like to trim this down a little bit. We don't need many of the dependencies

shibumi:
@rafael-azevedo I have tested this PR today. This PR can be merged. The config file gets loaded accordingly now, sending service logs to a cluster works as well. I used the following snippet to test it in the cluster-missing.go:
```go
		ocm, err := GetOCMClient()
		if err != nil {
			return fmt.Errorf("could not create ocm client: %w", err)
		}

		cluster, err := ocm.GetClusterInfo(incidentID)
		if err != nil {
			return fmt.Errorf("error: %w", err)
		}

		err = ocm.SendCHGMServiceLog(cluster)
		if err != nil {
			return err
		}
```

pkg/ocm/ocm_config.go
```diff
@@ -0,0 +1,136 @@
...
+	"os"
+	"path/filepath"
+
+	sdk "github.com/openshift-online/ocm-sdk-go"
+)
+
+// Config is the type used to store the configuration of the client.
+// There's no way to line-split or predefine tags, so...
+//nolint:lll
+type Config struct {
```
ramonbutter, line 16: are you sure we need this?

pkg/ocm/ocm_config.go
```diff
@@ -0,0 +1,136 @@
...
+	Scopes       []string `json:"scopes,omitempty" doc:"OpenID scope. If this option is used it will replace completely the default scopes. Can be repeated multiple times to specify multiple scopes."`
+	TokenURL     string   `json:"token_url,omitempty" doc:"OpenID token URL."`
+	URL          string   `json:"url,omitempty" doc:"URL of the API gateway. The value can be the complete URL or an alias. The valid aliases are 'production', 'staging' and 'integration'."`
+	User         string   `json:"user,omitempty" doc:"User name."`
+	Pager        string   `json:"pager,omitempty" doc:"Pager command, for example 'less'. If empty no pager will be used."`
+}
+
+// Load loads the configuration from the configuration file. If the configuration file doesn't exist
+// it will return an empty configuration object.
+func Load() (cfg *Config, err error) {
```
ramonbutter, line 34: Can we not use the Load function from ocm-cli and remove this part, as well as the the two function below? Sothat we at least are in sync with what ocm-cli expects and everyone uses as a config? I mean the PR looks way bigger than what was removed earlier in: https://github.com/openshift/configuration-anomaly-detection/commit/52bd34d9b67ad8c418d3c069b935dfe9d782d571

shibumi: Yes, this is intended. @rafael-azevedo said that he does not want to rely on ocm-cli, because ocm-cli was never intended for library usage. This is why I pulled in the changes manually, this way we are in the position to modify it and maintain it.

We can just use the ocm-cli library, but then we will pull in the nil pointer problem from ocm-cli + we will adopt any chanegs ocm-cli will do. The ocm-cli devs were pretty clear about this that we should not use their code as library

pkg/ocm/ocm_config.go
shibumi: Unfortunately yes, if we want to pull in the ocm-cli implementation as static code in our source.

PR #43, opened 2022-04-13
fix(Dockerfile): build executable instead of a lib

shibumi:
This fixes the build issues in the final container image. The root cause is our security flags that do not work when the binary has been compiled statically. They heavily rely on libc (something we don't want in our containers).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ bin/mockgen:
 	GOBIN=$(PWD)/bin go install -mod=readonly github.com/golang/mock/mockgen@v1.6.0
 
 cadctl/cadctl: cadctl/**/*.go pkg/**/*.go go.mod go.sum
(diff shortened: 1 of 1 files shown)


shibumi:
@rafael-azevedo  @ramonbutter  @georgettica so the problem is that the security hardening flags only work when the binary is linked dynamically. If we use a static compiled binary like the one in the image.. the binary will get compiled properly, but not linked correctly. 

This is why we saw the "no file or directory" dialog, because  Linux tried to execute the library as an executable.

shibumi:
@rafael-azevedo  that's some issue with the CI system being too slow.. we can increase the timeout in the linter: 
```
GOLANGCI_LINT_CACHE=/tmp/tmp.83P37wxQ7W ./bin/golangci-lint run
level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
```

PR #45, opened 2022-04-14
Updated delpoyment for addtional variables

rafael-azevedo:
Updated to add new variables in task-checks yaml


deploy/task-cad-checks.yaml
```diff
@@ -33,14 +33,19 @@ spec:
         secretKeyRef:
           name: cad-pd-token
           key: token
-    - name: CAD_JUMPROLE
```
shibumi, line 36: Aren't there environment variables missing? Policies etc?

PR #47, opened 2022-04-20
feat(ocm): add ocm client ID/secret support

shibumi:
This small PR adds support for OCM client ID/secret. See also: [OSD-11015](https://issues.redhat.com/browse/OSD-11015)

Note: The original issue speaks about tokens, but token is the wrong term here. We have to use a client ID + client secret


georgettica (approved):
/lgtm
/Approve
/Hold

deploy/task-cad-checks-client-secrets-ocm.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: cad-ocm-client-secret
+  namespace: configuration-anomaly-detection
+type: Opaque
+stringData:
```
georgettica, line 8: Cool!

pkg/ocm/ocm.go
```diff
@@ -43,14 +43,27 @@ func New(ocmConfigFile string) (Client, error) {
...
+		client.conn, err = cfg.Connection()
+		if err != nil {
+			return client, fmt.Errorf("failed to create new OCM connection from config file: %w", err)
+		}
+
+		return client, nil
 	}
 
-	client.conn, err = cfg.Connection()
+	client.conn, err = sdk.NewConnectionBuilder().URL(ocmURL).Client(ocmClientID, ocmClientSecret).Insecure(true).Build()
```
georgettica, line 64: Why insecure true?

pkg/ocm/ocm.go
```diff
@@ -43,14 +43,27 @@ func New(ocmConfigFile string) (Client, error) {
 	var err error
 	client := Client{}
 
-	cfg, err := newConfigFromFile(ocmConfigFile)
-	if err != nil {
-		return client, fmt.Errorf("failed to load config file: %w", err)
+	ocmClientID, hasOcmClientID := os.LookupEnv("OCM_CLIENT_ID")
```
georgettica, line 46: I chose to omit the prefix of `CAD` on the other envvars as we had a client that used the previous envvars.

If we are unique on using this we should prefix with a CAD to signify this is unique to this code.

Once we decide this will be generally used we can omit the prefix and push upstream

georgettica:
We might want to card for later the splitting of the ocm login into differey

georgettica:
Sorry, bit out for a while and I found lots of things I could nit.
I hope it's not too much / opposite to my lgtm + approve before

pkg/ocm/ocm.go
shibumi: That's a suggestion from the Slack Channel :thinking: I didn't ask for a reason. Are the OCM certificates self-signed? If not, we can set Insecure to False

deploy/task-cad-checks-client-secrets-ocm.yaml
```diff
@@ -0,0 +1,11 @@
...
+  OCM_CLIENT_ID: ocm-cad-staging
```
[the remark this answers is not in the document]
shibumi: Would you prefer reading these secrets from a file? :thinking: 
Or what  is your point here?

pkg/ocm/ocm.go
shibumi: I agree.. naming the env vars the same like in the vault makes sense

shibumi:
@georgettica no problem. I appreciate your comments :D

pkg/ocm/ocm.go
georgettica: Makes sense, should we card this for later aswell?
If we can write on each possible security hole the reasoning, it makes auditing focus on things that we can really fix

shibumi: I have checked api.openshift.com and api.stage.openshift.com.. both have valid certificates. I think we can set insecure to false

deploy/task-cad-checks-client-secrets-ocm.yaml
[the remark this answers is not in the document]
georgettica: No, I just think having the hardcoded clientid leaks some private data we have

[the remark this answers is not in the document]
shibumi: Done.. the history will be lost about if we squash this PR

pkg/ocm/ocm.go
```diff
@@ -43,19 +44,63 @@ func New(ocmConfigFile string) (Client, error) {
 	var err error
 	client := Client{}
 
-	cfg, err := newConfigFromFile(ocmConfigFile)
-	if err != nil {
-		return client, fmt.Errorf("failed to load config file: %w", err)
+	// The debug environment variable ensures that we will never use
+	// a ocm config file on a cluster deployment. The debug environment variable
+	// is only for local cadctl development
+	debugValue, hasDebug := os.LookupEnv("CAD_DEBUG")
```
rafael-azevedo, line 50:
```suggestion
	debugValue, ok:= os.LookupEnv("CAD_DEBUG")
	if !ok {
	debugValue = false
	}
```

pkg/ocm/ocm.go
```diff
@@ -43,19 +44,63 @@ func New(ocmConfigFile string) (Client, error) {
...
+
+	// verify via regex if the CAD_DEBUG env got set, but disabled.
+	// TODO: discuss if we really need this.
+	debugDisabled, err := regexp.MatchString("false|False|0|no|No", debugValue)
```
rafael-azevedo, line 54: WE can get rid of this

pkg/ocm/ocm.go
```diff
@@ -43,19 +44,63 @@ func New(ocmConfigFile string) (Client, error) {
...
+
+	if hasDebug && !debugDisabled {
```
rafael-azevedo, line 56:
```suggestion
	if debugValue == "true" 
```

pkg/ocm/ocm.go
shibumi: I see, but we also have to remove the regex check then, right? Do you want to edit your suggestion or shall I just implement it?

pkg/ocm/ocm.go
shibumi: Oh okay. Good

pkg/ocm/ocm.go
```diff
@@ -43,19 +43,59 @@ func New(ocmConfigFile string) (Client, error) {
...
 
-	cfg, err := newConfigFromFile(ocmConfigFile)
-	if err != nil {
-		return client, fmt.Errorf("failed to load config file: %w", err)
+	// The debug environment variable ensures that we will never use
+	// a ocm config file on a cluster deployment. The debug environment variable
+	// is only for local cadctl development
+	debugMode := os.Getenv("CAD_DEBUG")
+
+	if debugMode == "true" {
```
georgettica, line 51: Maybe parse this better? (To an actual boolean)

georgettica:
We can later export it of we need it

PR #49, opened 2022-04-26
feat(chgm): Make notes/logging human readable for CAD events

ninabauer:
When we send notes into incidents we are currently dumping out a lot of json data that is not easy to digest.

This PR will make the notes human readable.

See playground link for example of output.

[OSD-11014](https://issues.redhat.com//browse/OSD-11014)

https://go.dev/play/p/2h3BMITpdAk


pkg/services/chgm/chgm.go
```diff
@@ -116,17 +115,26 @@ type UserInfo struct {
 type InvestigateInstancesOutput struct {
 	NonRunningInstances []*ec2.Instance
 	User                UserInfo
-	StopTime            time.Time
```
georgettica, line 119: stoptime was not used in the codebase, thus removed

PR #50, opened 2022-04-27
Fixed nil return

rafael-azevedo:
This PR replaces a nil return with an empty object to prevent run time nil pointer errors.


georgettica:
this doesn't matter. as we don't care about that return value.
the only thing is it takes more characters IMO

rafael-azevedo:
@georgettica Since we are still returning the value . I think its better to return it correctly here to ensure should anyone else use this function in the future that it doesn't potentially cause that nil pointer.

PR #51, opened 2022-04-28
fix(deploy): based on tekton change

georgettica:
also fix README.md with specific steps

this is used to make sure CAD will deploy correctly


georgettica:
related to OSD-11297

georgettica:
@shibumi during the  f2f, you suggested moving the secret values to a dedicated file, do you have by any change how to fill these values?

PR #53, opened 2022-05-03
feat(hack): generate template file

georgettica:
now we can
- use 'make generate-template-file' to generate it
- all commands of 'make generate' will do all of the generations needed


hack/configuration-anomaly-detection-template.Template.yaml
```diff
@@ -0,0 +1,136 @@
+apiVersion: v1
+kind: Template
+parameters: []
+metadata:
+  name: configuration-anomaly-detection-template
+objects:
+- apiVersion: v1
```
shibumi, line 7: You can leave that namespace in the template, but the pipeline will ignore it. Namespaces are managed by App-SRE

hack/update-template/main.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	// marshal back into bytes
+	saasTemplateFileAsBytes, marshalBackErr := yaml.Marshal(&saasTemplateFile)
+	if marshalBackErr != nil {
+		panic(fmt.Errorf("could not marshal the bigMapping back: %v", marshalBackErr))
+	}
+	// print to stdout just so we know something happened and the code isn't broken
+	fmt.Printf("---\n%s\n---\n", string(saasTemplateFileAsBytes))
+	outputFile := filepath.Join(workingDirectory, "../configuration-anomaly-detection-template.Template.yaml")
```
shibumi, line 95: The target directory for the created file should be openshift/template.yaml (that's what all projects are doing). We can change the path in the app-sre deployment spec, but then it will be different to all other projects :thinking: 

I  don't have a clear opinion about this.

hack/update-template/main.go
```diff
@@ -0,0 +1,115 @@
...
+		saasTemplateFile.Objects = append(saasTemplateFile.Objects, fileAsMap)
+	}
+
+	// marshal back into bytes
+	saasTemplateFileAsBytes, marshalBackErr := yaml.Marshal(&saasTemplateFile)
+	if marshalBackErr != nil {
+		panic(fmt.Errorf("could not marshal the bigMapping back: %v", marshalBackErr))
+	}
+	// print to stdout just so we know something happened and the code isn't broken
+	fmt.Printf("---\n%s\n---\n", string(saasTemplateFileAsBytes))
```
shibumi, line 94: I think this can be shortened to: `---\n%s\n`
YAML has just "start" parameter and no end parameter as far as I know

hack/update-template/main.go
georgettica: Give the path and I will do it :)

georgettica:
I'll check the validate tomorrow

hack/update-template/main.go
shibumi: See my other PR for the path.. it's really just <project root>/openshift/template.yaml as specified here: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/38300/diffs#51d44de8da2934b1e8168e02e247073fe65f428f_0_41

Line 41

hack/update-template/main.go
shibumi: I don't see the "---" in your generated template file at all :thinking:  are you sure that your code works? do you want to add some tests to it?

georgettica: It's not required :/
I just posted the yaml files.
I will remove the --- from the print.

Testing might be odd here as I don't know what I am checking.

This was supposed to be a bit better than a big bash script/python code

shibumi: True, Go is the better solution for yes :)

I just didn't know that the "---"s  are not necessary :thinking: I thought we need them to differ between files/resources.
Did you try executing the YAML on a cluster? did it work?

georgettica: No, but feel free to check

maorfr:
the openshift template should assume it is deploying only what constitutes the "service", and not a full from-zero-to-hero bundle.

hence, you can remove the Namespace, the Subscription, etc.

maorfr:
created [APPSRE-5497](https://issues.redhat.com/browse/APPSRE-5497) to take a deep dive

georgettica:
might be related to APPSRE-5497

openshift/template.yaml
```diff
@@ -0,0 +1,251 @@
+apiVersion: v1
+kind: Template
```
shibumi, line 2: Is this an Openshift API resource? :thinking:  never heard of template on Kubernetes. If yes, we might have to modify my saas file PR on gitlab, because I didn't specify that API resource there yet.

georgettica: nope, no need. this is the file they process.

shibumi:
feel free to merge

georgettica:
I need to test the template

openshift/template.yaml
```diff
@@ -0,0 +1,251 @@
...
+      - name: CSS_JUMPROLE
+        value: XXXX
+      - name: SUPPORT_JUMPROLE
+        value: XXXX
+      envFrom:
+      - secretRef:
+          name: cad-aws-credentials
+      - secretRef:
+          name: cad-ocm-client-secret
+      image: quay.io/app-sre/cadctl@sha256:22d2a957d935e883f45addd35acc87450ce71ea2f94c33f2df70ff36b015486a
```
maorfr, line 250: this should be parameterized

openshift/template.yaml
```diff
@@ -0,0 +1,251 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: configuration-anomaly-detection-template
+parameters: []
+objects:
+- apiVersion: tekton.dev/v1beta1
+  kind: PipelineRun
```
maorfr, line 8: what does this do?

openshift/template.yaml
georgettica: yes, but to which value?

openshift/template.yaml
georgettica: if you mean the PipelineRun. it's for testing purposes, we should move it out as well (as it causes and error when you apply it)

georgettica:
errors I currently see when I run the template by a normal backplane user
```
Error from server (Forbidden): error when creating "STDIN": roles.rbac.authorization.k8s.io "cad-eventlistener-role" is forbidden: user "$CURRENT_USER" (groups=["system:serviceaccounts" "${CURRENT_GROUP}" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
{APIGroups:["*"], Resources:["configmaps"], Verbs:["list" "watch" "get"]}
{APIGroups:["*"], Resources:["secrets"], Verbs:["list" "watch" "get"]}
{APIGroups:["*"], Resources:["serviceaccounts"], Verbs:["list" "watch" "get"]}
{APIGroups:["tekton.dev"], Resources:["*"], Verbs:["*"]}
{APIGroups:["triggers.tekton.dev"], Resources:["*"], Verbs:["*"]}
Error from server (Forbidden): error when creating "STDIN": clusterroles.rbac.authorization.k8s.io is forbidden: User "$CURRENT_USER" cannot create resource "clusterroles" in API group "rbac.authorization.k8s.io" at the cluster scope
Error from server (NotFound): error when creating "STDIN": roles.rbac.authorization.k8s.io "cad-eventlistener-role" not found
Error from server (Forbidden): error when creating "STDIN": clusterrolebindings.rbac.authorization.k8s.io is forbidden: User "$CURRENT_USER" cannot create resource "clusterrolebindings" in API group "rbac.authorization.k8s.io" at the cluster scope
```

georgettica:
@maorfr I need to know the namespace.
when I run this locally using the admin perms, all passes aside of
```
Error from server (Invalid): error when creating "STDIN": ClusterRoleBinding.rbac.authorization.k8s.io "cad-eventlistener-clusterrolebinding" is invalid: subjects[0].namespace: Required value
```

openshift/template.yaml
maorfr: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/continuous-delivery-in-app-interface.md#automatically-generated-parameters

maorfr:
add a parameter of `NAMESPACE_NAME` with a default value of `configuration-anomaly-detection` and override it from the saas file to the right name?

openshift/template.yaml
shibumi: This is a test PipelineRun and should not be included.

openshift/template.yaml
shibumi: Ah yes, I forgot about this one. We want to parametrize this and very likely we have to change the names for the secrets as well

georgettica:
please review, or at least tell me in broad strokes what you think of this

deploy/serviceaccount.yaml
```diff
@@ -69,7 +69,6 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
   name: cad-eventlistener-role
-  namespace: configuration-anomaly-detection
```
fahlmant, line 72: Was this removal on purpose?

hack/update-template/main.go
```diff
@@ -0,0 +1,212 @@
...
+)
+
+func main() {
+	ex, getExecutableErr := os.Executable()
+	if getExecutableErr != nil {
+		panic(fmt.Errorf("could not get the Executable: %w", getExecutableErr))
+	}
+
+	workingDirectory := filepath.Dir(ex)
+	fmt.Printf("workingDirectory :: %s\n", workingDirectory)
```
fahlmant, line 45: Any reason not to do `os.Getcwd()` here?

hack/update-template/main.go
```diff
@@ -0,0 +1,212 @@
...
+
+	var filesNames []string
+	relativeDeployFolder := filepath.Join(workingDirectory, deployFolderPath)
```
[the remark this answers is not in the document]
georgettica: It's starting the search from the binaries folder, then traverses up

hack/update-template/main.go
georgettica: Yeah, it is where the caller is, not where the script is

hack/update-template/main.go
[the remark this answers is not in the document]
georgettica: These can be later parameterized if need be.
Let me know if that is necessary

georgettica:
I'll fix this tomorrow, I expected the code to work

hack/update-template/main.go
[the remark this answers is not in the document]
georgettica: No, it directs to the location of the binary

hack/update-template/main.go
```diff
@@ -0,0 +1,212 @@
...
+
+			removeNestedField(fileAsMap, "metadata", "namespace")
+
+			kind, kindExists := fileAsMap["kind"]
+			if !kindExists {
+				panic(fmt.Errorf("kind doesn't exist"))
+			}
+
+			apiVersion, apiVersionExists := fileAsMap["apiVersion"]
+			if !apiVersionExists {
```
fahlmant, line 107: This might be better rolled into a switch statement

openshift/template.yaml
georgettica: I will keep this up as @ninabauer might change the secrets in her PR

openshift/template.yaml
```diff
@@ -0,0 +1,242 @@
+apiVersion: template.openshift.io/v1
+kind: Template
+metadata:
+  name: configuration-anomaly-detection-template
+parameters:
+- name: IMAGE_DIGEST
+  value: sha256:22d2a957d935e883f45addd35acc87450ce71ea2f94c33f2df70ff36b015486a
```
shibumi, line 7: Question: These are only set when the parameter hasn't been set, correct?

openshift/template.yaml
```diff
@@ -0,0 +1,242 @@
...
+        value: XXXX
+      - name: CAD_SILENT_POLICY
+        value: XXXX
+      - name: CSS_JUMPROLE
+        value: XXXX
+      - name: SUPPORT_JUMPROLE
+        value: XXXX
+      envFrom:
+      - secretRef:
+          name: cad-aws-credentials
```
shibumi, line 238: I am not 100% sure, but these secret names might be different on the clusters. @maorfr  is there any way to verify the secrets on the app-sre clusters?

hack/update-template/unstructured-modification.go
```diff
@@ -0,0 +1,51 @@
...
+
+func nestedFieldNoCopy(obj map[interface{}]interface{}, fields ...string) (interface{}, bool, error) {
+	var val interface{} = obj
+
+	for _, field := range fields {
+		if val == nil {
+			return nil, false, nil
+		}
+		if m, ok := val.(map[interface{}]interface{}); ok {
+			val, ok = m[field]
```
shibumi, line 30: What's the purpose of vail if you don't use it inside of the loop? :thinking:

hack/update-template/unstructured-modification.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+	}
+	return val, true, nil
+}
+
+func removeNestedField(obj map[interface{}]interface{}, fields ...string) {
+	m := obj
+	for _, field := range fields[:len(fields)-1] {
+		if x, ok := m[field].(map[interface{}]interface{}); ok {
+			m = x
```
shibumi, line 45: Same here.. I don't understand how this loop works. Are you sure that the last line `delete...) shouldn't be in the loop?
Or would it make sense to break the loop if you've found the m?

hack/update-template/unstructured-modification.go
```diff
@@ -0,0 +1,51 @@
+package main
+
+// all of these have been pulled from k8s.io/apimachinery/pkg/apis/meta/v1/unstructured
+// this issue https://github.com/kubernetes/apimachinery/issues/138 explains why I needed to duplicate these funcs
```
shibumi, line 4: The issue doesn't really explain why you can't just import the package. The issue explains a problem `with` the package, but it doesn't explain what you've changed or what the package does different to the code in this file.

If this is the same code like in apimachinery... why do we copy it?

hack/update-template/unstructured-modification.go
shibumi: Ah okay never mind. It's about the **last valid field**! :D

hack/update-template/unstructured-modification.go
shibumi: Ah okay never mind. It's about the **last valid field**! :D

openshift/template.yaml
georgettica: yes

hack/update-template/unstructured-modification.go
georgettica: yeah, I will write it there aswell. good point

shibumi: That is much better

openshift/template.yaml
maorfr: see if this helps: https://gitlab.cee.redhat.com/service/app-interface#manage-secrets-via-app-interface-openshiftnamespace-1yml-using-vault

openshift/template.yaml
maorfr, line 7: consider removing `value` and adding `required: true` instead?

similar to https://github.com/openshift/aws-account-operator/blob/0dc92d1e582f8f381f9a212b2edc4725394e2cb6/hack/olm-registry/olm-artifacts-template.yaml#L11-L12

openshift/template.yaml
maorfr: just for my curiosity, what are these aws credentials?

openshift/template.yaml
georgettica: nah, I want the usual dev to have the correct image (and hopefully the correct one).
currently this image cannot pull the correct credentials, so it's ok I think

maorfr: so are you planning to maintain the image in this template forever?

georgettica: I was hoping I could, is that ask too much?
I can remove if it is

maorfr: i don't think it's feasible. you also can't know the digest before the image is pushed. and that only happens after a merge. :chicken: :egg:

maorfr: i think that would be better, and also consistent with other operators.

openshift/template.yaml
georgettica: These are generated using a an internal command to pull AWS creds.
For prod we have the AWS key and secret fields

openshift/template.yaml
georgettica: #60

PR #54, opened 2022-05-05
OSD-10193: added some more infos to contribution guide

ramonbutter:
This is a collection of information that can help to contribute and extend configuration-anomaly-detection

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,20 +1,59 @@
 # Contribution Guide
 
-## build and test
-to build the code you can use the
+CAD provides a terminal cli. It is a collection of functions to investigate potential cloud infrastructure configuration problems.
+Currently, CAD reacts to certain alerts that appear in Pagerduty.
+
+## Build
+
+To build the code, you can use the following target:
+
 ```shell
 make
 ```
-target.
 
-## verify error messages are unique
-in order to grep any part of an error message and see the exact location is was created in, use the following command:
+The binary is then available under `./cadctl/cadctl`
+
+## Testing
+
+Try to add as many unit tests as possible to the functions you are writing. Have a look into the other investigations about how the dependencies are mocked away. Think of the happy as well as unhappy paths.
+Keep the entry arguments for each investigation the same. The argument is a path to a JSON blob which should look similar to the one received by Pagerduty.
+
+```bash
+echo '{"event": {"data":{"id": "<pd-incident-id>"}}}' > ./payload
+cadctl cluster-missing --payload-path ./payload
+```
+
+NOTE: the JSON blob is reduced here to the important part that is analyzed by CAD atm. PD sends a larger JSON object.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,20 +1,58 @@
...
+CAD is in the end deployed on an OpenShift cluster. Different configurations are set depending on the environment it serves (in, stage, prod). However, when you design the investigation functions, keep in mind how the investigation will be runnable from the local machine, without the need to deploy them first into Kubernetes. This means that required configurations should be injected via env-variables to the CLI and not be fetched by f.e. a config map, and think of the dependent values.
+
+The dependencies so far are:
+
+- AWS
+- Pagerduty
+- OCM
+- Tekton
+  
+For testing an investigation, it may be sufficient to run CAD locally. A JSON blob with an incident ID from PD should be used as an argument. However, make sure you trigger the particular incident in a cluster.
```
georgettica, line 37: Do we have a hard openshift requirement?
Does OLM install seamlessly on k8s aswell and the `subscription` can be worked the same?

---
We do expose a route so maybe we are vendor locking a bit

CONTRIBUTING.md
```diff
@@ -1,20 +1,58 @@
...
+To test Tekton and the deployment configuration, put CAD on OpenShift behind a Tekton event-listener and use curl to trigger pipeline runs by using example payloads.
+
+## Coding Style
+
+Standard software engineering practices apply here:
+Please use small functions. Add comments when needed. Think about naming. Export reusable functionality. Use a linter. Try to apply a hexagonal architecture pattern. Add unit tests. Ask peers to review your PRs. Follow the [golang coding style](https://go.dev/doc/effective_go).
```
georgettica, line 43: I am not sure we are doing an hexagonal architecture

georgettica (commented):
Also a toil card we can do, make the error function also show the lines which have duplicate error messages

CONTRIBUTING.md
ramonbutter: what else? That pattern matched the most to what we did so far

georgettica: Ok,  good to know 😅
Unintended hexagonal developer

CONTRIBUTING.md
shibumi: Installing OLM in vanilla k8s should work, but routes won't work and you have to use an ingress instead.

georgettica: I guess we should caveat this at some point :)
Aside of that it's ok

georgettica:
@ramonbutter add description?

CONTRIBUTING.md
ramonbutter: I mean our CD pipelines are solely designed for OpenShift. However, I don't think it's difficult to make the setup work with k8s. As @shibumi pointed out the route needs to be replaced with ingress from the CRs side. But not too sure whether we actually need this right now in the CONTRIBUTING guide as we are anyway pretty much locked to our setup, meaning AWS STS clusters with 2 hops in the role chain. Cluster alerts propagate to PD. Maybe we can give more hints when there is a maintenance sprint, or?

georgettica:
we can merge this. but I wouldn't say this task is done :/

shibumi:
@georgettica  what is missing?

georgettica:
What we discussed offline, the input from a functional lead on what would it require to become a maintainer

rafael-azevedo:
> 

I will be adding that information in a seperate PR but I think this one is good to go

PR #55, opened 2022-05-06
Shibumi/create build deploy script for app sre

shibumi:


build_deploy.sh
```diff
@@ -0,0 +1,20 @@
...
+make test
+make validate
+
+IMAGE="quay.io/app-sre/configuration-anomaly-detection"
+IMAGE_TAG=$(git rev-parse --short=7 HEAD)
+
+docker build -t "${IMAGE_NAME}:latest" .
+docker tag "${IMAGE_NAME}:latest" "${IMAGE_NAME}:${IMAGE_TAG}"
+
+DOCKER_CONF="${PWD}/.docker"
```
georgettica, line 15: Not a blocker, but is this safer as it doesn't change the global docker config?

shibumi: Closed, because others told me we have to use openshift/boilerplate instead

PR #57, opened 2022-05-09
Shibumi/fix dockerfile location

shibumi:
This PR fixes our Dockerfile location


georgettica:
/lgtm
let's see this works correctly too

PR #58, opened 2022-05-09
fix: set redhat owned docker image

shibumi:
This sets the red hat owned base image for golang instead of the docker hub alpine one

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
-FROM golang:1.17.7-alpine3.15 as builder
-
-RUN apk add --update --no-cache ca-certificates tzdata git make bash && update-ca-certificates
+ARG BUILDER_IMG=registry.ci.openshift.org/openshift/release:golang-1.17
+FROM $BUILDER_IMG as builder
 
 ADD . /opt
 WORKDIR /opt


Dockerfile
```diff
@@ -1,6 +1,4 @@
-FROM golang:1.17.7-alpine3.15 as builder
-
-RUN apk add --update --no-cache ca-certificates tzdata git make bash && update-ca-certificates
+FROM registry.ci.openshift.org/openshift/release:golang-1.17 as builder
```
georgettica, line 1: we can add a ARG so the image can be build by other images, so if anyone wants to override the registry image, they have an alternative

shibumi: ARG has no way to pass a default option and passing the image name during the build process would mean that we have to modify the boilerplate :/

shibumi: I didn't know this. Thank you!

shibumi: I didn't know this. Thank you!

shibumi: @georgettica fixed

georgettica (commented):
/approve
/lgtm

georgettica:
/lgtm
/approve
as I didn't see I didn't do it

PR #60, opened 2022-05-10
chore(template): now force digest

georgettica:
this PR makes the code be more like other operators, and fixes the image url that is used


georgettica:
related to OSD-11381

maorfr:
LGTM

PR #62, opened 2022-05-11
fix: hardcore algorithm into image digest

shibumi:
The APP-SRE pipeline is not setting the algorithm in the image_digest. This is tracked in: https://issues.redhat.com/browse/APPSRE-5547

let us use a workaround for now


georgettica (approved):
/lgtm
/approve
Sad it's the case but ok!

shibumi:
closing, because we do it differently now. Instead of using image_digests we move back to image_tags + add another script layer that builds the image once and then retags it with the tags app-sre needs.

PR #63, opened 2022-05-12
fix: retag images for app-sre pipeline

shibumi:
This commit introduces a build_deploy.sh script that triggers the boilerplate
build, retags the image and pushes the image with a new tag

### DIFF
--- a/build_deploy.sh
+++ b/build_deploy.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+make osd-container-image-build-push
+
+# Generate version and tag information from inputs
+COMMIT_NUMBER=$(git rev-list `git rev-list --parents HEAD | egrep "^[a-f0-9]{40}$$"`..HEAD --count)
+CURRENT_COMMIT=$(git rev-parse --short=7 HEAD)
+IMAGE_TAG="v0.1.${COMMIT_NUMBER}-${CURRENT_COMMIT}"
+
+docker tag quay.io/app-sre/configuration-anomaly-detection:${IMAGE_TAG} quay.io/app-sre/configuration-anomaly-detection:${CURRENT_COMMIT}
+docker push quay.io/app-sre/configuration-anomaly-detection:${CURRENT_COMMIT}


openshift/template.yaml
```diff
@@ -233,5 +233,5 @@ objects:
           name: cad-ocm-client-secret
       - secretRef:
           name: cad-pd-token
-      image: ${REGISTRY_IMG}@${IMAGE_DIGEST}
+      image: ${REGISTRY_IMG}:${IMAGE_TAG}
```
maorfr, line 236: should be replaced in the `parameters` section as well

build_deploy.sh
```diff
@@ -0,0 +1,11 @@
...
+
+make osd-container-image-build-push
+
+# Generate version and tag information from inputs
+COMMIT_NUMBER=$(git rev-list `git rev-list --parents HEAD | egrep "^[a-f0-9]{40}$$"`..HEAD --count)
+CURRENT_COMMIT=$(git rev-parse --short=7 HEAD)
+IMAGE_TAG="v0.1.${COMMIT_NUMBER}-${CURRENT_COMMIT}"
+
+docker tag quay.io/app-sre/configuration-anomaly-detection:${IMAGE_TAG} quay.io/app-sre/configuration-anomaly-detection:${CURRENT_COMMIT}
+docker push quay.io/app-sre/configuration-anomaly-detection:${CURRENT_COMMIT}
```
maorfr, line 11: you should handle authentication before pushing.

this will include:
- creating a docker config file and performing a login with it. example: https://github.com/app-sre/qontract-server/blob/19a6b118c9b0d9444e405f4c5135bfce50319e09/build_deploy.sh#L4-L6
- pushing while specifying the config file. example: https://github.com/app-sre/qontract-server/blob/19a6b118c9b0d9444e405f4c5135bfce50319e09/Makefile#L61-L63

shibumi:
@maorfr  I wonder if we can just include this into the `project.mk` file:
```
IMAGE_REGISTRY?=quay.io
IMAGE_REPOSITORY?=app-sre
IMAGE_NAME?=configuration-anomaly-detection
DOCKERFILE?=./Dockerfile
define ADDITIONAL_IMAGE_SPECS
Dockerfile $(IMAGE_REGISTRY)/$(IMAGE_REPOSITORY)/$(IMAGE_NAME):$(CURRENT_COMMIT)
endef
```

This way I don't need the script, but I am not sure if it would work. That's also what @NautiluX  has suggested

maorfr:
it would mean an image is built twice. since we will only use the additional image, i don't feel there will be issues with rebuilding. please feel free to choose!

openshift/template.yaml
shibumi: fixed

shibumi:
I decided to try it like @NautiluX  has suggested... I hope that works, I am not a makefile magician :D feel free to have a look @georgettica  or @NautiluX

PR #66, opened 2022-05-16
chore: add more approvers

shibumi:
Add more approvers for faster development cycles

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,4 +1,5 @@
 reviewers:
+- Nikokolas3270
 - Tessg22
 - georgettica
 - ninabauer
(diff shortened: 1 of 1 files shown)


shibumi:
Interesting.. so the auto merge does not work yet. I would have assumed that giving the app-sre-bot write access to this repository would allow auto-merging :thinking:

georgettica:
So I would say it might be because of our prow config is missing something

shibumi:
@georgettica  yes, you are correct. I assumed that the app-sre-bot would be responsible for this, but turns out it's the prow Ci bot

shibumi:
@georgettica  we can copy this behavior from here: https://github.com/openshift/release/tree/master/ci-operator/config/openshift/aws-account-operator

PR #67, opened 2022-05-17
feat(deploy): minimize RBAC footprint

georgettica:
currently CAD has a clusterrole && clusterrolebinding that handle the
clusterinterceptors. This permission was requested by the infra team to
be removed from the deployment.

as such, we minimized as much as we could from the deployment and came
up with the following changes in this patchset:
- remove clusterrole (DEPENDENCY ADDED FOR tekton clusterrole)
- minimize role permissions
- attempt to change clusterrolebinding to clusterrole (failed)

unrelated but done here anyways:
- prettify deploy/task-cad-checks.yaml : while using `yq`-foo, the repo
  will only show changes in the line needed, and not whitespace fixes
- move one of the resources from v1alpha1 -> v1beta1 : was supposed to
  fix a deploy bug but I am not sure it did.


georgettica:
related to OSD-11324

Makefile
```diff
@@ -77,12 +77,6 @@ generate-markdown: $(MARKDOWN_SOURCES) bin/embedmd
 
 ## CI actions 
 
-# pulled from https://github.com/openshift/boilerplate/blob/056cba90733136e589ac2c4cd45238fd6207cfbd/Makefile#L10-L11
```
shibumi, line 80: When was this created? did we created it manually or via the boilerplate?

shibumi:
just added a question about the `isclean` target

deploy/task-cad-checks.yaml
```diff
@@ -5,34 +5,34 @@ metadata:
...
-    envFrom:
-    - secretRef:
-        name: cad-aws-credentials
-    - secretRef:
-        name: cad-ocm-client-secret
-    - secretRef:
-        name: cad-pd-token
+    - name: check-infrastructure
+      # this sha is broken, should fix when we onboard
+      image: quay.io/app-sre/cadctl@sha256:22d2a957d935e883f45addd35acc87450ce71ea2f94c33f2df70ff36b015486a
```
shibumi, line 14: Oh one thing: Do we want to set the correct image name here? the image is located at: quay.io/app-sre/configuration-anomaly-detection

patjlm: and the tag should be used instead of the sha i guess

openshift/template.yaml
```diff
@@ -40,7 +40,6 @@ objects:
           value: $(tt.params.payload)
         pipelineRef:
           name: cad-checks-pipeline
-        serviceAccountName: cad-sa
```
patjlm, line 43: which SA will be used here ? If it ever needs a specific RBAC, it would be good to keep a dedicated one. I don't think so, but just checking

patjlm: a quick test showed that the `pipeline` serviceaccount would be used by default. I don't think your pipeline has any need on the local cluster by default, so I'd use a custom SA there maybe.

deploy/task-cad-checks.yaml
georgettica: It is, in the openshift template

openshift/template.yaml
georgettica: Yeah, I can revert back to your suggestion, thanks!

shibumi: I am confused now :D @patjlm  didn't you suggest that we can just leave serviceAccountName empty and use the pipeline SA? Why the sudden revert? (Just want to understand this, I am fine with both decisions :) )

shibumi: ok, I understand

patjlm: I don't mind leaving this as is for now. But the least-privilege principle is definitely something you should target to implement

shibumi: I agree

shibumi: @georgettica  do you want to fix the SA line? I was wrong ^^ I think you wanted to leave the custom SA during our meeting and I told you to remove it. Sorry :/

deploy/task-cad-checks.yaml
shibumi: I think this is resolved... as long as it is in the openshift template. We should also make sure to use the openshift template for dev deployments in the future. Today, we used the deploy folder :thinking: although both should be the same, right?

georgettica: Yes

PR #70, opened 2022-05-18
fix(deploy): Rename role and role-binding

shibumi:
The role and role-binding are for the pipelinerun and not for the eventListener. Let us rename them accordingly.


georgettica:
You didn't update the template, run  ` make validate` until it passes
You will need to commit the changes there / decide if you don't want it to change

shibumi:
Thank you @georgettica!

PR #72, opened 2022-05-19
fix: remove namespaces

shibumi:
We shall not specify namespaces, because otherwise it's impossible to deploy to cad and cad-stage namespaces with the different pipelines.

For dev deployments we can either do `oc apply -f -n cad` or set the current project as context in oc
via `oc project cad`.


georgettica:
I just needed to change one thing in the conversation script, and it would have worked :(
K

PR #73, opened 2022-05-23
fix(ocm): Fix debugMode for OCM

shibumi:
strconv.ParseBool raises an error when debugMode is empty, thus
we have to set it to false if the value is empty.

Fix: [OSD-11695](https://issues.redhat.com//browse/OSD-11695)


georgettica:
What card is that related to?

shibumi:
@georgettica  I have attached the card for it :)

PR #75, opened 2022-05-25
fix(aws): Change ListStoppedInstances to ListNonRunningInstances

ninabauer:
ListStoppedInstances updated to ListNonRunningInstances to reflect that it is not only stopped instances.

[OSD-11731](https://issues.redhat.com//browse/OSD-11731)


georgettica:
You just needed to change the deps aswell :/

PR #76, opened 2022-05-25
fix(aws): Change ListStoppedInstances to ListNonRunningInstances

ninabauer:
[osd-11731](https://issues.redhat.com//browse/osd-11731)

ListStoppedInstances updated to ListNonRunningInstances to reflect that it is not only stopped instances.


georgettica:
/lgtm
/approve
this look great!

and it makes the fmt clearer for the user :)
/hold
as I want to allow others to look aswell

PR #77, opened 2022-05-25
fix(chgm): make status messages visible in pod logs

ninabauer:
[OSD-11712](https://issues.redhat.com//browse/OSD-11712)

Makes status messages visible in pod logs.

Also a few minor typo fixes.

### DIFF
--- a/deploy/task-cad-checks.yaml
+++ b/deploy/task-cad-checks.yaml
@@ -21,7 +21,7 @@ spec:
           file=${folder}/payload.json
           echo '$(params.payload)' > $file
           # run the cadctl command
-          RESULT=$(cadctl cluster-missing --payload-path $file )
+          cadctl cluster-missing --payload-path $file
       # envFrom should pull all the secret information as envvars, so key names should be uppercase
       envFrom:
         - secretRef:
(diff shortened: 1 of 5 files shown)


shibumi:
Good job Nina!

/lgtm
/approve
/hold

holding because others might want to review, too. (Looking especially at you @georgettica )

deploy/task-cad-checks.yaml
```diff
@@ -21,7 +21,7 @@ spec:
           file=${folder}/payload.json
           echo '$(params.payload)' > $file
           # run the cadctl command
-          RESULT=$(cadctl cluster-missing --payload-path $file )
+          cadctl cluster-missing --payload-path $file
```
georgettica, line 24: The reason this was wrapped in a RESULT, was to pipe the output to another task for processing..
We should also have the json output saved to a file and passed as an output for future plumbing

shibumi: This other task does not exist as of now, right?

The problem here is that due to the RESULT variable the stdout gets never forwarded to stdout. The variable catches it and this is why we don't see stdout messages in the container logs.

Can we do both some how? Printing and inserting into a file? The `tee` command should work this way, right?

shibumi:
I am going to unhold this for now, Ron gave his okay in Slack.

PR #80, opened 2022-05-27
fix: cluster.Region().Name() returns an empty string

shibumi:
hence we have to use ID() to retrieve the correct region the cluster.

Fixes: [OSD-11734](https://issues.redhat.com//browse/OSD-11734)


shibumi:
I am still investigating this, but looks like we have a bug in OCM, where Name() does not return the region. ID() seems to work on the other hand.

Tessg22:
> // NOTE: Currently for all cloud providers and all regions `id` and `name` have exactly

/lgtm

PR #82, opened 2022-05-31
fix: polling timeout

shibumi:
Before, we have always selected the first resource, that's why we restarted the loop again.
With this fix, we are now looping correctly over all resources in each event.

This PR is brought to you by great teamwork from @ninabauer  @Tessg22 and me :)


shibumi:
This probably needs further testing, because I am unsure if we can just remove the part about the "timer"

PR #89, opened 2022-06-02
fix(chgm): add error when 0 nonrunning instances are found

ninabauer:
[osd-11802](https://issues.redhat.com//browse/osd-11802)

### DIFF
--- a/pkg/services/chgm/chgm.go
+++ b/pkg/services/chgm/chgm.go
@@ -212,7 +212,7 @@ func (c Client) investigateInstances() (InvestigateInstancesOutput, error) {
 
 	if len(stoppedInstances) == 0 {
 		// UserAuthorized: true so SRE will still be alerted for manual investigation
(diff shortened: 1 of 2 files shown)


ninabauer:
/hold

WIP

pkg/services/chgm/chgm.go
```diff
@@ -207,7 +211,8 @@ func (c Client) investigateInstances() (InvestigateInstancesOutput, error) {
 	// fmt.Printf("stoppedInstances ::: %#v\n", stoppedInstances)
 
 	if len(stoppedInstances) == 0 {
-		return InvestigateInstancesOutput{UserAuthorized: true}, nil
+		// UserAuthorized: true so SRE will still be alerted for manual investigation
+		return InvestigateInstancesOutput{UserAuthorized: true, Error: "no non running instances found, terminated instances may have already expired"}, fmt.Errorf("no non running instances found, terminated instances may have already expired")
```
shibumi, line 215: If we return an error here, we will never create the note at PD. You have to do this instead:

```
return InvestigateInstancesOutput{UserAuthorized: true, Error: "no non running instances found, terminated instances may have already expired"}, nil
```

PR #95, opened 2022-06-07
feat: add cad dashboard playground

ninabauer:
adds starter playground for cad dashboard


shibumi:
@ninabauer  I think you have to put this into the `deploy/` folder and then re-run `make generate` to create the openshift template file for it.

PR #114, opened 2022-06-16
fix: remove AWS token depletion

shibumi:
Amazon AWS tokens are used for API throttling. Every call to an API endpoint
costs us one token. Before this commit, our listAllInstancesAttribute
exhausted all our available AWS tokens, this lead to AWS throttling us.

The new function does only ONE request instead of one request for every token.
In this request we gather a maximum of 1000 results for a 2hour timeframe.
This should be enough for us to handle CHGM alerts

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -366,20 +366,18 @@ func (c Client) ListAllTerminatedInstances() ([]*cloudtrail.Event, error) {
 }
 
 func (c Client) listAllInstancesAttribute(att *cloudtrail.LookupAttribute) ([]*cloudtrail.Event, error) {
+	// We only look up events that are not older than 2 hours
+	since := time.Now().Add(time.Duration(-2) * time.Hour)
+	// We only look up 1000 events maximum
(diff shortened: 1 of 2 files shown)


shibumi:
I tested this PR with my stage cluster. Seems to work:
```
Sending CHGM ServiceLog...
Silencing Alert...
Attaching Note Is user authorized: 'false' 

UserName : 'osdManagedAdmin' 

The amount of non running instances is: '10' 

Instance IDs: '[i-027e2812159da23c3 i-071042c56fed3d97c i-07e8483f620d8471e i-0b3118cac5080ffd1 i-0af8a031839a0f355 i-0cda8c163b538777b i-031b382f293de53c4 i-0f02e908d6b5f9611 i-073e50fda81bcac08 i-038f64ed09e107bb5]' 

ServiceLog Sent: 'Action required: cluster not checking in' 
```

Nikokolas3270:
Seems indeed a good idea to let AWS do the loop internally rather than doing it ourself.
Also from a scalability perspective it looks fine to limit the output.

PR #127, opened 2022-07-23
iterate over pages on pageinated ListInstance results for aws pkg

ramonbutter:
was looking over some of the old code, I wrote...
and believe to found a bug, that will cause an infinity loop of calling the AWS API for the case the instances are returned on several pages.  
The nextToken has to be stored in the Request, as it is done [here](https://github.com/evalphobia/aws-sdk-go-wrapper/blob/0ad241a2f698d46f056aee89f79dc5624cd5cdba/cloudtrail/client.go#L64).

Great to see though that things are still going forward with CAD.

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -179,13 +179,13 @@ func (c Client) listInstancesWithFilter(filters []*ec2.Filter) ([]*ec2.Instance,
 		if err != nil {
 			return []*ec2.Instance{}, err
 		}
-		nextToken := out.NextToken
 		for _, res := range out.Reservations {
 			instances = append(instances, res.Instances...)
 		}
-		if nextToken == nil {
+		if out.NextToken == nil {
 			break
 		}
+		in.NextToken = out.NextToken
 	}
 	return instances, nil
 }


rafael-azevedo:
@ramonbutter nice catch ! And yes CAD still going strong its deployed in production and has caught ~95% CHGM. Feel free to contribute anytime hope all is well !

shibumi:
Funny, I thought I have removed all of these infinite for loops.

We have fixed a few other of these loops in this PR here: https://github.com/openshift/configuration-anomaly-detection/pull/114/files

Are you sure that we should use these tokens @ramonbutter? The problem with using these tokens is, that we will easily run into API throttling, That's why we have increased the maxResults and set a better startTime instead.

I will have a closer look at this on monday

ramonbutter:
I am pretty sure that with the current approach and the result contains out of several pages, it would run and call the API in an infinity loop until it is throttled and than error out of the loop.  Probably this was actually the cause for the throttling, and not the appearance of multiple pages here and there in the results for big clusters.
Usually this should not happen, as you can list many entries on one page. So only for chgm for massive clusters, this would have been the case. For listing instances I believe you would still want to look at all, as otherwise would limit your investigate on a part of the clusters. So yes, I guess the tokens have to be used to navigate between pages.
However, for the attributes it may make sense to limit it. Not sure anymore what they were used for, just saw that pretty certain flaw. Anyway, I am sure you will figure it out. 
Hope also all is good

shibumi:
I am a little divided here.. @ramonbutter's solution fixes the infinite loop for the token, but I am really afraid that we might run into  a token depletion and thus into a throttling of the Amazon AWS API.

According to: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html we have only 50 tokens in our Bucket with a refill rate of 10 tokens per second.

In the past, we have depleted this pool rather quickly. I am really afraid that CAD could request 50 pages, leading to a token depletion.

The big question I have is: How likely is this event?

shibumi:
However, we should definitely merge this for now, because without Ramon's fix we are running into an infinite loop.

shibumi:
@ramonbutter never mind.. I just realized that your commit is for `DescribeInstances` and not for `ListEvents`. This is fine and the correct solution and I am wrong. I am just wondering that this code works on production. LOL

PR #131, opened 2022-09-09
Adding error propagation to PD

Makdaam:
so that the SRE on-call knows that CAD won't enrich the issue without looking into pod logs.
Example issue that can be avoided in the future https://issues.redhat.com/browse/OSD-11710

Imports rearranged by linter.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ test:
 
 .PHONY: lint
 lint: bin/golangci-lint lint-only-hack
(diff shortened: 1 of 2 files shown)


rafael-azevedo (commented):
@Makdaam I'd love to merge this can you investigate why the tests are failing ?

Makdaam:
@rafael-azevedo thanks for the ping. Extended the timeout of lint from 60s (default) to 120s.

typeid:
@Makdaam can we close this PR? `main` diverged to far off and from what I see, it is already similarly implemented there.

PR #134, opened 2022-10-05
OSD-12961; OSD-12957 - Integrate Limited Support into CAD's CHGM logic

tnierman:
Completes: https://issues.redhat.com/browse/OSD-12961; https://issues.redhat.com/browse/OSD-12957

---

These changes add logic to update the Limited Support status for a cluster based on the state of it's Cluster Has Gone Missing (CHGM) alert. 

When the CHGM alert fires initially, CAD evaluates the cluster and determines whether the alert was caused by a customer action. In cases where this is determined to be the case, the cluster is placed into Limited Support. Otherwise, the alert is forwarded based on the environment's PagerDuty escalation policy.

When the CHGM alert resolves, CAD once again evaluates the cluster and determines whether it is sufficiently healthy to be removed from Limited Support. This investigation consists of the following: 1) checking that all EC2 instances owned by the cluster in the customer's VPC are running, and 2) whether the number of EC2 instances in the customer's VPC matches the expected number in OCM (ensuring the customer has not deleted a machine). 
  - If the cluster passes the investigation, all Limited Support reasons, both CHGM and CCAM, matching the templates used by CAD are removed. (That is, limited support reasons added to the cluster manually are not guaranteed to be removed by CAD at this time.) 
  - If a cluster does not pass investigation, an alert is created on the environment's deadmanssnitch service using an integration called `Dead Man's Snitch` (the service for each environment is determined via the webhook payload). 

For `resolved` incidents in particular, most actions are retried until they succeed. This is purposeful, to prevent clusters from remaining in Limited Support due to a transient error (ie- PagerDuty throttling, AWS region down) when investigating them or removing a LS reason. 

In all cases (alert triggering or resolving), if CAD is unable to login to the cluster's AWS environment due to a missing support role, the Cloud Credentials Are Missing (CCAM) Limited Support reason is sent and no further action is taken. (As mentioned above, if this LS reason is sent when the alert triggers, but is found to be fixed when the alert is resolved, then that LS reason is removed as well). It's expected that customers will follow the normal LS workflow to remove this reason from their clusters if CAD does not do so.

### DIFF
--- a/cadctl/cmd/cluster-missing/cluster-missing.go
+++ b/cadctl/cmd/cluster-missing/cluster-missing.go
@@ -18,9 +18,11 @@ package clustermissing
 
 import (
 	"fmt"
-	"github.com/openshift/configuration-anomaly-detection/pkg/services/ccam"
 	"os"
 	"path/filepath"
+	"time"
+
+	"github.com/openshift/configuration-anomaly-detection/pkg/services/ccam"
 
 	"github.com/openshift/configuration-anomaly-detection/pkg/aws"
 	ocm "github.com/openshift/configuration-anomaly-detection/pkg/ocm"
@@ -65,21 +67,6 @@ func run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("could not start pagerdutyClient: %w", err)
 	}
 
-	eventType, err := pdClient.ExtractEventTypeFromPayload(payloadPath, pagerduty.RealFileReader{})
-	if err != nil {
-		return fmt.Errorf("could not determine event type: %w", err)
-	}
-
-	switch eventType{
-	case pagerdutyIncidentResolved:
-		return evalutateRestoredCluster(awsClient, ocmClient, pdClient)
-	default:
-		return evaluateMissingCluster(awsClient, ocmClient, pdClient)
-	}
-}
-
-// evaluateMissingCluster will take appropriate action against a cluster whose incident is not resolved
-func evaluateMissingCluster(awsClient aws.Client, ocmClient ocm.Client, pdClient pagerduty.Client) error {
 	incidentID, err := pdClient.ExtractIncidentIDFromPayload(payloadPath, pagerduty.RealFileReader{})
 	if err != nil {
 		return fmt.Errorf("GetIncidentID failed on: %w", err)
@@ -89,6 +76,11 @@ func evaluateMissingCluster(awsClient aws.Client, ocmClient ocm.Client, pdClient
 	externalClusterID, err := pdClient.ExtractExternalIDFromPayload(payloadPath, pagerduty.RealFileReader{})
 	if err != nil {
 		return fmt.Errorf("GetExternalID failed on: %w", err)
+	}
+
+	eventType, err := pdClient.ExtractEventTypeFromPayload(payloadPath, pagerduty.RealFileReader{})
+	if err != nil {
+		return fmt.Errorf("could not determine event type: %w", err)
 	}
 
 	fmt.Printf("ClusterExternalID is: %s\n", externalClusterID)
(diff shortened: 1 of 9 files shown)


pkg/pagerduty/pagerduty.go
```diff
@@ -23,6 +23,10 @@ const (
 	InvalidInputParamsErrorCode = 2001
 	// pagerDutyTimeout is the chosen timeout for api requests, can be changed later
 	pagerDutyTimeout = time.Second * 30
+	// CADEmailAddress is the email address for the 'Configuration-Anomaly-Detection' PagerDuty User
+	CADEmailAddress = "sd-sre-platform+pagerduty-configuration-anomaly-detection-agent@redhat.com"
```
mjlshen, line 27: Does CAD have a config file? These values seem more suitable for that rather than hard-coded into the binary.

If not, I'm good with doing this for now

pkg/pagerduty/pagerduty.go
```diff
@@ -186,6 +190,52 @@ func (c Client) GetAlerts(incidentID string) ([]Alert, error) {
 	return res, nil
 }
 
+// CreateNewAlert triggers an alert using the Deadmanssnitch integration for the given service.
+// If the provided service does not have a DMS integration, an error is returned
+func (c Client) CreateNewAlert(description string, details interface{}, serviceID string) error {
```
mjlshen, line 195: Does details need to be an `interface{}`?

We seem to be using this struct as input whenever we call this in this PR, so I think we should define this specific struct and enforce the type on `details`

```go
struct {
		ClusterID  string `json:"Cluster ID"`
		Error      string `json:"Error"`
		Resolution string `json:"Resolution"`
		SOP        string `json:"SOP"`
	}
```

cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -143,39 +142,194 @@ func evaluateMissingCluster(awsClient aws.Client, ocmClient ocm.Client, pdClient
...
+			sleepBeforeRetrying(retries, 300)
+			retries++
+
+			serviceID, err = chgmClient.ExtractServiceIDFromPayload(payloadPath, pagerduty.RealFileReader{})
+		}
+		return serviceID
+}
+
+// sleepBeforeRetrying sleeps with an exponential backoff based on the current retry, up to the given maximum sleep time.
+func sleepBeforeRetrying(currentRetry, maxSleepSeconds int) {
```
mjlshen, line 322: Do we want an ability to cap the number of retries with this function directly?

I noticed that we sometimes manually cap `retries <= 3` when this is called and then sometimes forget to manually cap it. I think if we add an input for `maxRetries` and also return an `error`, we can return a non-nil error if the amount of retries has exceeded and we can just rely on the infinite for loop to handle `for err != nil`

pkg/pagerduty/pagerduty.go
tnierman: No, CAD seems to use environment variables over config files which allows it to easily be configured at the cluster-level.

This isn't a value that should be altered between environments, so I think leaving it as a constant makes sense here.

mjlshen: Thanks, sounds good to me

cadctl/cmd/cluster-missing/cluster-missing.go
tnierman: Retry logic is only present when evaluating restored clusters (cluster's that had their CHGM alert resolve), and the number of retries is only capped when attempting to remove clusters from Limited Support after a successful investigation. This was intentional, since, if we fail to remove Limited Support reasons after several attempts, this likely indicates a non-transient failure that needs to be escalated to Primary (ie- OCM outage, malformed reason, CAD misconfiguration). 

The number of retries is left unbounded in all cases where we need to escalate a condition to Primary. The reasoning here being, unlike cases where a CHGM has just triggered (evaluating a missing cluster), there is no other fallback option. If the investigation fails when the CHGM triggers, the alert will eventually escalate to Primary for examination. But if the investigation fails when the CHGM resolves, there is no signal that will escalate to Primary by default,  and we risk leaving clusters in Limited Support beyond the intended timeframe.

pkg/pagerduty/pagerduty.go
tnierman: At the moment, the `details` argument just reflects the `details` field for an [Event object in go-pagerduty](https://pkg.go.dev/github.com/pagerduty/go-pagerduty#Event), which is an `interface{}`. We could create a more structured-structure, but personally I like the flexibility it gives callers in defining only the fields relevant to the alert being created

mjlshen: Thanks, mirroring upstream makes sense to me.

cadctl/cmd/cluster-missing/cluster-missing.go
tnierman: Since CAD doesn't follow the typical operator reconcile-loop architecture, I don't think this is as much of a concern (we don't run the risk of blocking other cluster evaluations by suspending the process for X amount of time, etc).

tnierman:
I've pushed an update to include the SOPs added in https://github.com/openshift/ops-sop/pull/2268

sam-nguyen7 (commented):
/lgtm

PR #135, opened 2022-11-29
Add cloud provider check before chgm investigation

RaphaelBut:
Cad currently only supports aws clusters and throws errors like this on gcp.
```
Assuming role failed, potential CCAM alert. Investigating... error:  3 failed to assume into jump-role: RequestError: send request failed
caused by: Post "https://sts.us-east1.amazonaws.com/": dial tcp: lookup sts.us-east1.amazonaws.com on 10.120.0.10:53: no such host
Error: credentials are there, error is different: 3 failed to assume into jump-role: RequestError: send request failed
caused by: Post "https://sts.us-east1.amazonaws.com/": dial tcp: lookup sts.us-east1.amazonaws.com on 10.120.0.10:53: no such host
```

This PR adds a check for the clusters cloud provider and throws a clear error message in case it is not 'aws'.
```
Incident ID is: Q2YUU901Z5FKY7
Error: cloudprovider is not supported: gcp
```

Closes:
- https://issues.redhat.com/browse/OSD-13417

### DIFF
--- a/cadctl/cmd/cluster-missing/cluster-missing.go
+++ b/cadctl/cmd/cluster-missing/cluster-missing.go
@@ -85,9 +85,9 @@ func run(cmd *cobra.Command, args []string) error {
 
 	fmt.Printf("ClusterExternalID is: %s\n", externalClusterID)
 
-	cloudProvider, err := ocmClient.GetCloudProvider(externalClusterID)
+	cloudProvider, err := ocmClient.GetCloudProviderID(externalClusterID)
 	if err != nil {
-		return fmt.Errorf("GetCloudProvider failed on: %w", err)
+		return err
 	} else if cloudProvider != "aws" {
 		return fmt.Errorf("cloudprovider is not supported: %s", cloudProvider)
 	}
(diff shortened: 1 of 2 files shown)


pkg/ocm/ocm.go
```diff
@@ -182,6 +182,24 @@ func (c Client) getClusterResource(clusterID string, resourceKey string) (string
 	return response.Body().Resources()[resourceKey], nil
 }
 
+// GetCloudProvider returns the cloud provider name for a given cluster as a string
+func (c Client) GetCloudProvider(identifier string) (string, error) {
```
typeid, line 186: To avoid confusing it with the `cluster.GetCloudProvider` function. This function wraps the extraction process of `cloudProvider.ID` .

```suggestion
// GetCloudProviderID returns the cloud provider name for a given cluster as a string
func (c Client) GetCloudProviderID(identifier string) (string, error) {
```

cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -85,6 +85,13 @@ func run(cmd *cobra.Command, args []string) error {
 
 	fmt.Printf("ClusterExternalID is: %s\n", externalClusterID)
 
+	cloudProvider, err := ocmClient.GetCloudProvider(externalClusterID)
```
typeid, line 88: In case you go with the below suggestion

```suggestion
	cloudProvider, err := ocmClient.GetCloudProviderID(externalClusterID)
```

cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -85,6 +85,13 @@ func run(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return fmt.Errorf("GetCloudProvider failed on: %w", err)
```
typeid, line 90: **todo:** Remove duplicate error text. You can just return err here, as the error message is formatted that way already in the `GetCloudProvider` function

pkg/ocm/ocm.go
RaphaelBut: +1, included the suggested change

PR #136, opened 2022-11-29
OSD-13901 - Use heredocs when generating payload to avoid quoting errors

tnierman:
Completes: https://issues.redhat.com/browse/OSD-13901

---

These changes modify the payload generation logic, to avoid evaluating the quoting used within.

This is necessary because the string `Dead Man's Snitch` frequently appears in these payloads. When there's an odd number of these present in the data, bash interprets this as a mismatched quotation and the cluster evaluation never starts.

### DIFF
--- a/deploy/task-cad-checks.yaml
+++ b/deploy/task-cad-checks.yaml
@@ -19,7 +19,9 @@ spec:
           # save the payload to a file
           folder=$(mktemp -d)
           file=${folder}/payload.json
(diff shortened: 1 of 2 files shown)


rafael-azevedo (approved):
/lgtm

mjlshen (approved):
/lgtm

rafael-azevedo (approved):
/lgtm

PR #142, opened 2022-12-15
Additional bugfixes for CAD-LS integration

tnierman:
These changes do the following:
- Avoid duplicating CHGM LS reasons if the cluster already has one
- Allow CCAM alerts to be silenced, even if a cluster already has a CCAM alert

### DIFF
--- a/cadctl/cmd/cluster-missing/cluster-missing.go
+++ b/cadctl/cmd/cluster-missing/cluster-missing.go
@@ -156,13 +156,22 @@ func evaluateMissingCluster(chgmClient chgm.Client, incidentID string, externalC
 		return nil
 	}
 
-	// Post CHGM limited support
-	fmt.Println("Sending CHGM limited support reason")
-	reason, err := chgmClient.PostCHGMLimitedSupport(externalClusterID)
+	lsExists, err := chgmClient.CHGMLimitedSupportExists(externalClusterID)
 	if err != nil {
-		return fmt.Errorf("failed posting limited support reason: %w", err)
+		return fmt.Errorf("failed to determine if limited support reason already exists: %w", err)
 	}
-	res.LimitedSupportReason = reason
+
+	if !lsExists {
+		// Post CHGM limited support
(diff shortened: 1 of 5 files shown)


cadctl/cmd/cluster-missing/cluster-missing.go
```diff
@@ -156,13 +156,22 @@ func evaluateMissingCluster(chgmClient chgm.Client, incidentID string, externalC
...
+		// Post CHGM limited support
+		fmt.Println("Sending CHGM limited support reason")
+		reason, err := chgmClient.PostCHGMLimitedSupport(externalClusterID)
+		if err != nil {
+			return fmt.Errorf("failed posting limited support reason: %w", err)
+		}
+		res.LimitedSupportReason = reason
+	} else {
+		fmt.Printf("Avoided posting duplicate CHGM Limited Support reason")
 	}
```
sam-nguyen7, line 174:
```suggestion
	}
```

pkg/services/ccam/ccam.go
```diff
@@ -86,13 +86,13 @@ func (c Client) Evaluate(awsError error, externalClusterID string, incidentID st
 	}
 	if lsExists {
 		fmt.Println("Avoided reposting duplicate CCAM limited support reason")
-		return nil
+	} else {
+		ls, err := c.PostCCAMLimitedSupportReason(c.cluster.ID())
+		if err == nil {
+			fmt.Printf("Added the following Limited Support reason to cluster: %#v\n", *ls)
+		}
```
sam-nguyen7, line 93: Can we follow the same pattern used above in cluster-missing, `if !lsExists`?

typeid:
Thanks for the fixes Trevor! 

/lgtm

cadctl/cmd/cluster-missing/cluster-missing.go
sam-nguyen7: I agree that it could be helpful context when investigating CAD. However, if possible, we should avoid having `else` block that does not have much logic.

PR #143, opened 2022-12-15
Ensure operator role is allowed to stop instances

sam-nguyen7:
Using strings.Contains() instead of strings.HasPrefix() to properly catch operator roles that are allowed to stop instances.

### DIFF
--- a/pkg/services/chgm/chgm.go
+++ b/pkg/services/chgm/chgm.go
@@ -93,9 +93,7 @@ func isUserAllowedToStop(username, issuerUsername string, userDetails CloudTrail
 		"openshift-machine-api-aws",
 	}
 	for _, operatorIamName := range operatorIamNames {
-		// I added a '-' in the end to make sure it's the same as close as I can to the aws username (cannot guess the uuid)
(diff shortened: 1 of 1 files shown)


pkg/services/chgm/chgm.go
```diff
@@ -95,7 +95,7 @@ func isUserAllowedToStop(username, issuerUsername string, userDetails CloudTrail
 	for _, operatorIamName := range operatorIamNames {
 		// I added a '-' in the end to make sure it's the same as close as I can to the aws username (cannot guess the uuid)
 		operatorIamName = fmt.Sprintf("%s-%s-", infraID, operatorIamName)
-		if strings.HasPrefix(username, operatorIamName) {
+		if strings.Contains(username, operatorIamName) {
```
typeid, line 98: **Question:** Why is this change necessary? Did the formatting for the username change, or did this never work to begin with?

sam-nguyen7: some STS role names do not conform to the infraID-operatorIamName format, so we were not properly escalating some CHGM before

PR #145, opened 2023-01-04
Check incident.trigger for existing LS reasons

sam-nguyen7:
https://issues.redhat.com/browse/OSD-14401

When DMS integration is re-enabled, CAD needs to account for CHGM alerts of clusters that are already in limited support. With this PR, CAD will ignore these CHGM alerts prior to them being escalated to Primary, and send them to Silent Test user.


rafael-azevedo (approved):
/lgtm

PR #147, opened 2023-01-13
Don't evaluate restored cluster if in limited support

tnierman:
https://issues.redhat.com/browse/OSD-14636

---

Adds logic to the `investigateStartedInstances` function to ignore clusters with a limited support status.


tnierman:
/hold 

this is wrong

tnierman:
/unhold
changes have been tested now

PR #151, opened 2023-01-30
Keep pipelines for a week

typeid:
**What?**

Keep tekton pipelines for a week instead of pruning everything but the last 20 every hour.

**Why?**

We need access to pipeline run logs and don't want them to be pruned away before investigating.


typeid:
/unhold

See https://redhat-internal.slack.com/archives/CCRND57FW/p1675080656903309

PR #153, opened 2023-02-08
OSD-14629 - CAD node/instance count check can misfire

zmird-r:
Updated the node count check to differentiate between master, infra and worker. Included the machine pools workers in the workers count.
The investigation of started instance gives an error only if there's a mismatch in the count of master or infra nodes.

### DIFF
--- a/pkg/services/chgm/chgm.go
+++ b/pkg/services/chgm/chgm.go
@@ -166,9 +166,9 @@ func (i InvestigateInstancesOutput) String() string {
 		msg += fmt.Sprintf("\nIssuerUserName: '%v' \n", i.User.IssuerUserName)
 	}
 	msg += fmt.Sprintf("\nNumber of non running instances: '%v' \n", len(i.NonRunningInstances))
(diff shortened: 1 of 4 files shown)


zmird-r:
Closes: https://issues.redhat.com/browse/OSD-14629

pkg/services/chgm/chgm.go
```diff
@@ -376,31 +412,132 @@ func (c Client) RemoveCCAMLimitedSupport(externalID string) (bool, error) {
...
+    return runningNodesCount, nil
+}
+
+// GetExpectedNodesCount returns the mininum number of nodes that are supposed to be in the cluster
 // We do not use nodes.GetTotal() here, because total seems to be always 0.
-func (c Client) GetNodeCount() int {
+func (c Client) GetExpectedNodesCount() (*ExpectedNodesCount, error) {
 	nodes, ok := c.cluster.GetNodes()
 	if !ok {
 		// We do not error out here, because we do not want to fail the whole run, because of one missing metric
```
RaphaelBut, line 454: polish: i believe this comment is leftover and can be removed as you do error out here

RaphaelBut (approved):
Thanks a lot for this change, great work!
this will hopefully reduce noise on primary a lot. 
/lgtm

PR #156, opened 2023-02-09
Hotfix pipelinepruner

RaphaelBut:
Why: Stage ran into quota issues after https://github.com/openshift/configuration-anomaly-detection/pull/151 was merged. We did not anticipate the amount of runs that actually come in. 

return to hourly purge but keep 100 runs for debug purposes

### DIFF
--- a/deploy/tekton-pruner.yaml
+++ b/deploy/tekton-pruner.yaml
@@ -15,7 +15,7 @@ spec:
         spec:
           containers:
             - args:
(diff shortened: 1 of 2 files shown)


typeid:
LGTM for a hotfix, but we definitely need to take a closer look at this, as we're going back in a state where we're not able to properly look into problematic pipelines older than a day or two.

/lgtm

RaphaelBut:
it is on the cadcon discussion list, we will discuss what our requirements are and open corresponding bug/story

PR #157, opened 2023-02-15
fix: osdmanagedadmin user prefix

ninabauer:
[OSD-15256](https://issues.redhat.com//browse/OSD-15256)

The osdmanagedadmin user does not always contain a suffix, sometimes it can be just "osdmanagedadmin". This fixes the authorized user check.


typeid:
@ninabauer Thanks for the quick fix! If it's not too much trouble for you, could you commit just the one line with the fix? The rest will make it harder to rebase other open PRs. 

I have noticed we don't have any format checks for the PRs, this results in exactly this issue. The files weren't formatted ahead of time. I will open a PR to add a format check, and merge it after Raphael's refactor to not cause him too much pain while rebasing.

PR #158, opened 2023-02-15
Add diagrams and documentation

typeid:
**What?** 

Add some diagrams to get more visibility on how CAD works.

Currently a draft as these diagrams represent the near future of CAD and aren't 100% accurate with the current state.


RaphaelBut:
I think its indeed a good idea to already merge these as they greatly help getting an overview. Very nice drawings! 

/lgtm

PR #159, opened 2023-02-16
Add gofmt to linters

typeid:
**Why?**

We need gofmt in the lint check to make sure everyone's formatting is the same.

Ideally, we want the golangci-lint file to look something like this: https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322#file-golangci-yml


typeid:
/hold
until Raphael's refactor is on `main`, as to not cause conflicts.

PR #168, opened 2023-03-27
Add note for unsupported infra

typeid:
**What?**

- refactor retries function to remove duplicate defaults and global variables (commit 1)
- Add pagerduty note when infrastructure is unsupported for investigation (commit 2)

**Why?**

- General maintainance
- Make it clear why CAD didn't run on a CHGM for GCP clusters


RaphaelBut:
Great job! 
/lgtm

PR #174, opened 2023-04-19
OSD-15569 : Add CCS/nonCCS differentiation

RaphaelBut:
The role 'OrganizationAccountAccessRole' is used by SRE in case of non CCS clusters and by the user otherwise.
This PR changes the isUserAllowedToStop function to respect the clusters flavor and return false when 'OrganizationAccountAccessRole' was used on a CCS cluster.

### DIFF
--- a/pkg/services/chgm/chgm.go
+++ b/pkg/services/chgm/chgm.go
@@ -227,7 +227,7 @@ func (c *Client) investigateRestoredCluster() (res InvestigateInstancesOutput, e
 // isUserAllowedToStop verifies if a user is allowed to stop/terminate instances
 // For this, we use a whitelist of partial strings that can be SRE
 // based on findings in https://issues.redhat.com/browse/OSD-16042
-func (c *Client) isUserAllowedToStop(username, issuerUsername string, userDetails CloudTrailEventRaw, infraID string) bool {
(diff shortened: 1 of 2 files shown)


RaphaelBut:
/hold 

for promotion

pkg/services/chgm/chgm.go
```diff
@@ -227,7 +227,7 @@ func (c *Client) investigateRestoredCluster() (res InvestigateInstancesOutput, e
 // isUserAllowedToStop verifies if a user is allowed to stop/terminate instances
 // For this, we use a whitelist of partial strings that can be SRE
 // based on findings in https://issues.redhat.com/browse/OSD-16042
-func isUserAllowedToStop(username, issuerUsername string, userDetails CloudTrailEventRaw, infraID string) bool {
+func (c *Client) isUserAllowedToStop(username, issuerUsername string, userDetails CloudTrailEventRaw, infraID string) bool {
```
typeid, line 230: **Question:** is this something we want to convert the function to a method for? 
I think we're partially passing stuff per parameter (e.g. infraID) and partially using it using the `c  *Client`s members now. 

Otherwise lgtm

RaphaelBut: Mhh good point. 
It works both ways and is a little strange here as this function is only called during and from the chgm service (same scope). I will change it back and pass per parameter tough I think it is cleaner that way.

PR #175, opened 2023-04-20
Make resolve investigation logging less confusing

typeid:
**Why**?

See https://redhat.pagerduty.com/incidents/Q3DIQB8OOFEBVS/timeline

We don't clarify what CAD does in the note, it looks like a second investigation.


typeid:
/hold
Until after promotion

PR #178, opened 2023-04-21
Escalate directly in case cloud provider or alert is not supported

typeid:
**What?**

- Directly escalate alerts not handled by CAD.

**Why?**

- We are currently not directly escalating GCP alerts, affecting our mean time to resolve for GCP CHGM.

Jira: [OSD-16111](https://issues.redhat.com//browse/OSD-16111)

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -68,7 +68,11 @@ func run(cmd *cobra.Command, args []string) error {
 	alertType := isAlertSupported(pdClient.GetTitle(), pdClient.GetServiceName())
 	if alertType == "" {
 		fmt.Printf("Alert is not supported by CAD: %s", pdClient.GetTitle())
-		pdClient.EscalateAlert()
+		err = pdClient.EscalateAlert()
+		if err != nil {
+			return err
+		}
+
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


RaphaelBut:
Nice! Important details when we start routing other alerts.
/lgtm

cadctl/cmd/investigate/investigate.go
```diff
@@ -85,11 +85,16 @@ func run(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+
 	// We currently have no investigations supporting GCP. In the future, this check should be moved on
 	// the investigation level, and we should be GCP or AWSClient based on this.
-	_, err = checkCloudProviderSupported(&ocmClient, &pdClient, externalClusterID, []string{"aws"})
-	if err != nil {
-		return fmt.Errorf("failed cloud provider check: %w", err)
+	if !cloudProviderSupported {
+		pdClient.EscalateAlertWithNote(fmt.Sprintf("CAD Investigation skipped: cloud provider is not supported."))
```
NautiluX: couldn't we pass the string directly instead of calling Sprintf?

cadctl/cmd/investigate/investigate.go
```diff
@@ -69,6 +68,7 @@ func run(cmd *cobra.Command, args []string) error {
 	alertType := isAlertSupported(pdClient.GetTitle(), pdClient.GetServiceName())
 	if alertType == "" {
 		fmt.Printf("Alert is not supported by CAD: %s", pdClient.GetTitle())
+		pdClient.EscalateAlert()
```
NautiluX: return value not checked (see pipeline)

cadctl/cmd/investigate/investigate.go
NautiluX: return value not checked (see pipeline)

NautiluX (changes requested):

cadctl/cmd/investigate/investigate.go
typeid: Indeed, thanks for pointing it out. Fixed this and the above. Unfortunately without https://issues.redhat.com/browse/OSD-15137 I can't run the lint job locally unless I figure out how to run it in a container with the correct go version, because the golangci-lint version is incompatible with my go 1.19.

PR #179, opened 2023-04-21
Upgrade golangci-lint to v1.52.2 and use local version if existing

typeid:
**What?**

Dynamically downloads golangci-lint v1.52.2. If golangci-lint is already installed, it uses the already installed golang-ci.lint.

This fixes issues when using a higher go version that doesn't fit the previously old hardcoded golangci-lint version of the makefile.

**Why?**

This results in nasty errors and blocks https://issues.redhat.com/browse/OSD-15137 because of a desync in the lint and go version.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,7 +14,6 @@ issues:
 linters:
   disable-all: true
   enable:
-    - deadcode
     - errcheck
     - gocritic
     - gocyclo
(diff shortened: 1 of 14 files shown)


.golangci.yml
```diff
@@ -25,9 +24,7 @@ linters:
     - misspell
     - revive
     - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
```
typeid, line 31: These are replaced by `unused`  by newer golangci-lint versions

PR #181, opened 2023-04-27
(WIP) Add command for checking cluster support credentials

AlexVulaj:
Adds a command to check for the presence of support credentials for a given cluster.

[OSD-16177](https://issues.redhat.com//browse/OSD-16177)

### DIFF
--- a/pkg/services/ccam/ccam.go
+++ b/pkg/services/ccam/ccam.go
@@ -3,6 +3,9 @@ package ccam
 
 import (
 	"fmt"
+	"github.com/openshift/configuration-anomaly-detection/pkg/aws"
+	"github.com/openshift/configuration-anomaly-detection/pkg/services/assumerole"
+	"os"
 	"regexp"
 
 	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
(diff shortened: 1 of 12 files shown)


cadctl/cmd/credentials/credentials.go
```diff
@@ -0,0 +1,73 @@
+package credentials
```
typeid, line 1: **suggestion:** I would probably rename the subfolder of this package to something like `periodic-checks`. Maybe you find a better term, the code can be used as non-periodic check, but I want to clearly separate it from all the stuff we have going on with PD routing.

cadctl/cmd/credentials/credentials.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/openshift/configuration-anomaly-detection/cadctl/cmd/investigate"
+	"github.com/openshift/configuration-anomaly-detection/pkg/services/assumerole"
+	"github.com/openshift/configuration-anomaly-detection/pkg/services/ccam"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+var Cmd = &cobra.Command{
+	Use:   "credentials",
+	Short: "Determines if a cluster's credentials are missing.",
```
typeid, line 14: **suggestion:** I know this is a WIP, but we need to have a clear plan of the following to finish the implementation:
- how do we want to trigger this check? (cronJob?)
- how many clusters do we want to check per check? How many times does cadctl run per pipeline?  Keep in mind pipelines are a limited resource (we have a hard cap)
- what passes the cluster ids to check to this check? The cronJob? Does CAD query the clusters itself (might make CAD harder to test)?



My suggestion:
- cronJob triggers every 24 hours, queries all clusters (in supported versions?) from OCM and starts a pipeline for every 100 clusters it queried from OCM.
- the credential check command takes the tekton pipeline payload (the list of ~100 clusters) as input, and performs the check for all these clusters. Keep in mind that the cluster states may already have altered between the query and the credentials checks. Some might already be uninstalled or currently uninstalling.

cadctl/cmd/credentials/credentials.go
```diff
@@ -0,0 +1,73 @@
...
+	Long: `Determines if a cluster's credentials are missing. If the command determines that they are,
+the cluster is placed into limited support. If the cluster is already in limited support, the command
+will evaluate if it the credentials have been restored and remove it from limited support if appropriate.`,
```
typeid, line 17:
```suggestion
will evaluate if the credentials have been restored and remove it from limited support if appropriate.`,
```

cadctl/cmd/credentials/credentials.go
```diff
@@ -0,0 +1,73 @@
...
+
+var clusterId string
+
+func init() {
+	Cmd.Flags().StringVarP(&clusterId, "cluster-id", "c", "", "OCM internal/external cluster id to run against.")
+	_ = Cmd.MarkFlagRequired("cluster-id")
+}
+
+func run(_ *cobra.Command, _ []string) error {
+	awsClient, err := investigate.GetAWSClient()
```
typeid, line 29: **thought:** We should probably move these public functions out of `investigate.go`, into pkg. (same for other `investigate.* functions`.

cadctl/cmd/credentials/credentials.go
```diff
@@ -0,0 +1,73 @@
...
+
+	supportRole, ok := os.LookupEnv("CAD_AWS_SUPPORT_JUMPROLE")
+	if !ok {
+		return fmt.Errorf("CAD_AWS_SUPPORT_JUMPROLE is missing")
+	}
+
+	cluster, err := ocmClient.GetClusterInfo(clusterId)
+	if err != nil {
+		return fmt.Errorf("could not retrieve cluster info for %s: %w", clusterId, err)
+	}
```
typeid, line 59: **chore:** duplicate code from `investigate.go` - this can be moved to a pkg.

cadctl/cmd/credentials/credentials.go
```diff
@@ -0,0 +1,73 @@
...
+
+	ccamClient := ccam.Client{
```
typeid, line 61: **issue:** This won't work, as the `Evaluate` function has API calls to pagerduty.

typeid (commented):
Awesome to see this progressing so quickly Alex! 
Generally this goes in the right direction, added some hopefully helpful comments to help you progress on this.

cadctl/cmd/credentials/credentials.go
AlexVulaj: Hey Claudio, I'm in full agreement with your suggestion here. I'm out on vacation all next week but I'll look into some of this stuff when I get back.

pkg/services/ccam/ccam.go
```diff
@@ -65,16 +68,49 @@ type Client struct {
...
+			OcmClient: ocmClient,
+		},
+	}
+
+	cssJumprole, ok := os.LookupEnv("CAD_AWS_CSS_JUMPROLE")
+	if !ok {
+		return aws.Client{}, fmt.Errorf("CAD_AWS_CSS_JUMPROLE is missing")
+	}
+
+	supportRole, ok := os.LookupEnv("CAD_AWS_SUPPORT_JUMPROLE")
```
boranx, line 87: Nice implementation, I've one small question
are there default values for these parameters? If so, it might be nice adding it or mentioning in the comments

AlexVulaj:
Closing this PR. Due to this work dropping in priority, it's fallen so far out of contact with the main branch of CAD that we're better off rewriting at this point when it gets reprioritized.

PR #182, opened 2023-05-09
feat(OSD-13700): add network verifier for chgm

ninabauer:
[OSD-13700](https://issues.redhat.com//browse/OSD-13700)

Runs network verifier on CHGM alerts after the instance checks and updates note on PD with results.


RaphaelBut:
Very important change, this looks great.

pkg/services/networkverifier/networkverifier.go
```diff
@@ -0,0 +1,221 @@
...
+		KmsKeyID:        config.kmsKeyID,
+		SecurityGroupId: securityGroupID,
+	}
+
+	awsVerifier, err := awsverifier.NewAwsVerifier(credentials.AccessKeyID, credentials.SecretAccessKey, credentials.SessionToken, c.cluster.Region().ID(), "", true)
+	if err != nil {
+		return Undefined, "", fmt.Errorf("could not build awsVerifier %v", err)
+	}
+
+	awsVerifier.Logger.Warn(context.TODO(), "Using region: %s", c.cluster.Region().ID())
```
typeid, line 175: **question:** Why do we use the `awsVerifier` logger here, instead of logging as per usual?

pkg/services/networkverifier/networkverifier.go
```diff
@@ -0,0 +1,221 @@
...
+
+	if len(verifierExceptions) != 0 && len(verifierErrors) != 0 {
+		exceptionsSummary := verifierExceptions[0].Error()
+		errorsSummary := verifierErrors[0].Error()
+		// AddNote
+		return Undefined, "", fmt.Errorf(exceptionsSummary, errorsSummary)
+	}
+
+	if !out.IsSuccessful() {
+		failureSummary := verifierFailures[0].Error() // create from verifierFailures
```
typeid, line 189: **Question:** Can we safely ignore the other values in the `verifierFailures` array? I think we might want to pick them up as well.

pkg/services/networkverifier/networkverifier.go
```diff
@@ -0,0 +1,221 @@
...
+
+	awsVerifier.Logger.Warn(context.TODO(), "Using region: %s", c.cluster.Region().ID())
+
+	out := verifier.ValidateEgress(awsVerifier, vei)
+
+	verifierFailures, verifierExceptions, verifierErrors := out.Parse()
+
+	if len(verifierExceptions) != 0 && len(verifierErrors) != 0 {
+		exceptionsSummary := verifierExceptions[0].Error()
+		errorsSummary := verifierErrors[0].Error()
```
typeid, line 183: **Question:** Can we safely ignore the other values in the `exceptionsSummary`, `errorsSummary` arrays? We might want to pick them up as well.

typeid (commented):
Overall great work Nina! Sorry for the very pedantic review, just a lot of suggestions and small nitpicks, nothing to worry about! :)

Big +1 on testing this to only add notes on the alerts first, and then move to automating LS once we feel safe about it!

cadctl/cmd/investigate/investigate.go
```diff
@@ -133,16 +134,23 @@ func run(_ *cobra.Command, _ []string) error {
...
 
 	investigationClient := investigation.Client{}
-
+	networkVerifierClient := networkverifier.Client{
+		Service: networkverifier.Provider{
+			AwsClient: &customerAwsClient,
+			OcmClient: &ocmClient,
+			PdClient:  &pdClient,
+		},
+	}
```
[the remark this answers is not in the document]
typeid: networkVerifierClient will be re-used in that context, so initializing it there makes sense.

pkg/services/networkverifier/networkverifier.go
typeid: This picks up everything.

pkg/services/networkverifier/networkverifier.go
typeid: This picks up everything.

PR #184, opened 2023-06-09
fix: network verifier bugs

ninabauer:
OSD-13700

Resolves issue of verifier result being Failure when successful, formats PD notes, adds servicelog command in notes.

### DIFF
--- a/pkg/services/networkverifier/networkverifier.go
+++ b/pkg/services/networkverifier/networkverifier.go
@@ -4,6 +4,7 @@ package networkverifier
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/services/networkverifier/networkverifier.go
```diff
@@ -97,8 +98,8 @@ type VerifierResult int
 // Verifier outcomes
 const (
 	Undefined VerifierResult = 0
-	Failure
-	Success
+	Failure   VerifierResult = 1
+	Success   VerifierResult = 0
```
typeid, line 102: **question:**
How does this even compile? :surprisedpikachu:  we have 2 `VerifierResults` with value 0

PR #185, opened 2023-06-15
[OSD-15971] Check cluster state before adding limited support

zmird-r:
Ticket: [OSD-15971](https://issues.redhat.com/browse/OSD-15971)

This replicates this check https://gitlab.cee.redhat.com/service/uhc-clusters-service/-/blob/master/pkg/osd/limited_support_reason_service.go#L103 to ensure limited support can be set.

### DIFF
--- a/pkg/services/chgm/chgm.go
+++ b/pkg/services/chgm/chgm.go
@@ -69,6 +69,7 @@ type Service interface {
 	GetClusterMachinePools(clusterID string) ([]*v1.MachinePool, error)
 	PostLimitedSupportReason(limitedSupportReason ocm.LimitedSupportReason, clusterID string) error
 	IsInLimitedSupport(clusterID string) (bool, error)
+	IsLimitedSupportAllowed(clusterID string) (bool, error)
 	UnrelatedLimitedSupportExists(ls ocm.LimitedSupportReason, clusterID string) (bool, error)
 	LimitedSupportReasonExists(ls ocm.LimitedSupportReason, clusterID string) (bool, error)
 	DeleteLimitedSupportReasons(ls ocm.LimitedSupportReason, clusterID string) error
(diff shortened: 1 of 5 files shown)


pkg/services/chgm/chgm.go
```diff
@@ -142,6 +143,15 @@ func (c *Client) Triggered() error {
...
 
+	// Check if limited support can be set on the cluster
+	allowed, err := c.IsLimitedSupportAllowed(c.Cluster.ID())
+	if err != nil {
+		return c.EscalateAlertWithNote(fmt.Errorf("failed to determine if the cluster can be set to limited support: %w", err).Error())
+	}
+	if !allowed {
+		return c.EscalateAlertWithNote("cluster state does not allow limited support to be set")
+	}
+
```
typeid, line 154: **Suggestion:**
I think we can move this logic into `c.PostLimitedSupport(res.String())` as we would otherwise have the code duplicated every time we use `c.PostLimitedSupport(res.String())`. 

---

With the current implementation,  `IsLimitedSupportAllowed` and `PostLimitedSupport` is also not atomic: 
```
1) IsLimitedSupportAllowed?
2) if yes, PostLimitedSupport
```

In a very rare edge case, the following could happen:
```
1) IsLimitedSupportAllowed? 
-> cluster goes into `uninstalling state` after the check but before posting the limited support
2) PostLimitedSupport fails
```

Suggestion would be to make this atomic by always running `c.PostLimitedSupport(res.String())` but checking the returned error. 

e.g. (untested)
```
func (c Client) PostLimitedSupportReason(limitedSupportReason LimitedSupportReason, clusterID string) error {
	fmt.Printf("Sending limited support reason: %s\n", limitedSupportReason.Summary)

	ls, err := c.newLimitedSupportReasonBuilder(limitedSupportReason).Build()
	if err != nil {
		return fmt.Errorf("could not create post request: %w", err)
	}

	request := c.conn.ClustersMgmt().V1().Clusters().Cluster(clusterID).LimitedSupportReasons().Add()
	request = request.Body(ls)
	resp, err := request.Send()
        // If the operation is not allowed because the cluster is uninstalling , we don't care and skip the limited support and return no error here.
	if err != nil && !strings.Contains(err, "Operation is not allowed for a cluster in 'uninstalling' state.") {
		return fmt.Errorf("received error from ocm: %w. Full Response: %#v", err, resp)
	}

	return nil
}
```

zmird-r: So you are suggesting to just skip and not return errors if the LS can not be set for this case? Don't you want to handle that and escalate?

zmird-r: For example here where the LS can not be determined, the approach is to escalate https://github.com/openshift/configuration-anomaly-detection/blob/main/pkg/services/chgm/chgm.go#L104

typeid: For the specific case that the cluster is already uninstalling, yes, I believe we should not return an error. What action could be taken if this escalates to primary? 
For other cases we should escalate after the retries.

PR #189, opened 2023-06-28
Refactor CAD structure

typeid:
**What?**

_Refactoring:_
- Removes the overuse of `service` `provider` `client` idiom
- Removes duplicate code
- Removes duplicate interfaces
- Interfaces are moved to their respective component
- Introduction of `investigation.Resources` to wrap all resources needed for investigations
- Restructure investigation endpoint 

_Bugfixes:_
- Fixes a panic in the network verifier run

**Why?**
This PR refactors the CAD structure to facilitate future implementations.


pkg/services/chgm/mock/interfaces.go
```diff
@@ -3,257 +3,3 @@
 
 // Package mock is a generated GoMock package.
 package mock
```
RaphaelBut, line 5: Can this file be removed completely or do we need it for something? 
We might need to change the makefile to keep it from being generated again.

pkg/services/networkverifier/mock/networkverifiermock.go
```diff
@@ -0,0 +1,5 @@
+// Code generated by MockGen. DO NOT EDIT.
```
RaphaelBut, line 1: I think this should be fine to remove aswell. Same as above.

pkg/services/networkverifier/networkverifier.go
```diff
@@ -7,75 +7,17 @@ import (
...
-	if c.ClusterDeployment == nil {
-		id := c.Cluster.ID()
-		c.ClusterDeployment, err = c.GetClusterDeployment(id)
-		if err != nil {
-			return fmt.Errorf("could not retrieve Cluster Deployment for %s: %w", id, err)
-		}
-	}
-	return nil
-}
+//go:generate mockgen --build_flags=--mod=readonly -source $GOFILE -destination ./mock/networkverifiermock.go -package networkverifiermock
```
RaphaelBut, line 20: If we remove the mock mentioned in the comment above, please remove this as well.
Otherwise ignore :)

RaphaelBut (commented):
This is amazing work, great job.

pkg/services/chgm/mock/interfaces.go
typeid: Good catch! those are just old leftover files that I didn't notice.

Makefile
```diff
@@ -100,7 +100,6 @@ checks:  check-duplicate-error-messages
 # check-duplicate-error-messages will conform with 
 .PHONY: check-duplicate-error-messages
 check-duplicate-error-messages:
-	@(test $$(grep -Ir errors.New . | grep -v -e .*.md | wc -l) -eq 0) || (echo "please use fmt.Errorf and not errors.New" >&2 && exit 1)
```
typeid, line 103: I don't know why we have this check. It makes no sense. There is no reason not to use `errors.New()`, it is the normal way to create unformatted errors.

typeid:
Change after review: use references everywhere of SdkClients

I re-tested CAD on a pagerduty alert afterwards and everything still works fine :)

PR #190, opened 2023-06-29
Add regex to LS check

zmird-r:
Skips all cases where the limited support cannot be set due to cluster state.

### DIFF
--- a/pkg/ocm/ocm.go
+++ b/pkg/ocm/ocm.go
@@ -210,7 +210,7 @@ func (c Client) PostLimitedSupportReason(limitedSupportReason LimitedSupportReas
 	request := c.conn.ClustersMgmt().V1().Clusters().Cluster(clusterID).LimitedSupportReasons().Add()
 	request = request.Body(ls)
 	resp, err := request.Send()
-    r := regexp.MustCompile("Operation is not allowed for a cluster in '[a-zA-z]+' state")
+	r := regexp.MustCompile("Operation is not allowed for a cluster in '[a-zA-z]+' state")
 	if err != nil && !r.MatchString(err.Error()) {
 		return fmt.Errorf("received error from ocm: %w. Full Response: %#v", err, resp)
 	}
--- a/pkg/ocm/ocm.go
+++ b/pkg/ocm/ocm.go
@@ -7,7 +7,7 @@ import (
 	"net/http"
 	"os"
 	"strconv"
-	"strings"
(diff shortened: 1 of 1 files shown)


pkg/ocm/ocm.go
```diff
@@ -210,7 +210,8 @@ func (c Client) PostLimitedSupportReason(limitedSupportReason LimitedSupportReas
 	request := c.conn.ClustersMgmt().V1().Clusters().Cluster(clusterID).LimitedSupportReasons().Add()
 	request = request.Body(ls)
 	resp, err := request.Send()
-	if err != nil && !strings.Contains(err.Error(), "Operation is not allowed for a cluster in 'installing' state") {
+	r := regexp.MustCompile("Operation is not allowed for a cluster in '[a-zA-z]+' state")
+	if err != nil && !r.MatchString(err.Error()) {
```
typeid, line 214: I think we should only cover cases we know. If we ever run into CAD trying to put limited support on a cluster that is in another state, e.g. pending, we would want CAD to run in an error and figure out why we are putting limited support on such a state.

zmird-r: Wouldn't that create more noise for primary? If the state changes during the investigation then it would be more about investigation OCM rather than CAD and trying to understand what's the flow for the cluster state in that specific case. Isn't it more worth having a fallback check for generic edge cases rather than investigating each one of them?

typeid: > Wouldn't that create more noise for primary? 
This is something that should never happen, and it is an unknown case for us. IMO If it does we want to be alerted on it. This generic change would hide the error.

typeid: @RaphaelBut WDYT?

RaphaelBut: not an easy one
Failing to post a limited support status will escalate the alert to primary, who will then see the error in the notes (is this correct? ) and should be able to deal with the alert easily, as CAD did find something it deemed worthy for limited support. 

Problem I think is, that we will lose the information that something happened which we did not expect.
 If the state switched to deleting, we could ignore of course, but I would say lets try to implement the cases we know can happen and deal with them explicitly.
That should also most likely happen outside the actual post Limited support as different alerts may have different actions depending on how the state changed.
Do we have to action the chgm/deletion ccam/pending ccam/deleting ccam/error cases differently?

Just throwing out some thoughts ;)

PR #193, opened 2023-06-30
Fix: nil check for confusing return value of jumpRoles

typeid:
**What?**
- The bug was that we miss-use the return value and need to double check the the returned value on top of the error for `jumpRoles`... This is not ideal, but golang doesn't really offer a better choice other than adding a third parameter, which is also confusing.
- Changed all the return values to nil instead of initializing an empty object and passing the pointer of it back (unrelated to bug)


RaphaelBut:
unlucky but nice try with using proper types :grin: 

/lgtm

PR #194, opened 2023-07-03
OSD-13697: Add ClusterProvisioningDelay investigation

typeid:
**What?**
- Adds CPD investigation - the investigation currently only adds networkverifier notes to the pagerduty alert

**Caveats:** 
1) The alertmanager configuration for the app-sre service, which contains CPD alerts has a very different formatting from the deadmanssnitch one. 

    This results in:
    - the need to parse the initial cluster id differently (from the title instead of from an alert note field)
    - the need to convert the above parsed id to an external ID to call later functions

  
2) We now also need to handle missing cloud credentials differently: if the cluster is not in ready state, we don't want to put the cluster in limited support (as it would otherwise eventually end up putting installing clusters in LS).


typeid:
Holding for testing. PR can already be reviewed.
/hold

typeid:
Tested the following:
- CPD investigation run
- CCAM run with a cluster in uninstalling state
- CHGM investigation

PR #195, opened 2023-07-04
OSD-16112: Improve linting

zmird-r:
Ticket: [OSD-16112](https://issues.redhat.com/browse/OSD-16112)

- Adds linters to the golangci-lint config
- Adds an editorconfig to enforce usage of tab over spaces
- Adds a pre-commit hook created by the Makefile

The following lint errors raised by https://github.com/nakabonne/nestif remain up for discussion:
```
cadctl/cmd/investigate/investigate.go:103:2: `if !cloudProviderSupported` has complex nested blocks (complexity: 5) (nestif)
	if !cloudProviderSupported {
	^
pkg/services/chgm/chgm.go:57:2: `if res.UserAuthorized` has complex nested blocks (complexity: 7) (nestif)
	if res.UserAuthorized {
	^
pkg/services/chgm/chgm.go:135:2: `if res.Error != ""` has complex nested blocks (complexity: 5) (nestif)
	if res.Error != "" {
```

We could set the complexity to a value of 10 to prevent complicated code while avoiding the necessity of refactoring the current codebase.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = false
+
+[*.go]
+indent_style = tab
+tab_width = 4
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 include project.mk
 include boilerplate/generated-includes.mk
 
-GOLANGCI_LINT_VERSION=v1.52.2
+GOLANGCI_LINT_VERSION=v1.53.3
+PRE_COMMIT_HOOK = .git/hooks/pre-commit
+PRE_COMMIT_SCRIPT = hack/pre-commit.sh
 
 # Binaries used in Makefile
 bin/cobra:
@@ -28,10 +30,11 @@ getlint:
 # Required as mockgen is running without a relative path
 export PATH := $(PWD)/bin:$(PATH)
 
+
 # Actions
 .DEFAULT_GOAL := all
 .PHONY: all
-all: build lint test generate-markdown
+all: build lint test generate-markdown pre-commit
 
 # build uses the following Go flags:
 # -s -w for stripping the binary (making it smaller)
@@ -68,7 +71,7 @@ test-with-coverage:
 cadctl-install-local: cadctl/cadctl
 
 .PHONY: cadctl-install-local-force
-cadctl-install-local-force: 
+cadctl-install-local-force:
 	rm cadctl/cadctl >/dev/null 2>&1 || true
 	make cadctl-install-local
 
@@ -77,9 +80,9 @@ cadctl-install-local-force:
 MARKDOWN_SOURCES := $(shell find $(SOURCEDIR) -name '*.md')
 .PHONY: generate-markdown
 generate-markdown: $(MARKDOWN_SOURCES) bin/embedmd
-	./bin/embedmd -w $(MARKDOWN_SOURCES) 
+	./bin/embedmd -w $(MARKDOWN_SOURCES)
 
-## CI actions 
+## CI actions
 
 .PHONY: coverage
 coverage: hack/codecov.sh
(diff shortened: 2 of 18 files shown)


.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = false
+
+[*.go]
+indent_style = tab
+tab_width = 4
```
typeid, line 11: **Question:** this comes on top/separately of gofmt, and I believe it requires installing an extension in VSCode. Do you think this will result in the opposite of what it's supposed to do: different formatting?

hack/pre-commit.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+ROOT_DIR=$(git rev-parse --show-toplevel)
+GOFILES=$(git diff --cached --name-only --diff-filter=ACM | grep --color=never -E '^.*\.go$')
```
typeid, line 4: **Question:** could we re-use this to only lint changed files on the pre-commit hook?

pkg/services/chgm/chgm.go
```diff
@@ -127,7 +125,7 @@ func InvestigateResolved(r *investigation.Resources) error {
 		})
 		if err != nil {
 			logging.Errorf("Failed to add note '%s' to incident: %s", investigationNotNeededNote, err)
-			return nil
+			return err
```
typeid, line 128: **thought:** I think this was nil because we don't care about the failed note add in this case, but it doesn't hurt to change it to err. I think it makes more sense now :)

pkg/services/chgm/chgm.go
```diff
@@ -301,10 +298,11 @@ func (i investigateInstancesOutput) string() string {
...
 	msg += fmt.Sprintf("\nNumber of expected instances:\n\tMaster: '%v'\n\tInfra: '%v'\n\tMin Worker: '%v'\n\tMax Worker: '%v'\n",
 		i.ExpectedInstances.Master, i.ExpectedInstances.Infra, i.ExpectedInstances.MinWorker, i.ExpectedInstances.MaxWorker)
-	var ids []string
-	for _, nonRunningInstance := range i.NonRunningInstances {
+	// var ids []string
+	ids := make([]string, len(i.NonRunningInstances))
+	for i, nonRunningInstance := range i.NonRunningInstances {
 		// TODO: add also the StateTransitionReason to the output if needed
-		ids = append(ids, *nonRunningInstance.InstanceId)
+		ids[i] = *nonRunningInstance.InstanceId
```
typeid, line 305: Nice preallocation!

Makefile
```diff
@@ -28,10 +30,11 @@ getlint:
 # Required as mockgen is running without a relative path
 export PATH := $(PWD)/bin:$(PATH)
 
+
 # Actions
 .DEFAULT_GOAL := all
 .PHONY: all
-all: build lint test generate-markdown
+all: build lint test generate-markdown pre-commit
```
typeid, line 37: **issue:** We lint twice in this target

typeid (commented):
Good stuff Zakaria! Added some small comments (mostly questions) but overall it's looking great!
Also: I'm fine raising the complexity to 10.

hack/pre-commit.sh
zmird-r: I tried that but golangci-lint only runs on packages, not on single files.

Makefile
zmird-r: I don't see how, the other targets do not contain linting.

typeid: `pre-commit` and `lint` both lint, no?

hack/pre-commit.sh
typeid: Ah that's unfortunate, but not a big deal.

Makefile
zmird-r: The pre-commit target just adds the hook to the `.git/hooks` dir, the lint script then runs only when you commit.

typeid: Ah... that indeed makes sense. Sorry for the confusion!

PR #196, opened 2023-07-05
Add remaining checks from 'osdctl cluster cpd' to cpd investigation

RaphaelBut:
This adds the rest of the checks that are run with 'osdctl cluster cpd' command to the cpd investigation for https://issues.redhat.com/browse/OSD-13697

Why: this will allow us to remove the highlighting for the cpd command in CPDS sop and eventually we should be able to deprecate it completely.

osdctl implementation https://github.com/openshift/osdctl/blob/master/cmd/cluster/cpd.go#L69 

I did not change the escalation behavior, so everything should still get escalated but with some more specific information.
Tested manually.
Please keep it a draft for now. I am still testing the BYOVPC subnet validation.

### DIFF
--- a/pkg/services/cpd/cpd.go
+++ b/pkg/services/cpd/cpd.go
@@ -65,14 +65,22 @@ func InvestigateTriggered(r *investigation.Resources) error {
 
 	if r.Cluster.AWS().SubnetIDs() != nil && len(r.Cluster.AWS().SubnetIDs()) > 0 {
 		logging.Info("Checking BYOVPC to ensure subnets have valid routing")
+		escalate := false
 		for _, subnet := range r.Cluster.AWS().SubnetIDs() {
 			isValid, err := isSubnetRouteValid(r.AwsClient, subnet)
 			if err != nil {
 				logging.Error(err)
 			}
 			if !isValid {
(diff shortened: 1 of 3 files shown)


RaphaelBut:
Tested the BYOVPC logic manually.
Can be reviewed :)

RaphaelBut:
/unhold 
Discussed the issue of having osdctl's code duplicated here and came to the conclusion to go ahead with it for now. 
CPD is not the regular case for osdctl as its more focused on providing general utility instead of automating complete sops. 
We agreed that for example the BYOVPC check should live upstream in the verifier but until that is implemented and/or our cpd investigation becomes obsolete by installer improvements we will have the code duplicated here.

We need to make sure we track any important changes or bug fixes in osdctl cluster cpd though.

pkg/services/cpd/cpd.go
```diff
@@ -64,3 +114,93 @@ func InvestigateTriggered(r *investigation.Resources) error {
...
+		Filters: []*ec2.Filter{
+			{
+				Name:   awsSdk.String("association.subnet-id"),
+				Values: []*string{awsSdk.String(subnetID)},
+			},
+		},
+	})
+	if err != nil {
+		return false, fmt.Errorf("failed to describe route tables associated to subnet %s: %w", subnetID, err)
+	}
```
typeid, line 132: **nitpick:**
We could wrap this into a function reflecting the above comment. This would make the code self explanatory and the function could move into the `aws.go` file. 
```
getSubnetRouteTables(subnetID string) ... {}
```

typeid (approved):
All nitpicks are non blocking, feel free to update if you believe it makes sense to do so.

Everything else looks good to me  - I'll lgtm and put a hold in case you want to change anything, if not, feel free to remove :)

typeid:
Still not fully sure about the ready check not returning, but it shouldn't hurt. 

/lgtm

PR #198, opened 2023-07-13
fix: OSD-17489 empty external cluster id

RaphaelBut:
Fixes bug with empty external cluster ids on CPDs.
We do not need to do another call to ocm to get the cloud provider as we already have the full cluster object at that point.

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -104,7 +104,7 @@ func run(_ *cobra.Command, _ []string) error {
 		return err
 	}
 
-	cloudProviderSupported, err := checkCloudProviderSupported(ocmClient, externalClusterID, []string{"aws"})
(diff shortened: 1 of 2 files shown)


cadctl/cmd/investigate/investigate.go
```diff
@@ -280,14 +280,14 @@ func GetPDClient(webhookPayload []byte) (*pagerduty.SdkClient, error) {
 
 // checkCloudProviderSupported takes a list of supported providers and checks if the
 // cluster to investigate's provider is supported
-func checkCloudProviderSupported(ocmClient *ocm.SdkClient, externalClusterID string, supportedProviders []string) (bool, error) {
-	cloudProvider, err := ocmClient.GetCloudProviderID(externalClusterID)
```
typeid, line 284: **nitpick (non-blocking):** We can remove this `ocmClient.GetCloudProviderID`  function as it's not unused.

typeid:
/hold

Remove hold if you believe we should keep the unused function.

PR #200, opened 2023-07-16
Build CPD pagerduty note to contain all checks

typeid:
**Why?**
We need to clarify what checks CAD runs, currently it just prints whether the network verifier passed or failed, which results in people running the `osdctl cpd` command.

Example new ouput:
Screenshot from 2023-07-16 11-13-30

### DIFF
--- a/pkg/services/cpd/cpd.go
+++ b/pkg/services/cpd/cpd.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/openshift/configuration-anomaly-detection/pkg/aws"
 	"github.com/openshift/configuration-anomaly-detection/pkg/investigation"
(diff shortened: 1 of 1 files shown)


pkg/services/cpd/cpd.go
```diff
@@ -41,36 +42,45 @@ func GetCPDAlertInternalID(alertTitle string) (string, error) {
...
+		notesSb.WriteString("⚠️ This cluster is in a ready state, thus provisioning succeeded\n.")
+		err := r.PdClient.AddNote(notesSb.String())
 		if err != nil {
 			logging.Error("could not add clusters ready state to incident notes")
 		}
 	}
+	notesSb.WriteString("✅ Cluster installation did not yet finish\n")
 
 	// Check if DNS is ready, exit out if not
 	if !r.Cluster.Status().DNSReady() {
```
typeid, line 60: **thought:** In this case we should probably let CAD resolve the alert, but I doubt this case will ever trigger.

RaphaelBut: Agreed but I would suggest an iterative approach, once the first SRE encounters it we can review the condition.

PR #205, opened 2023-07-18
Rename 'service' notion to 'investigation'

typeid:
**Why?**
Makes it less confusing as there's services in pd and services in CAD. `Investigation` makes much more sense.


RaphaelBut:
nice change! i like the naming way better

PR #207, opened 2023-07-20
Remove antipattern tests that test our imports

typeid:
**Why?**

- We have 40k lines of mocked code, just to test the AWS SDK. The AWS SDK is already tested by AWS, we should not test this - it's an antipattern.

**What?**
- Remove the 40k lines of mocked code
- Remove the few tests that basically re-tested the aws sdk, or tested error bubbling.


RaphaelBut:
ty, seems like a sound change to me :D

PR #208, opened 2023-07-20
OSD-17456 Add alert handling for hypershift SilenceAlerts

typeid:
**What?**
Implement [OSD-17456](https://issues.redhat.com//browse/OSD-17456).

CAD can now handle SilenceAlerts (hypershift inhibition alerts). It will resolve all old alerts when a SilenceAlert is triggered.

The next steps would be to add orchestration rules to all hypershift pd services that route only the SilenceAlerts to CAD.

**Why?**

We use alert inhibition as silencing method for hypershift. The issue with this is that inhibiting alerts does not result in active alerts being resolved. This is very frustrating e.g. when starting to work on an alert but the cluster ends up being mid uninstallation.


typeid:
/unhold

Checklist:
- some feedback on this - done
- look into how many new pipelines would get routed through CAD - done
- discuss this with the team - done
- test on live alerts - done

PR #209, opened 2023-07-26
Add metric instrumentation

RaphaelBut:
Adds metrics for https://issues.redhat.com/browse/OSD-12591 

Small change to pass investigation resources to jumpRoles to have the context we need to set labels on the metric

I only added some metrics from the top of my mind, so please see this as WIP and add any thought on metric names, what metrics to capture and so on.

Needs testing as well but the metrics should not interfere with anything.

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -156,8 +156,7 @@ func run(_ *cobra.Command, _ []string) error {
 
 	// Try to jump into support role
 	// This triggers a cloud-credentials-are-missing investigation in case the jumpRole fails.
-	investigationResources := &investigation.Resources{AlertType: alertType, Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: baseAwsClient, OcmClient: ocmClient, PdClient: pdClient}
-	customerAwsClient, err := jumpRoles(investigationResources)
+	customerAwsClient, err := jumpRoles(cluster, baseAwsClient, ocmClient, pdClient, alertTypeString)
 	if err != nil {
 		return err
 	}
@@ -167,7 +166,7 @@ func run(_ *cobra.Command, _ []string) error {
 		return nil
 	}
 
-	investigationResources.AwsClient = customerAwsClient
+	investigationResources := &investigation.Resources{AlertType: alertType, Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: customerAwsClient, OcmClient: ocmClient, PdClient: pdClient}
 
 	run := investigation.NewInvestigation()
 
@@ -200,7 +199,7 @@ func run(_ *cobra.Command, _ []string) error {
 }
 
 // jumpRoles will return an aws client or an error after trying to jump into support role
-func jumpRoles(r *investigation.Resources) (*aws.SdkClient, error) {
+func jumpRoles(cluster *v1.Cluster, baseAwsClient aws.Client, ocmClient ocm.Client, pdClient pagerduty.Client, alertType string) (*aws.SdkClient, error) {
 	cssJumprole, ok := os.LookupEnv("CAD_AWS_CSS_JUMPROLE")
 	if !ok {
 		return nil, fmt.Errorf("CAD_AWS_CSS_JUMPROLE is missing")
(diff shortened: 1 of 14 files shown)


pkg/metrics/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+// Push collects and pushes metrics to the configured pushgateway
+func Push() {
+	var promPusher *push.Pusher
+	if pushgateway := os.Getenv("CAD_PROMETHEUS_PUSHGATEWAY"); pushgateway != "" {
+		promPusher = push.New(pushgateway, "cad").Format(expfmt.FmtText)
+		promPusher.Collector(Alerts)
+		promPusher.Collector(LimitedSupportSet)
+		promPusher.Collector(LimitedSupportLifted)
+		promPusher.Collector(ServicelogPrepared)
+		err := promPusher.Add()
```
thavlice, line 29: Note that there two possible ways to get the metrics to the pushgateway:

- method `Push`
- method `Add`

The difference is that `Push` replace ALL  old metrics with new ones (updates the metrics that are in the new push and deletes those that are not),
whereas `Add` updates the old metrics with new and leaves the unmentioned as they are.


For our example there is no difference, as we are pushing ALL the metrics everytime.

**We should use** `Add` as it saves the previous value in case some metric is not collected or sent properly.

cadctl/cmd/investigate/investigate.go
```diff
@@ -83,15 +84,16 @@ func run(_ *cobra.Command, _ []string) error {
 
 	logging.Infof("Incident link: %s", pdClient.GetIncidentRef())
 
-	// After this point we know CAD will investigate the alert
-	// so we can initialize the other clients
-	alertType, err := isAlertSupported(pdClient.GetTitle())
-	if err != nil {
+	alertType := isAlertSupported(pdClient.GetTitle())
+	alertTypeString := alertType.String()
```
typeid, line 88: Nice refactoring there! Lovely :)

cadctl/cmd/investigate/investigate.go
```diff
@@ -221,33 +219,32 @@ func jumpRoles(cluster *v1.Cluster, baseAwsClient aws.Client, ocmClient ocm.Clie
...
+		return nil, ccam.Evaluate(err, r)
 	}
 	logging.Info("Successfully jumpRoled into the customer account. Removing existing 'Cloud Credentials Are Missing' limited support reasons.")
-	return customerAwsClient, ccam.RemoveLimitedSupport(cluster, ocmClient, pdClient)
+	return customerAwsClient, ccam.RemoveLimitedSupport(r)
 }
 
 // isAlertSupported will return the alertname as enum type of the alert if it is supported, otherwise an error
-func isAlertSupported(alertTitle string) (investigation.AlertType, error) {
+func isAlertSupported(alertTitle string) investigation.AlertType {
```
typeid, line 235: Good stuff! Makes it a lot less confusing.

pkg/investigations/ccam/ccam.go
```diff
@@ -93,6 +99,9 @@ func RemoveLimitedSupport(cluster *v1.Cluster, ocmClient ocm.Client, pdClient pa
 		logging.Info("Primary has been alerted")
 		return nil
 	}
+	if removedReason {
+		metrics.LimitedSupportLifted.WithLabelValues(r.AlertType.String(), r.PdClient.GetEventType(), ccamLimitedSupport.Summary).Inc()
+	}
```
typeid, line 104: **thought:** While I think it's great to decouple this from the logic of deleting the limited support reason, I'm wondering how this will evolve in the future as this currently allows forgetting to create/update a metric. 

A wrapper around the function to also increment metrics also seems wrong-ish. Not sure what the correct way to implement this is, but I'm totally fine leaving it decoupled.

typeid (changes requested):
Good stuff! Just one issue I see with passing the full `investigationResources` through the assumeRoleing.

cadctl/cmd/investigate/investigate.go
```diff
@@ -167,7 +164,8 @@ func run(_ *cobra.Command, _ []string) error {
 
 	// Try to jump into support role
 	// This triggers a cloud-credentials-are-missing investigation in case the jumpRole fails.
-	customerAwsClient, err := jumpRoles(cluster, baseAwsClient, ocmClient, pdClient)
+	investigationResources := &investigation.Resources{AlertType: alertType, Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: baseAwsClient, OcmClient: ocmClient, PdClient: pdClient}
+	customerAwsClient, err := jumpRoles(investigationResources)
```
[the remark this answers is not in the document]
RaphaelBut: Very good points. I ended up reverting it and passing only the additional alertType for populating the metric label in the ccam investigation. 
My thought here was to pass the whole context/investigationResources which included the alertType but introducing a possibility to mix up aws accounts here is very dangerous indeed :D

RaphaelBut:
Rebased

typeid:
/lgtm

As discussed, let's split up the assumeRole function in a followup PR.

RaphaelBut:
Opened PR for assumeRole https://github.com/openshift/configuration-anomaly-detection/pull/217

PR #211, opened 2023-08-01
Remove edge case code for clusterID fetching

typeid:
**What?**
We recently updated the PD service orchestrations and alertmanager configurations to all be consistent. The `cluster_id` is now **always** in alert.custom_details.cluster_id. Therefore, we no longer need edge case handling _per service_ to retrieve the cluster_id.

I renamed all occurences of `external id` to `cluster id`, as we have don't know which id is contained in the cluster_id field, per example for CPD it's the internal one, for CHGM it's the external one. This is already handled by CAD and always converted to internal_id in the first parts of the `investigation.go` code. 

I also had to extract custom details parsing from the `GetAlerts` function, as it made no sense to have it in there. We now have a function that gets alerts for an incident, and a function that extracts the needed details from the alerts. The `localAlert` objects were renamed, as they are not alerts, they are details of alerts (only clusterID and alertID).

**Why?**

https://issues.redhat.com/browse/OSD-17557

**Tests**

Tested for SilenceAlerts and CHGM.


typeid:
/hold

We should merge this after SilenceAlerts canary testing is complete.

pkg/pagerduty/pagerduty.go
```diff
@@ -477,55 +481,24 @@ func (c *SdkClient) GetAlerts(incidentID string) ([]Alert, error) {
...
-
-	internalBody := internalCHGMAlertBody{}
-
-	err = yaml.Unmarshal([]byte(externalBody), &internalBody)
-	if err != nil {
-		// TODO: add errcheck for this specific error
-		externalBody = strings.ReplaceAll(externalBody, `\n`, "\n")
-		err = yaml.Unmarshal([]byte(externalBody), &internalBody)
+// GetAlertListDetails will retrieve the required details for a list of alerts and return an array of alertDetails
+// in the same array slot sequence
```
RaphaelBut, line 488: polish: same order :D?

RaphaelBut (commented):
Nice change! much cleaner than before
Left one nag but thats not blocking.

PR #212, opened 2023-08-02
Silence CPD if OCM error is known and customer was notified via OCM

typeid:
**Why?**
https://issues.redhat.com/browse/OSD-17726

The case that OCM sent a SL is a noop for SRE, thus the alert is now being silenced instead of escalated.

### DIFF
--- a/pkg/investigations/cpd/cpd.go
+++ b/pkg/investigations/cpd/cpd.go
@@ -13,8 +13,6 @@ import (
 	"github.com/openshift/configuration-anomaly-detection/pkg/networkverifier"
 )
 
-const unknownProvisionCode = "OCM3999"
-
 // GetCPDAlertInternalID gets the internal ID from a CPD pagerduty alert's title
 // We need this function as CPD's alert formatting is an edge case to our other alerts.
 func GetCPDAlertInternalID(alertTitle string) (string, error) {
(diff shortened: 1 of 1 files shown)


pkg/investigations/cpd/cpd.go
```diff
@@ -65,8 +67,8 @@ func InvestigateTriggered(r *investigation.Resources) error {
 
 	// Check if the OCM Error code is a known error
 	if len(r.Cluster.Status().ProvisionErrorCode()) > 0 && r.Cluster.Status().ProvisionErrorCode() != unknownProvisionCode {
```
mjlshen, line 69: This does mean that it is a known error, but does not necessarily mean that SRE does not need to investigate it/that a SL has been sent. There some relatively generic errors which we define so that we can classify our CPDs, e.g. `GeneralOperatorDegraded` is classified as a known error code, but does not send a service log. Every CPD reaching PagerDuty should generally imply that an investigation is needed.

mjlshen (changes requested):
/hold

pkg/investigations/cpd/cpd.go
typeid: Ah, I see. I'll change the PR to instead make it clearer that SRE has to investigate anyway. Thanks for that key information!

PR #215, opened 2023-08-07
fix: nilptr deref in networkverifier package

typeid:
**What?**
Evaluate the errors (and evaluate them before further processing) in the network verifier package.

**Why?**

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x22be185]

goroutine 1 [running]:
github.com/openshift/configuration-anomaly-detection/pkg/networkverifier.initializeValidateEgressInput(0xc000858fc0, 0xc00017a5e0?, {0x310c460, 0xc000814280})
	github.com/openshift/configuration-anomaly-detection/pkg/networkverifier/networkverifier.go:34 +0x45
```


mjlshen (commented):
/lgtm

PR #221, opened 2023-08-29
[OSD-17702] stop removing limited support if cluster is constantly re…

zmird-r:
Ticket: [OSD-17702](https://issues.redhat.com/browse/OSD-17702)


pkg/ocm/ocm.go
```diff
@@ -237,9 +239,57 @@ func (c *SdkClient) LimitedSupportExists(ls LimitedSupportReason, internalCluste
 	return false, nil
 }
 
+// isLimitedSupportReasonFlapping checks if a cluster has been put in limited support too many times
+// Returns true if the cluster has been put in limited support at least 2 times in the last 4 hours
```
typeid, line 243: **question:** I wonder if this is a good value, is this based on the ticket/slack threads data? This value needs to catche the occurences where we have automation fighting the hive restarts.

pkg/ocm/ocm.go
```diff
@@ -237,9 +239,57 @@ func (c *SdkClient) LimitedSupportExists(ls LimitedSupportReason, internalCluste
...
+	isFlapping, err := c.isLimitedSupportReasonFlapping(ls, internalClusterID)
+	if err != nil {
+		return false, err
+	}
+
+	if isFlapping {
+		logging.Infof("Limited support reason is flapping `%s`", ls.Summary)
+		return false, nil
+	}
+
```
typeid, line 292: **issue:** `DeleteLimitedSupportReasons` currently does not document any of this behavior. 

**suggestion:** the flapping timespan could be a parameter of `DeleteLimitedSupportReasons`, and should be documented in the functions comment. Value `0` for the flappyness span could be "always delete the reason".

typeid (changes requested):

pkg/ocm/ocm.go
```diff
@@ -237,9 +239,57 @@ func (c *SdkClient) LimitedSupportExists(ls LimitedSupportReason, internalCluste
...
+		return false, fmt.Errorf("could not retrieve cluster info: %w", err)
+	}
+
+	externalID := info.ExternalID()
+
+	// Rerieve service logs
+	resp, err := c.conn.ServiceLogs().V1().Clusters().Cluster(externalID).ClusterLogs().List().Send()
+	if err != nil {
+		return false, fmt.Errorf("could not list service logs history: %w", err)
+	}
```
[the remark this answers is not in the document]
zmird-r: The service_logs service doesn't accept internal IDs. You can check for yourself:
`ocm get /api/service_logs/v1/clusters/$CLUSTER_ID/cluster_logs `

pkg/ocm/ocm.go
```diff
@@ -237,9 +239,57 @@ func (c *SdkClient) LimitedSupportExists(ls LimitedSupportReason, internalCluste
...
+
+	serviceLogs, ok := resp.GetItems()
+	if !ok {
+		return false, nil
+	}
+
+	// Check if the cluster has been put in limited support at least 2 times in the last 4 hours
+	fourHoursAgo := time.Now().Add(-4 * time.Hour)
```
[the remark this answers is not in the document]
zmird-r: I don't think that the 2 functions are similar, they have different argument types and return different value types.
```
func (t Time) Add(d Duration) Time
func (t Time) Sub(u Time) Duration
```

pkg/ocm/ocm.go
zmird-r: I reviewed again the past cases and upped it to 24 hours. I think it should cover it better now.

pkg/ocm/ocm.go
zmird-r: Can you elaborate more on why we need it as a parameter and maybe give an example of a case in which you want to add and remove limited support multiple times?

pkg/ocm/ocm.go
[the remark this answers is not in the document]
typeid: TIL, that seems like a gap in ocm. As long as we don't call this on any clusters that don't have an external ID yet (e.g. not fully installed), we should be fine.

pkg/ocm/ocm.go
typeid: Currently, a customer can trigger CHGM by stopping nodes. The main reason we need this change is that a the nodes stopped by the customer usually get restarted by hive. This means that if the customer has automation in place to stop nodes (e.g. some AWS lambda), we have hive fighting the lambda, and the customer is spammed by limited support emails.

For other alerts/investigations (in the future), we should not have that problem, and we could default to putting clusters in and out of limited support purely based on the actions of the customer. 

Adding the parameter also makes it clearer to the caller that this is happening "under the hood". 

If you don't want to add this as parameter, I'm also okay with just documenting the behavior clearly in `DeleteLimitedSupportReasons`.

pkg/ocm/ocm.go
```diff
@@ -237,9 +239,57 @@ func (c *SdkClient) LimitedSupportExists(ls LimitedSupportReason, internalCluste
...
+
+	var filteredLogs []*servicelogsv1.LogEntry
+	for _, log := range serviceLogs.Slice() {
+		if log.ServiceName() == "LimitedSupport" &&
+			log.Username() == "service-account-ocm-cad-production" &&
```
typeid, line 270: One more small question: should we add staging here as well, so that it behaves the same on staging and prod for testing?

pkg/ocm/ocm.go
zmird-r: I documented the behavior, I think the parameter can be easily added when it is needed.

pkg/ocm/ocm.go
zmird-r: I also added the staging account to the condition.

typeid:
Good stuff!
/lgtm

PR #223, opened 2023-09-05
fix: check if alert is supported before decoding cluster_id

typeid:
**What?**
- Moves logic to decode `cluster_id` after checking if the alert is supported by CAD. 

**Why?**
Some unsupported alerts don't contain the `cluster_id`, e.g. the AVO alerts on the `app-sre-alertmanager` service. This causes CAD pipelines to exit ungracefully. 
Therefore, we need to move the logic to decode the cluster_id after the check if an alert is supported.


RaphaelBut:
Nice catch!
/lgtm

PR #225, opened 2023-09-07
Bug Fix: Wrong CHGM node count 

zmird-r:
**Bug fix:**

The node count is wrong on every CHGM alert. Example:

```
Is user authorized: 'true' 

UserName : '<username>' 

Number of non running instances: '1' 

Number of running instances:
	Master: '3'
	Infra: '2'
	Worker: '2'

Number of expected instances:
	Master: '3'
	Infra: '2'
	Min Worker: '4'
	Max Worker: '4'
```

Recently the machine pool `worker` has been added to the machine pool list and contains the default workers count. We don't need to call the `GetCompute` or the `GetAutoscaleCompute` functions on the `ClusterNodes` object, we can just count machine pool nodes.

Every cluster now has the following machine pool:
```
 {
      "kind": "MachinePool",
      "href": "/api/clusters_mgmt/v1/clusters/<cluster-id>/machine_pools/worker",
      "id": "worker",
      "autoscaling": {
        "min_replicas": 2,
        "max_replicas": 3
      },
      "instance_type": "r5.xlarge",
      "availability_zones": [
         "us-east-1"
      ],
      "subnets": [
         "subnet-example"
      ],
      "root_volume": {
        "aws": {
          "size": 128
        }
      }
    }
```


RaphaelBut:
Hey, nice change Zakaria! Way cleaner.
/lgtm

PR #226, opened 2023-09-14
feat(OSD-18454): automate invalid routing SL for CPD

typeid:
**What?**

This PR automates sending service logs for CPDs due to misconfigured routes.

https://issues.redhat.com/browse/OSD-18454


pkg/investigations/cpd/cpd.go
```diff
@@ -35,6 +39,7 @@ func InvestigateTriggered(r *investigation.Resources) error {
 		if err != nil {
 			logging.Error("could not add clusters ready state to incident notes")
 		}
+		return r.PdClient.EscalateAlert()
```
typeid, line 42: We should escalate this in case it happens, so that we're made aware of it.

typeid:
Removed the UUID, we don't need it. We can post SLs directly to the internal ID, but our SRE tooling (osdctl) has a bug that prevents them from showing.

PR #227, opened 2023-09-15
[OSD-18424] Add new check if cluster recently woke up.

bergmannf:
When clusters are in hibernation for >= 30 days the internal certificates used by kubelet will have expired and CSRs must be approved by hand to get the cluster into a healthy state again.

### DIFF
--- a/pkg/investigations/chgm/chgm.go
+++ b/pkg/investigations/chgm/chgm.go
@@ -55,6 +55,17 @@ func InvestigateTriggered(r *investigation.Resources) error {
 		return r.PdClient.SilenceAlertWithNote(res.string() + "Unrelated limited support reason present on cluster, silenced.")
 	}
 
+	longHibernation, hibernationErr := investigateHibernation(r.Cluster, r.OcmClient)
+	if hibernationErr != nil {
+		logging.Warn(fmt.Errorf("Could not check hibernation status of cluster: %w", err))
+	}
+	if longHibernation {
+		err = r.PdClient.AddNote(fmt.Sprintf("⚠️ Cluster was hibernated more than %.0f days - investigate CSRs and kubelet certificates: see https://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_gone_missing.md#24-hibernation", hibernationTooLong.Hours()/24))
+		if err != nil {
+			logging.Error("could not add hibernation comment to incident notes")
+		}
+	}
+
 	if res.UserAuthorized {
 		logging.Info("The customer has not stopped/terminated any nodes.")
 		// Run network verifier
@@ -66,6 +77,7 @@ func InvestigateTriggered(r *investigation.Resources) error {
 			if err != nil {
 				logging.Error("could not add failure reason incident notes")
 			}
+
 			return r.PdClient.EscalateAlertWithNote(res.string())
 		}
 
@@ -208,6 +220,22 @@ func investigateRestoredCluster(r *investigation.Resources) (res investigateInst
 	}
 
 	return res, nil
+}
+
+// investigateHibernation checks if the cluster was recently woken up from
+// hibernation. If clusters are hibernated for more than 30 days, the internal
+// certificates of the kubelets can expire and CSRs need to be approved
+// manually:
+// - https://github.com/openshift/hive/blob/master/docs/hibernating-clusters.md
+func investigateHibernation(cluster *v1.Cluster, client ocm.Client) (bool, error) {
+	hibernations, err := getHibernationStatusForCluster(client, cluster)
+	if err != nil {
+		return false, err
+	}
+	if len(hibernations) == 0 {
+		return false, nil
+	}
+	return hibernatedTooLong(hibernations, time.Now()), nil
 }
 
 // isUserAllowedToStop verifies if a user is allowed to stop/terminate instances
--- a/pkg/investigations/chgm/chgm_hibernation_check.go
+++ b/pkg/investigations/chgm/chgm_hibernation_check.go
@@ -1,0 +1,89 @@
+package chgm
+
+import (
+	"sort"
+	"time"
+
+	v1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	servicelogsv1 "github.com/openshift-online/ocm-sdk-go/servicelogs/v1"
+	"github.com/openshift/configuration-anomaly-detection/pkg/ocm"
+)
+
+const recentWakeupTime = 2 * time.Hour
+
+// 30 Days is always a problem as kubelet certificates will be expired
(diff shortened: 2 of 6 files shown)


pkg/investigations/chgm/chgm_hibernation_check.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+const RECENT_WAKEUP_TIME = 2 * time.Hour
+
+// 30 Days is always a problem as kubelet certificates will be expired
+const HIBERNATION_TOO_LONG = 30 * 24 * time.Hour
+const HIBERNATION_START_EVENT = "cluster_state_hibernating"
+const HIBERNATION_END_EVENT = "cluster_state_ready"
+const HIBERNATION_ONGOING_EVENT = "cluster_state_hibernating"
+const HIBERNATION_RESUME_EVENT = "cluster_state_resuming"
```
typeid, line 19: **Question (non-blocking):** would it make sense to keep those consts private to the package instead?

pkg/investigations/chgm/chgm_hibernation_check.go
```diff
@@ -0,0 +1,87 @@
...
+
+type HibernationPeriod struct {
```
typeid, line 21: **Question:** why does this have public?

pkg/investigations/chgm/chgm_hibernation_check.go
```diff
@@ -0,0 +1,87 @@
...
+	HibernationDuration time.Duration
+	DehibernationTime   time.Time
+}
+
+func HibernatedTooLong(hibernations []*HibernationPeriod, now time.Time) bool {
```
typeid, line 26: **Question:** Same as above, does this have public?

pkg/investigations/chgm/chgm_hibernation_check.go
```diff
@@ -0,0 +1,87 @@
...
+func getHibernationStatusForCluster(ocmClient ocm.Client, cluster *v1.Cluster) ([]*HibernationPeriod, error) {
+	filter := "log_type='cluster-state-updates'"
+	clusterStateUpdates, err := ocmClient.GetServiceLog(cluster, filter)
+	if err != nil {
+		return nil, err
+	}
+	return CreateHibernationTimeLine(clusterStateUpdates.Items().Slice()), nil
+}
+
+func CreateHibernationTimeLine(clusterStateUpdates []*servicelogsv1.LogEntry) []*HibernationPeriod {
```
typeid, line 53: **Question:** same question about it being public :D

pkg/investigations/chgm/chgm_hibernation_check_test.go
```diff
@@ -0,0 +1,119 @@
+package chgm
```
typeid, line 1: Nice tests! :)

typeid (commented):
Looking very good, just one small error check that I'd add and a few open questions about encapsulation :)

pkg/investigations/chgm/chgm.go
```diff
@@ -55,6 +55,14 @@ func InvestigateTriggered(r *investigation.Resources) error {
 		return r.PdClient.SilenceAlertWithNote(res.string() + "Unrelated limited support reason present on cluster, silenced.")
 	}
 
+	longHibernation, hibernationErr := investigateHibernation(r.Cluster, r.OcmClient)
+	if hibernationErr != nil {
+		logging.Info(fmt.Errorf("Could not check hibernation status of cluster: %w", err))
+	}
+	if longHibernation {
+		err = r.PdClient.AddNote(fmt.Sprintf("Cluster was hibernated more than %.0f days - investigate CSRs and kubelet certificates.", HIBERNATION_TOO_LONG.Hours()/24))
```
zmird-r, line 63: One small suggestion: I see that we have an SOP now, could we maybe link to it in the notes since it provides the resolution steps? https://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_gone_missing.md#24-hibernation

pkg/investigations/chgm/chgm_hibernation_check.go
bergmannf: Ah yes - I guess `UPCASE` for all constants, is just something I have a hard time not doing, due to all the other languages, but I can put those private.

bergmannf:
/hold
I just realized that under the slew of output for the ginkgo tests, I totally forgot that some of my own tests are still failing.

pkg/investigations/chgm/chgm_hibernation_check.go
bergmannf: I kind of forgot that all of those are in the same package, so I made all of them private now.

pkg/investigations/chgm/chgm_hibernation_check.go
bergmannf: I kind of forgot that all of those are in the same package, so I made all of them private now.

pkg/investigations/chgm/chgm.go
```diff
@@ -55,6 +55,17 @@ func InvestigateTriggered(r *investigation.Resources) error {
 		return r.PdClient.SilenceAlertWithNote(res.string() + "Unrelated limited support reason present on cluster, silenced.")
 	}
 
+	longHibernation, hibernationErr := investigateHibernation(r.Cluster, r.OcmClient)
+	if hibernationErr != nil {
+		logging.Warn(fmt.Errorf("Could not check hibernation status of cluster: %w", err))
+	}
+	if longHibernation {
+		err = r.PdClient.AddNote(fmt.Sprintf("Cluster was hibernated more than %.0f days - investigate CSRs and kubelet certificates: see hhttps://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_gone_missing.md#24-hibernation", hibernationTooLong.Hours()/24))
```
typeid, line 63:
```suggestion
		err = r.PdClient.AddNote(fmt.Sprintf("⚠️ Cluster was hibernated more than %.0f days - investigate CSRs and kubelet certificates: see hhttps://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_gone_missing.md#24-hibernation", hibernationTooLong.Hours()/24))
```

Just so that this is easier to catch when looking at the pagerduty alert. Text without flashy emojis there tends to get overlooked, especially with the wall of text we post for CHGM.

zmird-r, line 63:
```suggestion
		err = r.PdClient.AddNote(fmt.Sprintf("Cluster was hibernated more than %.0f days - investigate CSRs and kubelet certificates: see https://github.com/openshift/ops-sop/blob/master/v4/alerts/cluster_has_gone_missing.md#24-hibernation", hibernationTooLong.Hours()/24))
```

PR #229, opened 2023-09-27
feat(OSD-18650): [stopgap] support alerts from clusters with new backplane flow

typeid:
https://issues.redhat.com/browse/OSD-18650

**What?**
- this adds a check in CAD for clusters using the new flow and escalates them if the clusters aren't in limited support

**Why?**
- new AWS backplane flow will be promoted to production this week or the following, we do not support this flow so we must make sure that we don't accidentally put clusters in limited support due to missing access

### DIFF
--- a/pkg/ocm/ocm.go
+++ b/pkg/ocm/ocm.go
@@ -3,6 +3,7 @@ package ocm
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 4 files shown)


pkg/ocm/ocm.go
```diff
@@ -432,3 +433,46 @@ func (c *SdkClient) listLimitedSupportReasons(internalClusterID string) ([]*v1.L
...
 	return reasons.Items().Slice(), nil
 }
+
+func (c *SdkClient) getOrgfromClusterID(cluster *v1.Cluster) (string, error) {
+	subID, ok := cluster.Subscription().GetID()
+	if !ok {
+		return "", fmt.Errorf("failed getting subcription id from cluster object")
+	}
+
+	resp, err := c.conn.AccountsMgmt().V1().Subscriptions().List().Search(fmt.Sprintf("id like '%s'", subID)).Size(1).Send()
```
davidleerh, line 443: I was thinking it'd be simpler to use the following instead:
`resp, err := c.conn.AccountsMgmt().V1().Subscriptions().Subscription(subID).Get().Send()`
That way, we can remove the checks (slice, length etc...) below. WDYT?

typeid: That indeed would have been easier, thanks for pointing that out. 

I refactored to use the different api endpoint and base the check on the cluster only (not org).

mjlshen (approved):
/lgtm

PR #230, opened 2023-09-28
Set correct metric endpoint env

RaphaelBut:
Naming things is hard.


typeid (approved):
It indeed is hard... :D 

/lgtm

PR #237, opened 2023-10-06
dashboard: fix SLO, change default to prod and 30d

RaphaelBut:
Fixes the SLO query by selecting only cad-check pods. Otherwise we are off by a few pods (gateway, pruner...).
Updates the defaults to production environment with a 30d time range


RaphaelBut:
/hold cancel
Good to go

PR #239, opened 2023-10-26
fix(OSD-19366): check cluster state before proceeding with an investigation

typeid:
**What?** 
This PR fixes https://issues.redhat.com/browse/OSD-19366.

Changes: 
- added a check before running any investigations to make sure the cluster is in a state where we can run an investigation (currently just not "uninstalling").
- refactored the main logic to move the bigger pre-check if a cluster can have investigated into a new function

**Why?** 
Summary of the issue:
- When a cluster is mid-uninstall, the OCM API does not contain any `cluster deployment` objects. This resulted in CAD throwing 
  ```{
    "level": "fatal",
    "timestamp": "2023-10-21T00:59:25Z",
    "msg": "could not retrieve Cluster Deployment for clusterid: failed to unmarshal client response () with error: unexpected end of JSON input",
    "cluster_id": "clusterid",
    "pipeline_name": "cad-check-bf7d4927-0c73-458b-aea3-825d600045fe"
  } 
  ```
- This only happened for CHGM resolve events (the alerts resolved off clusters uninstalling).

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -79,7 +79,7 @@ func run(_ *cobra.Command, _ []string) error {
 	alertTypeString := alertType.String()
 	metrics.Inc(metrics.Alerts, alertTypeString, pdClient.GetEventType())
 
-	// handle unsupported alerts
+	// Escalate all unsupported alerts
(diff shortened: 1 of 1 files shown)


NautiluX (commented):
/lgtm

PR #243, opened 2023-11-07
fix(OSD-19596): skip aws access compatibility check for non-sts clusters

typeid:
**What?**
- Skip the CAD aws access compatibility check for non sts clusters (this was causing pipeline failures, as non-sts clusters have no sts support role).
- Reorders the logic if an investigation should be started on a cluster. The previous logic was getting too complicated with the additional STS logic, so I split it up which causes code duplication but is easier to read.

**Why?**

```
could not query sts support role from ocm: status is 400, identifier is '400', code is 'CLUSTERS-MGMT-400' and operation identifier is '123456': Retrieving SRE support role arn is only available for AWS STS clusters only
```

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -240,72 +240,64 @@ func GetPDClient(webhookPayload []byte) (*pagerduty.SdkClient, error) {
 	return client, nil
 }
 
-// checkCloudProviderSupported takes a list of supported providers and checks if the
-// cluster to investigate's provider is supported
-func checkCloudProviderSupported(cluster *v1.Cluster, supportedProviders []string) (bool, error) {
+func isAWS(cluster *v1.Cluster) (bool, error) {
 	cloudProvider, ok := cluster.GetCloudProvider()
 	if !ok {
 		return false, fmt.Errorf("Failed to get clusters cloud provider")
 	}
 
-	for _, provider := range supportedProviders {
-		if cloudProvider.ID() == provider {
-			return true, nil
-		}
-	}
-
-	logging.Infof("Unsupported cloud provider: %s", cloudProvider.ID())
-	return false, nil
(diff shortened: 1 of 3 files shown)


cadctl/cmd/investigate/investigate.go
```diff
@@ -240,72 +240,64 @@ func GetPDClient(webhookPayload []byte) (*pagerduty.SdkClient, error) {
 	return client, nil
 }
 
-// checkCloudProviderSupported takes a list of supported providers and checks if the
-// cluster to investigate's provider is supported
-func checkCloudProviderSupported(cluster *v1.Cluster, supportedProviders []string) (bool, error) {
+func isAWS(cluster *v1.Cluster) (bool, error) {
```
typeid, line 243: Simplified this, I don't think we will need the extensability for now, and it make the code after this more readable.

pkg/ocm/ocm.go
```diff
@@ -436,8 +436,14 @@ func (c *SdkClient) listLimitedSupportReasons(internalClusterID string) ([]*v1.L
 }
 
 // AwsClassicJumpRoleCompatible check whether or not the CAD jumprole path is supported by the cluster
-func (c *SdkClient) AwsClassicJumpRoleCompatible(clusterID string) (bool, error) {
-	resp, err := c.conn.ClustersMgmt().V1().Clusters().Cluster(clusterID).StsSupportJumpRole().Get().Send()
+func (c *SdkClient) AwsClassicJumpRoleCompatible(cluster *v1.Cluster) (bool, error) {
+	// If the cluster is STS, check if it compatible.
+	if cluster.AWS().STS().Empty() {
```
Tof1973, line 441: I suspect AWS() can be null. Wouldn't it be an issue for further calls ?

Tof1973: Forget about it, I guess that is ok ;)

PR #246, opened 2023-11-17
OSD-19710 Replace node shutdown limited support with service log

typeid:
https://issues.redhat.com/browse/OSD-19710

This PR refactors CAD as follows:
- CAD no longer removes limited support
- CAD no longer puts clusters in limited support due to node shutdowns

With these two changes:
- we no longer need to route resolving alerts through CAD
- we no longer need to assume a cluster might be in limited support when an alert reaches CAD

Instead of putting cluster into limited support for shut down nodes, CAD now sends a service log in case customer stopped instances, as this is also what an SRE would do. 

Reasoning why we can replace limited support with a service log here:
- there is no additional value in limited support over a service log, as a cluster with an active CHGM alert  due to shut down infrastructure can't alert SRE (no active alertmanager pods)
- limited support also didn't "soak up" any alerts that were triggered as a side effect of the CHGM, as has been determined as part of https://github.com/openshift/configuration-anomaly-detection/pull/245 

Further changes:
- cleanup: CHGM note is generated via stringbuilding


test/ocm_test.json
```diff
@@ -1,10 +0,0 @@
-{
```
typeid, line 1: This file is unused

cadctl/cmd/investigate/investigate.go
```diff
@@ -255,45 +254,30 @@ func isAWS(cluster *v1.Cluster) (bool, error) {
 // - the AWS account access flow is supported by CAD
 // Performs according pagerduty actions and returns whether CAD needs to investigate the cluster
 func clusterRequiresInvestigation(cluster *v1.Cluster, ocmClient *ocm.SdkClient, pdClient *pagerduty.SdkClient) (bool, error) {
+	// 1. Cluster is not in uninstalling state
 	if cluster.State() == v1.ClusterStateUninstalling {
```
typeid, line 258: This might not even happen anymore. This usually happened for CHGMs resolving due to cluster uninstalls only. Doesn't hurt keeping it here though, we might hit this in rare cases with other alerts as well.

pkg/investigations/ccam/ccam.go
```diff
@@ -86,33 +73,11 @@ func RemoveLimitedSupport(cluster *v1.Cluster, ocmClient ocm.Client, pdClient pa
...
+		return err
 	}
 	if removedReason {
 		metrics.Inc(metrics.LimitedSupportLifted, alertType, pdClient.GetEventType(), ccamLimitedSupport.Summary)
 	}
 	return nil
 }
-
-// buildAlertForCCAM will return a NewAlert populated with cluster id and the specific error
-func buildAlertForCCAM(lsError error, clusterID string) pagerduty.NewAlert {
```
typeid, line 108: Paging primary with this might be more confusing than just letting it reach primary the normal way. This basically never happens, and when it does it would just cause confusion as to what has and what hasn't been done here.

pkg/investigations/chgm/chgm.go
```diff
@@ -21,88 +21,77 @@ import (
...
+	ServiceName:  "SREManualAction",
+	Description:  "Your cluster is no longer checking in with Red Hat OpenShift Cluster Manager. Possible causes include stopped instances or a networking misconfiguration. If you have stopped the cluster instances, please start them again - stopping instances is not supported. If you intended to terminate this cluster then please delete the cluster in the Red Hat console",
+	InternalOnly: false,
 }
 
-// CAUTION
-
 // InvestigateTriggered runs the investigation for a triggered chgm pagerduty event
 func InvestigateTriggered(r *investigation.Resources) error {
+	var notesSb strings.Builder
```
typeid, line 34: In this function I did the following:
- inverted the logic to return early when user is not authorized
- used stringbuilder to build a nice note for primary and have the note in line with the CPD one (which is way less complex to understand)

pkg/investigations/ccam/ccam.go
```diff
@@ -70,49 +49,7 @@ func Evaluate(cluster *v1.Cluster, awsError error, ocmClient ocm.Client, pdClien
...
-		// E.g. we land here if we run into a CPD alert where credentials were removed (installing state)
+		// E.g. we land here if we run into a CPD alert where credentials were removed (installing state) and don't want to put it in LS yet.
 		return pdClient.EscalateAlertWithNote(fmt.Sprintf("Cluster has invalid cloud credentials (support role/policy is missing) and the cluster is in state '%s'. Please investigate.", cluster.State()))
 	}
 }
-
-// RemoveLimitedSupport will remove any CCAM limited support reason from the cluster,
-// if it fails to do so, it will try to alert primary
-// Run this after cloud credentials are confirmed
-func RemoveLimitedSupport(cluster *v1.Cluster, ocmClient ocm.Client, pdClient pagerduty.Client, alertType string) error {
```
RaphaelBut, line 81: question: i was under the impression, we keep the CCAM LS logic as is for now. 
With this change we will continue to put clusters into LS for ccam, but do not remove it anymore.

cadctl/cmd/investigate/investigate.go
```diff
@@ -138,20 +138,13 @@ func run(_ *cobra.Command, _ []string) error {
...
-	}
-
 	investigationResources := &investigation.Resources{AlertType: alertType, Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: customerAwsClient, OcmClient: ocmClient, PdClient: pdClient}
 
 	run := investigation.NewInvestigation()
 
 	switch alertType {
 	case investigation.ClusterHasGoneMissing:
 		run.Triggered = chgm.InvestigateTriggered
-		run.Resolved = chgm.HandleResolved
```
RaphaelBut, line 154: when we roll out this change to servicelogs, do we need a transition to remove all the LS reasons that are currently applied to triggered chgm's?

edit: just checked https://github.com/openshift/configuration-anomaly-detection/pull/245/files#diff-f8afe34e2c404d34ac56578a6d1617797fd3029613be45ff0f688b21f315f5e3R477 
My understanding is, we will blindly remove LS from resolving chgm. This will slowly remove LS reasons from "old" chgm alerts, regardless of actual LS. Once we no longer have old chgm we can remove the LS removal logic as well. 

Might be worth to keep the handleResolved here (should not interfere with the rest of the change?), switch to servicelogs and remove this line further down the line, to catch as many of the old alerts as possible?
However, removing it would work as well with the caveat that we increase manual LS removal load for MCS.

pkg/investigations/ccam/ccam.go
RaphaelBut: i think we need to explicitly think about ccam as well. We mainly came from a perspective of removing LS automation for chgm. If we keep posting LS on ccam, my feeling is we need at least some possibility of removing it automatically.

pkg/investigations/chgm/chgm.go
RaphaelBut: praise

RaphaelBut (commented):
awesome work, thank you!

cadctl/cmd/investigate/investigate.go
```diff
@@ -138,20 +138,13 @@ func run(_ *cobra.Command, _ []string) error {
 		return ccam.Evaluate(cluster, err, ocmClient, pdClient, alertTypeString)
 	}
 
-	logging.Info("Successfully jumpRoled into the customer account. Removing existing 'Cloud Credentials Are Missing' limited support reasons.")
```
RaphaelBut, line 141: i think we need to explicitly think about ccam as well. We mainly came from a perspective of removing LS automation for chgm. If we keep posting LS on ccam, my feeling is we need at least some possibility of removing it automatically.

pkg/investigations/ccam/ccam.go
typeid: We have https://issues.redhat.com/browse/SDE-2424 for cyclic CCAM setting and removals with EDA. I think this is the cleaner streamlined way. 

The reason I want to keep LS for CCAM for now is that it often isn't the root to the issue resulting in a page. Sending a SL without telling the customer their cluster is not being supported doesn't match our action of sending and muting the alert.

pkg/investigations/ccam/ccam.go
typeid: I think this might be our current best option, see above comment. Possibly removing a LS reason that is not really related to an alert when an alert resolves is kind of odd behaviour.

cadctl/cmd/investigate/investigate.go
typeid: > However, removing it would work as well with the caveat that we increase manual LS removal load for MCS.
Given that most CHGM alerts are very likely clusters that have already been wiped and/or internal e2e clusters, I think this variant is totally fine.

typeid:
@RaphaelBut fyi I'm planning to hold this until after we release CAD with the new backplane flow. I just wanted to open this to get an overview of what needs to be done.

pkg/investigations/ccam/ccam.go
RaphaelBut: Agreed that it needs to be LS and pulled out of chgm logic. My concern was with having no automatic removal anymore, increasing the load for manual removal. But after this change we will have different behavior anyway: Not checking the aws account on resolving chgms -> we cannot tell if we have access -> no automatic removal possible.

pkg/investigations/chgm/chgm.go
```diff
@@ -460,36 +411,12 @@ func extractUserDetails(cloudTrailEvent *string) (CloudTrailEventRaw, error) {
...
-func HandleResolved(r *investigation.Resources) error {
-	logging.Info("Remove 'Cluster has gone missing' limited support reason if any...")
-
-	removedReasons := false
-	err := utils.WithRetries(func() error {
-		var err error
-		removedReasons, err = r.OcmClient.DeleteLimitedSupportReasons(chgmLimitedSupport, r.Cluster.ID())
-		return err
-	})
+// postChgmSLAndSilence will put the cluster into limited support
```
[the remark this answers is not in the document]
typeid: Good catch! :)

pkg/investigations/chgm/chgm.go
```diff
@@ -460,36 +411,12 @@ func extractUserDetails(cloudTrailEvent *string) (CloudTrailEventRaw, error) {
...
-		return fmt.Errorf("failed to remove limited support")
-	}
-
-	if removedReasons {
-		metrics.Inc(metrics.LimitedSupportLifted, r.AlertType.String(), r.PdClient.GetEventType(), chgmLimitedSupport.Summary)
+		return fmt.Errorf("failed sending service log: %w", err)
 	}
 
-	return nil
+	return pdCli.SilenceAlertWithNote("Customer stopped instances. Sending SL and silencing alert.")
```
mjlshen, line 421:
```suggestion
	return pdCli.SilenceAlertWithNote("Customer stopped instances. Sent SL and silencing alert.")
```

nit, since the SL has already been sent successfully here

pkg/investigations/chgm/chgm.go
```diff
@@ -21,88 +21,77 @@ import (
...
+		logging.Error("Network verifier ran into an error: %s", err.Error())
+		notesSb.WriteString(fmt.Sprintf("⚠️ NetworkVerifier failed to run:\n\t %s", err))
 
-		logging.Info("Network verifier passed. Escalating to SRE")
-		err = r.PdClient.AddNote(fmt.Sprintln("Network verifier passed."))
+		err = r.PdClient.AddNote(notesSb.String())
 		if err != nil {
-			logging.Error("could not add passed message to incident notes")
+			// We do not return as we want the alert to be escalated either no matter what.
+			logging.Error("could not add failure reason incident notes")
```
mjlshen, line 79:
```suggestion
			logging.Error("could not add failure reason to incident notes")
```

mjlshen (commented):
Makes sense overall

pkg/investigations/ccam/ccam.go
typeid: I had checked this with a pd query, and over the span of 4 weeks this resulted in only 3 cases where we would maybe want to remove limited support manually.

Taking only incidents in the past month, everything that had nodes shutdown for longer than that has certs expired and should stay in LS: 
```
pd incident:list --services=prod-deadmanssnitch --since=11/1/2023 --notes | grep -c "Your cluster is no longer checking in with Red Hat OpenShift Cluster Manager. Possible causes include stopped instances or a networking misconfiguration. If you have stopped the cluster instances, please start them again - stopping instances is not supported. If you intended to terminate this cluster then please delete the cluster in the Red Hat console"
Getting incidents 2/2 👍, 0/2 👎... done
Getting notes for 34 incidents 34/34 👍, 0/34 👎... done

3
```

typeid:
/retest

lint timed out

PR #251, opened 2023-12-19
Introduce a new label that will keep CAD active 

rafael-azevedo:
Introduce a new label that will keep alerts active if the ClusterDeployment has a support exception label so that normal CAD workflow works.


rafael-azevedo:
since this is a hotfix and I wanted to modify the code as little as possible I had to duplicate the support exception label and checking code as to avoid having to create a mechanism to pass around this variable.

cadctl/cmd/investigate/investigate.go
```diff
@@ -222,12 +223,17 @@ func isAWS(cluster *cmv1.Cluster) (bool, error) {
 	return cloudProvider.ID() == "aws", nil
 }
 
+// ClusterDeploymentSupportExceptionLabel is the label indicating the cluster is under limited support
+// and the PagerDuty service should be enabled even if the cluster is in limited support
+// TODO:() Centrailize or Remove this varaiable once https://issues.redhat.com/browse/XCMSTRAT-427 is completed
+var ClusterDeploymentSupportExceptionLabel string = "hive.openshift.com/support-exception"
```
typeid, line 229: Why would we tackle this in CAD instead of the pagerduty operator layer (which is the component disabling services when a cluster is in LS)? If we want to make sure support exception clusters alert SRE, we should tackle it on a higher level than the alert receiving end (e.g. CAD).

Also, we are planning to revert the workaround that results in the DMS service not being disabled during limited support, meaning after that alerts for clusters that are in LS should not even make it to CAD.

rafael-azevedo: @typeid there is a customer support related issue that is related to an incident that occurred this is a BU temporary hot-fix until we are able to address it post holiday

typeid: Gotcha. Is there a specific reason we tackle this only in CAD though? Did the cluster trigger a CHGM that was skipped due to LS?
What's with all the other cluster related alerts that are not CHGM?

typeid: Ah, I see there's https://github.com/openshift/pagerduty-operator/pull/253. Nevermind the question then :)

pkg/investigations/chgm/chgm.go
```diff
@@ -30,6 +31,11 @@ var chgmLimitedSupport = ocm.LimitedSupportReason{
 	Details: "Your cluster is no longer checking in with Red Hat OpenShift Cluster Manager. Possible causes include stopped instances or a networking misconfiguration. If you have stopped the cluster instances, please start them again - stopping instances is not supported. If you intended to terminate this cluster then please delete the cluster in the Red Hat console",
 }
 
+// ClusterDeploymentSupportExceptionLabel is the label indicating the cluster is under limited support
+// and the PagerDuty service should be enabled even if the cluster is in limited support
+// TODO:() Centrailize or Remove this varaiable once https://issues.redhat.com/browse/XCMSTRAT-427 is completed
+var ClusterDeploymentSupportExceptionLabel = "hive.openshift.com/support-exception"
```
jharrington22, line 37:
```suggestion
var ClusterDeploymentSupportExceptionLabel = "ext-managed.openshift.io/support-exception"
```

@rafael-azevedo I updated the label

jharrington22 (changes requested):

cadctl/cmd/investigate/investigate.go
```diff
@@ -222,12 +223,17 @@ func isAWS(cluster *cmv1.Cluster) (bool, error) {
 	return cloudProvider.ID() == "aws", nil
 }
 
+// ClusterDeploymentSupportExceptionLabel is the label indicating the cluster is under limited support
+// and the PagerDuty service should be enabled even if the cluster is in limited support
+// TODO:() Centrailize or Remove this varaiable once https://issues.redhat.com/browse/XCMSTRAT-427 is completed
+var ClusterDeploymentSupportExceptionLabel = "hive.openshift.com/support-exception"
```
sam-nguyen7, line 229:
```suggestion
var ClusterDeploymentSupportExceptionLabel = "ext-managed.openshift.io/support-exception"
```

cadctl/cmd/investigate/investigate.go
```diff
@@ -239,7 +245,19 @@ func clusterRequiresInvestigation(cluster *cmv1.Cluster, ocmClient *ocm.SdkClien
...
+		if supportExValue, err := strconv.ParseBool(cd.Labels[ClusterDeploymentSupportExceptionLabel]); err == nil {
+			hasSupportException = supportExValue
+		}
+
+		ls, err := ocmClient.IsInLimitedSupport(cluster.ID())
+		if err != nil {
+			return false, err
+		}
+
+		if ls && !hasSupportException {
```
typeid, line 258: This change breaks the logic of escalation in case of a GCP alert, with this PR, CAD would try to investigate GCP clusters and run into errors instead of escalating them.

typeid (changes requested):
Commented some issues this PR would introduce.

The logic introduced into CAD with limited support alerts reaching it is making this overly complicated and very error prone.
I suggest we wait merging this and instead put focus on https://github.com/openshift/configuration-anomaly-detection/pull/246 which takes away all the complicated LS logic. With that PR and a revert on the handling of the DMS service by pagerduty-operator (= put service on hold when a cluster is in LS), we can handle supportex clusters completely transparently. The mentioned PR basically means CAD no longer needs to check for LS, as only non-LS alerts reach it, thus voiding this hotfix PR. 

I'm happy to follow up on this tomorrow morning and see if we can push it along.

PR #255, opened 2024-01-10
 OSD-20190: automatically send network SL for chgms

RaphaelBut:
Implements https://issues.redhat.com/browse/OSD-20190
The servicelog that got prepared before now gets sent automatically.
Additionally, we now silence when DMS ('nosnch.in') is unreachable  on a missing cluster as we know it fires because it cannot reach DMS.

If the servicelog fails to send it falls back to posting the template on the incident, so SRE can manually send it.

### DIFF
--- a/pkg/investigations/chgm/chgm.go
+++ b/pkg/investigations/chgm/chgm.go
@@ -29,6 +29,14 @@ var chgmSL = ocm.ServiceLog{
 	InternalOnly: false,
 }
 
+var egressSL = ocm.ServiceLog{
+	Severity:     "Error",
+	Summary:      "Action required: Network misconfiguration",
+	ServiceName:  "SREManualAction",
+	Description:  "Your cluster requires you to take action. SRE has observed that there have been changes made to the network configuration which impacts normal working of the cluster, including lack of network egress to these internet-based resources which are required for the cluster operation and support: %s. Please refer to the following firewall pre-requisites which are required for PrivateLink clusters: https://docs.openshift.com/rosa/rosa_install_access_delete_clusters/rosa_getting_started_iam/rosa-aws-prereqs.html#osd-aws-privatelink-firewall-prerequisites_prerequisites. Please revert changes.",
+	InternalOnly: false,
+}
+
 // InvestigateTriggered runs the investigation for a triggered chgm pagerduty event
 func InvestigateTriggered(r *investigation.Resources) error {
 	var notesSb strings.Builder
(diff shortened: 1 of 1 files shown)


pkg/investigations/chgm/chgm.go
```diff
@@ -77,8 +85,23 @@ func InvestigateTriggered(r *investigation.Resources) error {
...
+		}
+		metrics.Inc(metrics.ServicelogSent, r.AlertType.String(), r.PdClient.GetEventType())
+		notesSb.WriteString(fmt.Sprintf("⚠️ NetworkVerifier found unreachable targets. \n %s\n The servicelog has been sent.\n", failureReason))
+
+		if strings.Contains(failureReason, "nosnch.in") {
+			logging.Info("Alert is expected as the cluster is not able to reach deadmanssnitch. Silencing.")
+			notesSb.WriteString("Silencing the alert, as the cluster is not able to reach deadmanssnitch.")
+			return r.PdClient.SilenceAlertWithNote(notesSb.String())
+		}
+		notesSb.WriteString("Deadmanssnitch is not blocked. Please investigate the cluster.")
```
typeid, line 104: I think this might be confusing for the engineer receiving the alert with the note: the text will be really long and the `Deadmanssnitch is not blocked. Please investigate the cluster` will be somewhere after the long list of urls.

What do you think about these three options...
**option 1: add urls at the end** - do you think this would make it clear that it is unrelated to the alert? 
```
🤖 Automated CHGM pre-investigation 🤖
===========================
✅ Customer did not stop nodes.
✅ Deadmanssnitch URL is not blocked. 
🤖 NetworkVerifier sent a service log for unreachable targets: url1,url2,url3,url4
```
**option 2: leave out urls to not confuse SRE**
```
🤖 Automated CHGM pre-investigation 🤖
===========================
✅ Customer did not stop nodes.
✅ Deadmanssnitch URL is not blocked. 
```
**option 3: let SRE post the SL in case DMS isn't blocked**
This would be the current option: automated SLs only for DMS blocked. It would still allow the SRE to send or not send if they don't wish to.

typeid (commented):
One "UX" nitpick. Rest LGTM, let me know what you think.

pkg/investigations/chgm/chgm.go
RaphaelBut: Its a good point. I think I will go for option one, but also a note saying please investigate? 
Otherwise it may be confusing if you do a cluster context and see a network SL has just been sent for the cluster. wdyt?

RaphaelBut: testing here btw https://redhat.pagerduty.com/incidents/Q39EWN72NX2QJH/timeline

RaphaelBut:
```
🤖 Automated CHGM pre-investigation 🤖
===========================
✅ Customer did not stop nodes.
⚠️ NetworkVerifier found unreachable targets and sent the SL, but deadmanssnitch is not blocked! 
⚠️ Please investigate this cluster.
Unreachable: 
events.pagerduty.com:443
```

typeid (approved):
/lgtm

PR #259, opened 2024-03-07
update dashboard with new stage datasource

RaphaelBut:
We migrated to a new stage cluster and need to use its data source. 
When we migrate production cad, we will need to update the prod datasource as well.

Also small fix for two panels showing nodata as we no longer investigate resolved chgms.


typeid (approved):
/lgtm

PR #260, opened 2024-03-12
Add base image to dependabot configuration

RaphaelBut:
Recude toil and time to update the base image

I excluded the builder image due to the golang version being pinned.


typeid (approved):
/lgtm
/approve

PR #263, opened 2024-04-12
OSD-22320: Update SL severities to match OCM guidelines

typeid:
**What?**
CAD sends service logs with the "Error" severity, which does not exist. I updated the severities to match the OCM guidelines. 

**Why?**
https://issues.redhat.com/browse/OSD-22320


typeid:
Somehow a breaking change made of network verifier made it into CAD, but went under the radar. I'm not sure how the dep bump happened without triggering a unit test pipeline.

PR #272, opened 2024-05-06
Set resource limits to all containers

RaphaelBut:
Add resource requests and limits to all containers.
Infered some sensible values from production metrics. 

The TaskRuns only get a limit. This way the resource requests are set to the same value as the limit and we get QoS class Guaranteed. As we do not retry the TaskRuns, we should make them be evicted last. 

Closes: https://issues.redhat.com/browse/OSD-22524

### DIFF
--- a/deploy/pipeline-trigger.yaml
+++ b/deploy/pipeline-trigger.yaml
@@ -61,3 +61,16 @@ metadata:
 spec:
   triggers:
   - triggerRef: cad-pipe-listener
+  resources:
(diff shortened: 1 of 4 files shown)


RaphaelBut:
/hold 

Had weird golang and golangci issues.
Propose we do https://github.com/openshift/configuration-anomaly-detection/pull/271 first

typeid (approved):
/lgtm
/approve

PR #275, opened 2024-05-07
Fix invalid tekton step definition

RaphaelBut:
We use tekton v1beta1 and the field is 'resources' instead of 'computeResources'
https://tekton.dev/docs/pipelines/tasks/#defining-steps


typeid (approved):
/lgtm
/approve

PR #276, opened 2024-05-08
[OSD-22536] Add new routetables check for public subnets.

bergmannf:
This checks that a subnet does not have a route to `0.0.0.0/0` to an internet gateway. This usually denotes a public subnet.

Currently this keeps the old check for MapPublicIpOnLaunch as well if no IGW route is found.


typeid (approved):
/lgtm
/approve

PR #277, opened 2024-05-15
OSD-17681 Refactor entrypoint to facilitate adding new investigations

typeid:
This PR facilitates the entrypoint:
- we now handle every investigation as triggered incident
- logic to map incident to an investigation is now in a single function


cadctl/cmd/investigate/investigate.go
```diff
@@ -123,52 +122,27 @@ func run(_ *cobra.Command, _ []string) error {
...
-func isAlertSupported(alertTitle string) investigation.AlertType {
+// getInvestigation will return the investigation function for the identified alert
+func getInvestigation(alertTitle string) *investigation.Investigation {
 	// We currently map to the alert by using the title, we should use the name in the alert note in the future.
 	// This currently isn't feasible yet, as CPD's alertmanager doesn't allow for the field to exist.
 
 	// We can't switch case here as it's strings.Contains.
 	if strings.Contains(alertTitle, "has gone missing") {
-		return investigation.ClusterHasGoneMissing
+		return investigation.NewInvestigation(chgm.InvestigateTriggered, "ClusterHasGoneMissing")
```
RaphaelBut, line 141: polish: remove the "Triggered" from both functions, as we do not longer care about event type
chgm.Investigate
cpd.Investigate

RaphaelBut (commented):
Looks good! Nice cleanup.
We will need to check the grafana dashboard afterwards for any eventtype label issues.

README.md
```diff
@@ -28,30 +32,59 @@
...
+-  WebhookV3 
+-  EventOrchestrationWebhook
 
-## CAD CLI
+The required investigation is identified by CAD based on the incident and its payload. 
+As PagerDuty itself does not provide finer granularity for webhooks than service-based, CAD filters out the alerts it should investigate. For more information, please refer to https://support.pagerduty.com/docs/webhooks.
+
+To add a new alert investigation:
+- create a mapping for the alert to the `getInvestigation` function in `investigate.go` and write a corresponding CAD investigation (e.g. `Investigate()` in `chgm.go`).
+- if the alert is not yet routed to CAD, add a webhook to the service your alert fires on.
```
RaphaelBut, line 48: are we covered in terms of side effects when adding this webhook?

15 min auto escalate on cad failures for example

RaphaelBut: Quickly checked against our services, the services escalation policy needs to be updated as well to "Openshift Escalation Policy Automation (Stage)"

README.md
```diff
@@ -28,30 +32,59 @@
...
+### Running cadctl for an incident ID
+1) Export the required ENV variables, see [required ENV variables](#required-env-variables).
+2) Create a payload file containing the incident ID
+  ```bash
+  export INCIDENT_ID=
+  echo '{"event": {"data":{"id": "${INCIDENT_ID}"}}}' > ./payload
+  ```
+1) Run `cadctl` using the payload file
+  ```bash
+  ./cadctl/cadctl investigate --payload-path payload
```
RaphaelBut, line 65: bonus points if you make step 2 and 3 as a oneliner 
( ah typo as well in the enum ;) )

README.md
typeid: I updated it and kept it broad as this depends on what we want to do: for staging per example, we don't really care about the escalation policy when it comes to just running `cadctl`.

README.md
typeid: I kept the steps separate to make it clear what is being done. To re-test, you would only have to re-run step 3. Fixed the typo, thanks! :)

PR #279, opened 2024-05-17
PoC: Investigate ClusterMonitoringErrorBudgetBurn due to a misconfigured UWM configmap

typeid:
https://issues.redhat.com/browse/OSD-23312

This PR showcases how CAD could use kube api with an example implementation of the remediation to https://issues.redhat.com/browse/OCPBUGS-33863. 

The implementation is hidden behind a `CAD_EXPERIMENTAL_ENABLED` flag that we will enable on staging after enabling kube-api access in https://issues.redhat.com/browse/OSD-23311.

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -26,6 +26,7 @@ import (
 	investigation "github.com/openshift/configuration-anomaly-detection/pkg/investigations"
 	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/ccam"
 	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/chgm"
+	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/clustermonitoringerrorbudgetburn"
 	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/cpd"
 	"github.com/openshift/configuration-anomaly-detection/pkg/logging"
 	"github.com/openshift/configuration-anomaly-detection/pkg/managedcloud"
@@ -151,6 +152,9 @@ func getInvestigation(alertTitle string) *investigation.Investigation {
 	logging.Warn("Flag CAD_EXPERIMENTAL_ENABLED is set, experimental CAD investigations are enabled!")
 
 	// Experimental investigations go here
+	if strings.Contains(alertTitle, "ClusterMonitoringErrorBudgetBurnSRE") {
+		return investigation.NewInvestigation(clustermonitoringerrorbudgetburn.Investigate, "ClusterMonitoringErrorBudgetBurnSRE")
+	}
 
 	return nil
 }
--- a/pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
+++ b/pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
@@ -1,0 +1,79 @@
+// Package clustermonitoringerrorbudgetburn contains remediation for https://issues.redhat.com/browse/OCPBUGS-33863
+package clustermonitoringerrorbudgetburn
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	configv1 "github.com/openshift/api/config/v1"
+	investigation "github.com/openshift/configuration-anomaly-detection/pkg/investigations"
+	k8sclient "github.com/openshift/configuration-anomaly-detection/pkg/k8s"
+	"github.com/openshift/configuration-anomaly-detection/pkg/logging"
+	"github.com/openshift/configuration-anomaly-detection/pkg/notewriter"
+	"github.com/openshift/configuration-anomaly-detection/pkg/ocm"
+	"k8s.io/apimachinery/pkg/fields"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var uwmMisconfiguredSL = ocm.ServiceLog{
+	Severity:     "Major",
+	Summary:      "Action required: review user-workload-monitoring configuration",
+	ServiceName:  "SREManualAction",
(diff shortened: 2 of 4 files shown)


pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
```diff
@@ -0,0 +1,87 @@
...
+		r.OcmClient.PostServiceLog(r.Cluster.ID(), &uwmMisconfiguredSL)
+		return r.PdClient.SilenceAlertWithNote(notesSb.String())
+	}
+
+	notesSb.WriteString("✅ Monitoring CO not degraded due to a broken UWM configmap\n")
+	return r.PdClient.EscalateAlertWithNote(notesSb.String())
+}
+
+// Check if the `Available` status condition reports a broken UWM config
+func statusMatchesSymptom(monitoringCo *configv1.ClusterOperator) bool {
```
mrWinston, line 77: Major: Can you give this function a less generic name that makes it clear what it's checking? Something like `isUWMConfigInvalid` or so.

pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
```diff
@@ -0,0 +1,87 @@
...
+		Summary:      "Action required: review user-workload-monitoring configuration",
+		ServiceName:  "SREManualAction",
+		Description:  "Your cluster's user workload monitoring is misconfigured: the user-workload-monitoring-config ConfigMap in the openshift-user-workload-monitoring namespace is misconfigured. Please review your configuration alongside the product documentation: https://access.redhat.com/documentation/en-us/red_hat_openshift_service_on_aws/4/html/monitoring/configuring-the-monitoring-stack#.",
+		InternalOnly: false,
+	}
+)
+
+func Investigate(r *investigation.Resources) error {
+	ctx := context.Background()
+	scheme := runtime.NewScheme()
```
mrWinston, line 31: Minor: I'd find the `investigate` function much easier to parse if we moved the whole schema registration part into `k8sclient.New()`.

cadctl/cmd/investigate/investigate.go
```diff
@@ -142,6 +143,19 @@ func getInvestigation(alertTitle string) *investigation.Investigation {
 	} else if strings.Contains(alertTitle, "ClusterProvisioningDelay -") {
 		return investigation.NewInvestigation(cpd.Investigate, "ClusterProvisioningDelay")
 	}
+
+	// Return early if experimental features are not enabled
+	experimentalFeatures, experimentalExists := os.LookupEnv("CAD_EXPERIMENTAL_ENABLED")
+	if !experimentalExists || strings.ToUpper(experimentalFeatures) != "TRUE" {
```
mrWinston, line 149: Minor: Because we don't actually care about the distinction between `CAD_EXPERIMENTAL_ENABLED` being unset and empty, you just use `os.Getenv()` and simplify this statement to:
```
if strings.ToUpper(os.Getenv("CAD_EXPERIMENTAL_ENABLED")) != "TRUE" {
...
```
[os.Getenv()](https://pkg.go.dev/os#Getenv)

mrWinston (changes requested):

cadctl/cmd/investigate/investigate.go
typeid: Thanks!

pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
typeid: The k8sclient needs different schemas based on what we want to do with it, I'm not sure how much of an overhead it is to just add all schema initializations for all functions in that one place. Do you know what the standard practice is?

pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
typeid: Good point, thanks!

typeid:
I was able to test this already thanks to backplane-cli using my ocm-token.

typeid:
Waiting for [this backplane api ](https://github.com/openshift/backplane-cli/pull/421/files) change, as we need to be able to pass our ocm connection for the backplane kube-api auth

pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
mrWinston: yeah, i was wondering about that as well. Most of the times that i've seen this schema initializtion was during initialization of an operator (e.g. [aao](https://github.com/openshift/aws-account-operator/blob/master/main.go#L65)), where this is only called once. 
I wouldn't think that adding unneccessary schemes has a noticable performance impact, especially for this use case, but I don't have any proof to back that up :)

typeid: I'm happy with moving it, it definitely makes things cleaner. We can always move it back out later. 

Thanks for the feedback!

typeid:
/unhold

backplane-cli dependency PR merged. We're good to move ahead with this PR.

pkg/logging/logging.go
```diff
@@ -29,8 +29,8 @@ func InitLogger(logLevelString string, clusterID string) *zap.SugaredLogger {
 	config.EncoderConfig.TimeKey = "timestamp"
 	config.Level = logLevel
 	config.EncoderConfig.EncodeTime = zapcore.RFC3339TimeEncoder
-	config.EncoderConfig.StacktraceKey = "" // to hide stacktrace info
-	config.EncoderConfig.CallerKey = ""     // to hide caller info
+	config.EncoderConfig.StacktraceKey = ""   // to hide stacktrace info
+	config.EncoderConfig.CallerKey = "caller" // to hide caller info
```
typeid, line 33: This will allow us to remove the "double error check" from our makefile in an upcoming PR where I'm refactoring the Makefile.

typeid:
/unhold

Ready to merge! :)

zmird-r:
Looks good! /lgtm

PR #280, opened 2024-05-21
[PoC] Custom interceptor

typeid:
This is a PoC as part of RDA7: this PR adds a custom interceptor that allows routing all alerts to CAD without starting a new pipeline for each alert. To achieve this, we create a [custom interceptor](https://tekton.dev/docs/triggers/namespacedinterceptors/) in the form of a small http service that will determine whether or not we handle the alert or escalate it.


typeid:
TODOs:
- the interceptor must contain the logic for incident -> alert mapping
- deployment of the interceptor needs to be part of CAD deployment

README.md
```diff
@@ -61,7 +61,7 @@ Example usage:`./test/generate_incident.sh ClusterHasGoneMissing 2b94brrrrrrrrrr
 2) Create a payload file containing the incident ID
   ```bash
   export INCIDENT_ID=
-  echo '{"event": {"data":{"id": "${INCIDENT_ID}"}}}' > ./payload
```
typeid, line 64: This was the wrong format. The new format matches what we would get if we would add a default webhook to a service orchestration.

cadctl/cmd/investigate/investigate.go
```diff
@@ -131,20 +129,6 @@ func run(_ *cobra.Command, _ []string) error {
 	return alertInvestigation.Run(investigationResources)
 }
 
-// getInvestigation will return the investigation function for the identified alert
-func getInvestigation(alertTitle string) *investigation.Investigation {
```
typeid, line 135: Moved this to pkg/investigations/mapping: this is re-used by the interceptor so we don't have to add the investigation mapping in two places.

cadctl/cmd/investigate/investigate.go
```diff
@@ -161,24 +145,6 @@ func GetOCMClient() (*ocm.SdkClient, error) {
 	return ocm.New(cadOcmFilePath)
 }
 
-// GetPDClient will retrieve the PagerDuty from the 'pagerduty' package
-func GetPDClient(webhookPayload []byte) (*pagerduty.SdkClient, error) {
```
typeid, line 165: Moved to pkg/pagerduty, as we're re-using this in the interceptor.

cadctl/cmd/investigate/investigate.go
zmird-r: Is a new package needed? The function could live in the investigations package and the investigations package could be the one imported.

interceptor/test/e2e.sh
```diff
@@ -0,0 +1,66 @@
...
+
+echo "========= TESTS ============="
+# Test for a pre-existing alert we handle (ClusterProvisioningDelay)
+test_interceptor "Q12WO44XJLR3H3" "$EXPECTED_RESPONSE_CONTINUE"
+
+# Test for an alert we don't handle (alert called unhandled)
+test_interceptor "Q3722KGCG12ZWD" "$EXPECTED_RESPONSE_STOP"
+
+echo
+echo "Set the CAD_E2E_VERBOSE ENV variable for interceptor binary output."
```
zmird-r, line 63: Why hide the output as default? If the test fails, inspecting output would always be the next step, I can hardly see a case where you wouldn't want test output but if there is, it shouldn't be the default one.

zmird-r:
I added a couple of comments on minor things, apart from that it looks good.

cadctl/cmd/investigate/investigate.go
typeid: I think it makes sense to split them as we're splitting the usage from the implementation. I'd expect barely any modifications to the `investigation` package in the future, whereas the `mapping` package will constantly be modified. We might want different code reviewers for those in the future.

interceptor/test/e2e.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Load pd token from vault - needed by interceptor
+export VAULT_ADDR="https://vault.devshift.net"
+export VAULT_TOKEN="$(vault login -method=oidc -token-only)"
+for v in $(vault kv get  -format=json osd-sre/configuration-anomaly-detection/cad-testing | jq -r ".data.data|to_entries|map(\"\(.key)=\(.value|tostring)\")|.[]"); do export $v; done	
+unset VAULT_ADDR VAULT_TOKEN
+echo 
+
+# Run the interceptor, with output silenced if CAD_E2E_VERBOSE is not set
+if [[ -z "$CAD_E2E_VERBOSE" ]]; then
```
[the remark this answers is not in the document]
typeid: I would expect the test not to fail per default. While everything is green there is no need to have a very verbose output. Once tests fail, we want to look into the why. 

The output we're talking about is the interceptor binary output, so it's not relevant unless we have a failure.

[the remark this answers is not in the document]
typeid: That said it would be even better if the output is printed if tests fail automatically, I'll take a look.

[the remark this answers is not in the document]
typeid: Should be addressed now, thanks for the input! :)

cadctl/cmd/investigate/investigate.go
zmird-r: alright

typeid:
/retest 

Ran into a timeout, maybe I need to increase the timeout since we're now linting more things.

PR #282, opened 2024-05-27
Add kube-client

typeid:
Split up PR https://github.com/openshift/configuration-anomaly-detection/pull/279 into kube-client and the investigation


pkg/logging/logging.go
```diff
@@ -29,8 +29,8 @@ func InitLogger(logLevelString string, clusterID string) *zap.SugaredLogger {
 	config.EncoderConfig.TimeKey = "timestamp"
 	config.Level = logLevel
 	config.EncoderConfig.EncodeTime = zapcore.RFC3339TimeEncoder
-	config.EncoderConfig.StacktraceKey = "" // to hide stacktrace info
-	config.EncoderConfig.CallerKey = ""     // to hide caller info
+	config.EncoderConfig.StacktraceKey = ""   // to hide stacktrace info
+	config.EncoderConfig.CallerKey = "caller" // to hide caller info
```
mrWinston, line 33: Question: Is the comment still correct with this change?

mrWinston (commented):
Looks really good!

pkg/logging/logging.go
typeid: Good catch, that's an artifact of the past :)

PR #283, opened 2024-05-27
maint: add notewriter instead of building strings via sb directly

typeid:
This PR refactors to use a `NoteWriter` allowing to have a set format for notes we add to pagerduty.

### DIFF
--- a/pkg/notewriter/notewriter.go
+++ b/pkg/notewriter/notewriter.go
@@ -1,0 +1,66 @@
+package notewriter
+
+import (
+	"fmt"
+	"strings"
+
+	"go.uber.org/zap"
+)
+
+type NoteWriter struct {
+	investigationName string
+	sb                strings.Builder
+	logger            *zap.SugaredLogger
+}
+
+// New initializes a new NoteWriter with an optional logger.
+// The note is initialized with an investigation header in the following format:
+// 🤖 Automated %s pre-investigation 🤖
+// ===========================
+//
+// E.g.
+// 🤖 Automated CHGM pre-investigation 🤖
+// ===========================
+func New(investigationName string, logger *zap.SugaredLogger) *NoteWriter {
+	nw := &NoteWriter{investigationName, strings.Builder{}, logger}
+	nw.sb.WriteString(fmt.Sprintf("🤖 Automated %s pre-investigation 🤖\n", investigationName))
+	nw.sb.WriteString("===========================\n")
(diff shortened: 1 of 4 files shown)


pkg/notewriter/notewriter.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (n *NoteWriter) writeWithLog(format string, a ...any) {
+	if n.logger != nil {
+		n.logger.Infof(format, a...)
+	}
+
+	n.sb.WriteString(fmt.Sprintf(format, a...))
+}
+
+func (n *NoteWriter) AppendSuccess(format string, a ...any) {
```
RaphaelBut, line 35: polish: can we have comments for when to use which of the Append functions?

for append automation something like 
If cad is taking an action in your investigation please add it the the resulting pagerduty note with AppendAutomation

RaphaelBut: thought: you could also combine all three functions and let the indictor for success/failure/automation be some enum arg you pass 

it could make it easier to add new types later

pkg/notewriter/notewriter.go
```diff
@@ -0,0 +1,45 @@
...
+	nw.sb.WriteString(fmt.Sprintf("🤖 Automated %s pre-investigation 🤖\n", investigationName))
+	nw.sb.WriteString("===========================\n")
+	return nw
+}
+
+func (n *NoteWriter) String() string {
+	return n.sb.String()
+}
+
+func (n *NoteWriter) writeWithLog(format string, a ...any) {
```
RaphaelBut, line 27: praise: i love that we start logging it all that was a rather important step missing

RaphaelBut (commented):
looks good to me, comments are not blocking ;)

pkg/notewriter/notewriter.go
typeid: And on top of that, it's optional: no logger passed, no logging! :)

pkg/notewriter/notewriter.go
typeid: Thanks for pointing that out, lazy on my side! :) Added comments and format examples. 

For the thought you mentioned: I think the usage is easier as separate functions, as you get autocomplete. For structs this might be a bit less intuitive.

RaphaelBut: awesome !!

PR #289, opened 2024-06-21
feat: Escalate alerts for access protected clusters

RaphaelBut:
We are currently not running cad on access protected clusters. 
If the backplane error is due to access protection we escalate the alert with this information.
https://issues.redhat.com/browse/OSD-22153 
https://issues.redhat.com/browse/OSD-22155


pkg/investigations/ccam/ccam.go
```diff
@@ -22,6 +22,10 @@ var ccamLimitedSupport = ocm.LimitedSupportReason{
 func Evaluate(cluster *cmv1.Cluster, bpError error, ocmClient ocm.Client, pdClient pagerduty.Client, alertType string) error {
 	logging.Info("Investigating possible missing cloud credentials...")
 
+	if isAccessProtectionError(bpError.Error()) {
```
typeid, line 25: Hmm.. I'm not sure the check should be in this investigation. Do you think it would make more sense to move it to https://github.com/openshift/configuration-anomaly-detection/blob/24e690daf7ff8f891bf51d77028791c865976bde/cadctl/cmd/investigate/investigate.go#L200 next to the GCP check?

RaphaelBut: I am thinking the same yes. We can query ocm for the access protection status instead of going through a backplane failure.

typeid: Yup, let's do that! :)

typeid (approved):
/lgtm
/approve

PR #294, opened 2024-06-26
fix(OSD-24257): put clusters in LS if DMS/PD is blocked

typeid:
This PR updates CAD's CHGM behavior to put clusters that block PD/DMS into limited support. We need to do this as we can't guarantee SLA in those cases.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -3,7 +3,7 @@ run:
   modules-download-mode: readonly
   # Timeout for analysis, e.g. 30s, 5m.
   # Default: 1m
-  timeout: 5m
+  timeout: 15m
 
 issues:
   # on default, golangci-lint excludes gosec. Hence, we have to
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -123,7 +123,7 @@ func run(_ *cobra.Command, _ []string) error {
 		return ccam.Evaluate(cluster, err, ocmClient, pdClient, alertInvestigation.Name)
 	}
 
-	investigationResources := &investigation.Resources{InvestigationName: alertInvestigation.Name, Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: customerAwsClient, OcmClient: ocmClient, PdClient: pdClient}
+	investigationResources := &investigation.Resources{Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: customerAwsClient, OcmClient: ocmClient, PdClient: pdClient}
 
 	logging.Infof("Starting investigation for %s", alertInvestigation.Name)
 	return alertInvestigation.Run(investigationResources)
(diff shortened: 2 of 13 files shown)


.golangci.yml
```diff
@@ -3,7 +3,7 @@ run:
   modules-download-mode: readonly
   # Timeout for analysis, e.g. 30s, 5m.
   # Default: 1m
-  timeout: 5m
+  timeout: 15m
```
typeid, line 6: With the new binaries we have in this repo, we often time out with 5 min.

cadctl/cmd/investigate/investigate.go
```diff
@@ -123,7 +123,7 @@ func run(_ *cobra.Command, _ []string) error {
 		return ccam.Evaluate(cluster, err, ocmClient, pdClient, alertInvestigation.Name)
 	}
 
-	investigationResources := &investigation.Resources{InvestigationName: alertInvestigation.Name, Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: customerAwsClient, OcmClient: ocmClient, PdClient: pdClient}
+	investigationResources := &investigation.Resources{Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: customerAwsClient, OcmClient: ocmClient, PdClient: pdClient}
```
typeid, line 126: We didn't need the investigation name, it's better to have it as variable in the investigations to reduce bloat here.

pkg/investigations/chgm/chgm.go
```diff
@@ -22,21 +22,17 @@ import (
...
-	InternalOnly: false,
-}
+	chgmSL = ocm.ServiceLog{
+		Severity:     "Critical",
+		Summary:      "Action required: cluster not checking in",
+		ServiceName:  "SREManualAction",
+		Description:  "Your cluster is no longer checking in with Red Hat OpenShift Cluster Manager. Possible causes include stopped instances or a networking misconfiguration. If you have stopped the cluster instances, please start them again - stopping instances is not supported. If you intended to terminate this cluster then please delete the cluster in the Red Hat console",
+		InternalOnly: false,
+	}
+)
```
typeid, line 35: These were global variables we modify, which could lead to odd behavior. I created helper functions to create them.

pkg/investigations/chgm/chgm.go
```diff
@@ -110,6 +98,37 @@ func Investigate(r *investigation.Resources) error {
...
+			break
+		}
+	}
+
+	if setLS {
+		err := ocmCli.PostLimitedSupportReason(createEgressLS(egressFailures), cluster.ID())
+		if err != nil {
+			return err
+		}
+		metrics.Inc(metrics.LimitedSupportSet, investigationName, "EgressBlocked")
```
typeid, line 120: thought: we often run into issues with `metrics.Inc` due to the nature of it accepting everything, we should wrap the metric increases into functions to properly check parameters.

RaphaelBut: agreed

PR #296, opened 2024-06-26
Remove precommit hook

RaphaelBut:
we have improved the makefile 
Lets try our workflows without this hook, we are open to add it back though, if it turns out its needed.


typeid (approved):
/lgtm
/approve

PR #300, opened 2024-07-09
Don't use deprecated 'securityGroupId' in verifier library

AlexVulaj:
In the network verifier library, `securityGroupId` has long been deprecated in favor of `securityGroupIds`. This PR moves us over to use the latter so we aren't using a deprecated field.


RaphaelBut:
thanks @AlexVulaj ! ;)

/approve

PR #301, opened 2024-07-11
WIP: Add special handling for critical customer

RaphaelBut:
We do no longer post LimitedSupport if the target cluster is from a organization with the managed_critical_customer capability. Instead we redirect SRE to take additional steps for those clusters.

See https://issues.redhat.com/browse/OSD-24126

### DIFF
--- a/pkg/investigations/chgm/chgm.go
+++ b/pkg/investigations/chgm/chgm.go
@@ -85,7 +85,7 @@ func Investigate(r *investigation.Resources) error {
 			err := r.OcmClient.PostLimitedSupportReason(createEgressLS(failureReason), r.Cluster)
 			if err != nil {
 				notes.AppendWarning("NetworkVerifier found unreachable targets, deadmanssnitch is blocked! \nUnreachable: \n%s", failureReason)
-				notes.AppendWarning("Failed to post limited support: %v", err)
+				notes.AppendWarning("Failed to post limited support: %s", err.Error())
 				return r.PdClient.EscalateAlertWithNote(notes.String())
 			}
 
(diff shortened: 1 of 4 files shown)


pkg/ocm/ocm.go
```diff
@@ -224,6 +233,33 @@ func (c *SdkClient) PostLimitedSupportReason(limitedSupportReason *LimitedSuppor
 	return nil
 }
```
typeid, line 235: I wonder if we can make this cleaner. An error is probably not the ideal way to convey an expected failure.

RaphaelBut: Same :D Let me see...

pkg/investigations/chgm/chgm.go
```diff
@@ -82,9 +82,11 @@ func Investigate(r *investigation.Resources) error {
 		logging.Infof("Network verifier reported failure: %s", failureReason)
 
 		if strings.Contains(failureReason, "nosnch.in") {
-			err := r.OcmClient.PostLimitedSupportReason(createEgressLS(failureReason), r.Cluster.ID())
+			err := r.OcmClient.PostLimitedSupportReason(createEgressLS(failureReason), r.Cluster)
 			if err != nil {
-				return err
+				notes.AppendWarning("NetworkVerifier found unreachable targets, deadmanssnitch is blocked! \nUnreachable: \n%s", failureReason)
```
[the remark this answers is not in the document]
RaphaelBut: Nice catch. Did mindlessly copy this.
I think, it might be best to just remove the mention of DMS as its not relevant to the workflow here.
would look like this, wdyt?

🤖 Automated CHGM pre-investigation 🤖
===========================
✅ Customer did not stop nodes.
⚠️ NetworkVerifier found unreachable targets!
Unreachable: 
nosnch.in:443
⚠️Cannot post limited support on critical customers cluster. Please open a proactive case and silence the alert.

PR #304, opened 2024-08-15
maint: refactor mapping.go

typeid:
**What?** 
This PR refactors mapping.go to switch on strings.Contains(), making it easier to read and extend. 

**Why?**
Maintenance. 

**Notes to reviewer:**
I also looked at using a map, but it made it harder to read:
```
	// Map of alert title substrings to investigation constructors for standard investigations
	standardInvestigations := map[string]func() *investigation.Investigation{
		"has gone missing": func() *investigation.Investigation {
			return investigation.NewInvestigation(chgm.Investigate, "ClusterHasGoneMissing")
		},
		"ClusterProvisioningDelay -": func() *investigation.Investigation {
		},
	}
```


pkg/investigations/mapping/mapping.go
```diff
@@ -15,25 +15,26 @@ import (
...
-	logging.Warn("Flag CAD_EXPERIMENTAL_ENABLED is set, experimental CAD investigations are enabled!")
+	if experimentalFeaturesEnabled() {
+		logging.Warn("Flag CAD_EXPERIMENTAL_ENABLED is set, experimental CAD investigations are enabled!")
 
-	// Experimental investigations go here
-	if strings.Contains(alertTitle, "ClusterMonitoringErrorBudgetBurnSRE") {
-		return investigation.NewInvestigation(clustermonitoringerrorbudgetburn.Investigate, "ClusterMonitoringErrorBudgetBurnSRE")
+		// We don't care this is a single case switch (gocritic), this should just be extensible
+		switch { //nolint:gocritic
+		case strings.Contains(alertTitle, "ClusterMonitoringErrorBudgetBurnSRE"):
```
AlexVulaj, line 30: Instead of ignoring the linter here, can we add a "default" case that logs a message? This is probably helpful for the above statement as well.

AlexVulaj: Maybe something like...

```
default:
		logging.Infof("No investigation exists for %s", alertTitle)
```

typeid: I think that log makes sense, but rather than having it as default statement, it should be before the `return nil` to cover both `switch-cases`.  Then we still need the nolint statement, though. 

Let me know what you think about the latest push.

PR #305, opened 2024-08-21
Fix typo in log output when mapping is found in interceptor

AlexVulaj:
Self explanatory - this logger message is for when we find a mapping, but the text says the opposite.


typeid (approved):
/lgtm
/approve

PR #307, opened 2024-09-03
[OSD-22930] Migrate CAD to AWS SDK v2

bergmannf:
This PR migrates the codebase to the new AWS SDK version in preparation for easier creation of mocks.

Reference:

- https://aws.github.io/aws-sdk-go-v2/docs/migrating/

To better test the code we likely need to refactor a bit more to adhere to the recommendations.


pkg/aws/aws.go
```diff
@@ -122,16 +106,23 @@ func NewClientFromFileCredentials(dir string, region string) (*SdkClient, error)
 	}
 	accessKeyID := strings.TrimRight(string(accessKeyBytes), "\n")
 	secretKeyID := strings.TrimRight(string(secretKeyBytes), "\n")
+	// XXX: Does an empty session token work?
```
typeid, line 109: **chore:** any action to be taken for this comment? What does the `XXX` stand for?

pkg/aws/aws.go
```diff
@@ -145,77 +136,51 @@ func (c *SdkClient) AssumeRole(roleARN, region string) (*SdkClient, error) {
...
+				Name:   aws.String("instance-state-name"),
+				Values: []string{"running", "pending"},
 			},
 		},
+	})
+	if err != nil {
+		return nil, err
 	}
-	return c.listInstancesWithFilter(filters)
+	return output.Reservations[0].Instances, nil
```
typeid, line 155: **Question:** can this result in a panic? We're not sure there's reservations to my understanding. Previous code iterated through reservations and appended instances.

typeid (changes requested):
Thanks @bergmannf, left a few comments.

go.mod
```diff
@@ -5,6 +5,12 @@ go 1.21
 require (
 	github.com/PagerDuty/go-pagerduty v1.8.0
 	github.com/aws/aws-sdk-go v1.48.3
```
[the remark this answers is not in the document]
bergmannf: It was just used for the `.String()` that I replaced now - so it could finally be removed.

pkg/aws/aws.go
bergmannf: It seems that they only renamed what the parameter was called so I think there is no action: `XXX` just gets highlighted in my editor there are some functions to show me all occurrences in the project.

pkg/aws/aws.go
```diff
@@ -122,16 +106,23 @@ func NewClientFromFileCredentials(dir string, region string) (*SdkClient, error)
...
+	// XXX: Does an empty session token work?
 	return NewClient(accessKeyID, secretKeyID, "", region)
 }
 
+// FIXME: This seems quite useless
+// GetAWSCredentials gets the AWS credentials
+func (c *SdkClient) GetAWSCredentials() awsv2.Credentials {
+	return c.Credentials
+}
+
```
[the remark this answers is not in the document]
bergmannf: It's actually needed after all, because we only have the interface available - but to access the field it would need the struct, so the indirection is needed.

pkg/aws/aws.go
bergmannf: I was thinking of putting in paginator, but that would need more mocking on top - https://aws.github.io/aws-sdk-go-v2/docs/unit-testing/#mocking-paginators - so I think we can live without mocking those for now.

typeid (approved):
Thanks, looks great!

/lgtm
/approve

PR #308, opened 2024-09-04
Bumping network verivier version to v1.0.1

luis-falcon:
Bump network verifier version to v1.0.1


abyrne55 (changes requested):
As this is the first time CAD's using verifier v1, we'll need to review and probably revise all of CADs calls to `ValidateEgress()`

abyrne55:
We can close this as obsolete, as a newer version of verifier was released today
/close

PR #310, opened 2024-09-09
Replace nil check with len()

bergmannf:
As the slice is no longer a pointer, checking len() works in both cases:

- len(nil-slice) == 0
- len(empty-slice) == 0

See: https://go.dev/play/p/ah-hRMCmEWk


typeid (approved):
/lgtm
/approve

PR #311, opened 2024-09-23
fix(OSD-25690): pipeline failure for removed GetRole permission on installer

typeid:
**What?**

Handles the scenario where customers remove the `GetRole` permission of the installer role, which OCM requires to get the support role. With this PR, CAD will handle this scenario the same way as the other cases where customers broke permissions needed to access the cloud environment. 

**Why?**

This is a customer caused error that CAD can handle as "customer broke permissions" scenario route a notification to the customer.


Makdaam:
/lgtm
+ coverage of more failure modes

PR #316, opened 2024-10-10
OSD-25886 | Integrate osd-network-verifier v1.1.2

joshbranham:
This incorporates version `v1.1.2` of `osd-network-verifier`, which includes a breaking change to the `ValidateEgressInput` struct type. This additionally includes an update to fetching the machine image ID to no longer use the deprecated `GetAMIForRegion` function.

For some reason, I had to update the expectations of the `chgm` suite, although I didn't fundamentally change any AWS calls, things started failing. Would love if someone had an answer here.

Closes https://github.com/openshift/configuration-anomaly-detection/pull/315

### DIFF
--- a/pkg/networkverifier/networkverifier.go
+++ b/pkg/networkverifier/networkverifier.go
@@ -13,9 +13,10 @@ import (
 	"github.com/openshift/configuration-anomaly-detection/pkg/logging"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
-	"github.com/openshift/osd-network-verifier/pkg/helpers"
+	"github.com/openshift/osd-network-verifier/pkg/data/cloud"
+	"github.com/openshift/osd-network-verifier/pkg/data/cpu"
+	"github.com/openshift/osd-network-verifier/pkg/probes/curl"
 	"github.com/openshift/osd-network-verifier/pkg/proxy"
-	"github.com/openshift/osd-network-verifier/pkg/verifier"
 	onv "github.com/openshift/osd-network-verifier/pkg/verifier"
 	onvAwsClient "github.com/openshift/osd-network-verifier/pkg/verifier/aws"
 )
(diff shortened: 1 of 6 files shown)


pkg/networkverifier/networkverifier.go
```diff
@@ -56,13 +57,14 @@ func InitializeValidateEgressInput(cluster *v1.Cluster, clusterDeployment *hivev
 		"Name":                 "osd-network-verifier",
 	}
 
-	region := cluster.Region().ID()
-	if onvAwsClient.GetAMIForRegion(region) == "" {
-		return nil, fmt.Errorf("unsupported region: %s", region)
+	// We default to using the `curl` based probe for validating egress at this time
+	cloudImageID, err := curl.Probe{}.GetMachineImageID(cloud.AWSClassic, cpu.ArchX86, cluster.Region().ID())
+	if err != nil {
+		return nil, fmt.Errorf("failed getting machine image ID: %w", err)
```
abyrne55, line 63: No need to even set cloudImageID; the verifier will pick a sane default (using effectively the same logic you articulate here) when the `CloudImageID` field is left unset

pkg/networkverifier/networkverifier.go
```diff
@@ -13,9 +13,10 @@ import (
 	"github.com/openshift/configuration-anomaly-detection/pkg/logging"
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
-	"github.com/openshift/osd-network-verifier/pkg/helpers"
+	"github.com/openshift/osd-network-verifier/pkg/data/cloud"
+	"github.com/openshift/osd-network-verifier/pkg/data/cpu"
+	"github.com/openshift/osd-network-verifier/pkg/probes/curl"
 	"github.com/openshift/osd-network-verifier/pkg/proxy"
-	"github.com/openshift/osd-network-verifier/pkg/verifier"
 	onv "github.com/openshift/osd-network-verifier/pkg/verifier"
```
abyrne55, line 20: I appreciate you fixing this duplicate import, but perhaps we can keep the un-renamed version instead of the alias? Some people feel strongly about not using "ONV" 😅

joshbranham: Yeah good call, I agree, I tend to dislike package aliases :)

PR #318, opened 2024-10-24
[OSD-23696-Add-CAD-liveness-probe] Add liveness probe for interceptor

reedcort:
Adds liveness probe to CAD interceptor, so SRE can know when the interceptor service for CAD is down. This [PR](https://github.com/openshift/managed-cluster-config/pull/2252) will alert SRE when two or more CAD interceptor pods are crashloopoing.

https://issues.redhat.com/browse/OSD-23696


typeid (approved):
/lgtm
/approve

PR #319, opened 2024-11-04
Add loadbalancer service to the pipeline trigger

petrkotas:
This change adds a loadbalancer service to the pipeline trigger. Exposes the internal cluster to the pagerduty service. Only pagerduty IPs are allowed to access the pipeline trigger.

IPs might change, see:
- https://developer.pagerduty.com/docs/9a349b09b87b7-webhook-i-ps


typeid (approved):
/lgtm
/approve

PR #323, opened 2024-12-16
OSD-25983: making sure that the Makefile installs the tools needed by the various targets or at least display a meaningful error if some tool is not installed

Nikokolas3270:


rafael-azevedo (approved):
/lgtm

PR #324, opened 2024-12-19
Make dependabot work with multinode repo due to interceptor

RaphaelBut:
Our sub module interceptor requires to be updated ( might just be go mod tidy? ) when the outer mod receives updates.
See https://github.com/dependabot/dependabot-core/issues/6678


typeid (approved):
/lgtm
/approve

PR #326, opened 2025-01-07
Revert "Add loadbalancer service to the pipeline trigger"

typeid:
This reverts commit ed6f585f882ceb9539e96b5734f3d1cd63dd9f18.

The load balancer is created in app-interface separately from tekton. The reasoning for this:
- lack of tekton documentation on the topic
- we risk having the loadbalancer deleted by app-interface due to the annotation inheritance done by tekton, this is not an issue for most resources but for load balancers we map DNS entries to, it's a big deal
- we'll keep the deployment details apart from the code repo (although we have not done this in the past..)

### DIFF
--- a/deploy/pipeline-trigger.yaml
+++ b/deploy/pipeline-trigger.yaml
@@ -62,14 +62,11 @@ metadata:
   name: cad-event-listener
   annotations:
     triggers.tekton.dev/old-escape-quotes: "true"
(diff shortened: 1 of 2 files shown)


petrkotas (approved):
We verified that this does not work as expected, while also posing a risk of being auto deleted by tekton resource cleanup, as the tekton would inherit every object spawned from this.

PR #338, opened 2025-01-14
Align network misconfig LS message

RaphaelBut:
Solves failing to post LS on network misconfigs https://issues.redhat.com/browse/OSD-27300

The limitedsupport message defined here https://github.com/openshift/managed-notifications/blob/master/osd/limited_support/egressFailureLimitedSupport.json
does no longer include the blocked URLs, so we remove it here as well.  
Since networkverifier >= 1.0 the errors include parenthesis which break the current behavior ( as LS api does not allow for them, see https://issues.redhat.com/browse/OCM-13249 )

### DIFF
--- a/pkg/investigations/chgm/util.go
+++ b/pkg/investigations/chgm/util.go
@@ -20,8 +20,8 @@ func createEgressSL(blockedUrls string) *ocm.ServiceLog {
 	return &egressSL
 }
 
-func createEgressLS(blockedUrls string) *ocm.LimitedSupportReason {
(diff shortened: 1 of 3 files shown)


pkg/investigations/chgm/util.go
```diff
@@ -20,8 +20,8 @@ func createEgressSL(blockedUrls string) *ocm.ServiceLog {
 	return &egressSL
 }
 
-func createEgressLS(blockedUrls string) *ocm.LimitedSupportReason {
-	details := fmt.Sprintf("Your cluster requires you to take action. SRE has observed that there have been changes made to the network configuration which impacts normal working of the cluster, including lack of network egress to these internet-based resources which are required for the cluster operation and support: %s. Please revert changes, and refer to documentation regarding firewall requirements for PrivateLink clusters: https://access.redhat.com/documentation/en-us/red_hat_openshift_service_on_aws/4/html/prepare_your_environment/rosa-sts-aws-prereqs#osd-aws-privatelink-firewall-prerequisites_rosa-sts-aws-prereqs#", blockedUrls)
+func createEgressLS() *ocm.LimitedSupportReason {
```
typeid, line 23: **nit:** at this point we have a static LS reason and don't need to compute it anymore, so we can get rid of this function.

RaphaelBut: good catch, function removed

typeid (approved):
/lgtm
/approve

PR #341, opened 2025-01-15
Add metadata file to first investigation using backplane remediations

RaphaelBut:
See [SDE-3270](https://issues.redhat.com//browse/SDE-3270) for backplane remediations. 
We use the rbac in the metadata to create rbac on cluster and let cad use the kube-api for its investigation.

### DIFF
--- a/pkg/investigations/clustermonitoringerrorbudgetburn/metadata.yaml
+++ b/pkg/investigations/clustermonitoringerrorbudgetburn/metadata.yaml
@@ -1,0 +1,12 @@
+name: clustermonitoringerrorbudgetburn
+rbac:
+  roles: []
+  clusterRoleRules:
+    - verbs:
+        - "get"
+        - "list"
+      apiGroups:
+        - "config.openshift.io"
+      resources:
+        - clusteroperators
+customerDataAccess: false


pkg/investigations/clustermonitoringerrorbudgetburn/metadata.yaml
```diff
@@ -0,0 +1,12 @@
...
+  roles: []
+  clusterRoleRules:
+    - verbs:
+        - "get"
+        - "list"
+      apiGroups:
+        - "config.openshift.io"
+      resources:
+        - clusteroperators
+customerDataAccess: true
```
typeid, line 12: Question: does the true here have any effect? Also, would false not be more accurate here?

RaphaelBut: its used only for lockbox clusters to decide if we can go ahead

RaphaelBut: yeah false sounds better here, updating..

typeid (approved):
/lgtm

PR #342, opened 2025-01-16
[OSD-23494] Query all cloudtrail events for stop and terminate

bergmannf:
This commit uses the Paginator for CloudTrail events, instead of just running a single query.

This should prevent missing events, even though it might mean longer runtimes for CAD in case of very many events matching stop/terminate events.

It also will only retry the event-gathering instead of the all the actions that follow on the event, to keep the retry-loop smaller.

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -79,8 +79,7 @@ func NewClient(accessID, accessSecret, token, region string) (*SdkClient, error)
 		configv2.WithRegion(region),
 		configv2.WithCredentialsProvider(staticCredentials),
 		configv2.WithRetryer(func() awsv2.Retryer {
-			return retry.AddWithMaxBackoffDelay(retry.AddWithMaxAttempts(retry.NewStandard(), maxRetries),
-				time.Second*5)
+			return retry.AddWithMaxBackoffDelay(retry.AddWithMaxAttempts(retry.NewStandard(), maxRetries), time.Second*5)
 		}),
 	)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/aws/aws.go
```diff
@@ -229,88 +228,90 @@ func (c *SdkClient) PollInstanceStopEventsFor(instances []ec2v2types.Instance, r
 	idToCloudtrailEvent := make(map[string]cloudtrailv2types.Event)
 
 	var executionError error
+	var stoppedInstanceEvents []cloudtrailv2types.Event = make([]cloudtrailv2types.Event, 0)
+	var terminatedInstanceEvents []cloudtrailv2types.Event = make([]cloudtrailv2types.Event, 0)
 	err = wait.ExponentialBackoff(backoffOptions, func() (bool, error) {
```
typeid, line 233: **Question:** If a customer modifies their AWS permissions to give us a 403 when trying to list stopped instances  / terminated instance events, we would infinitely loop here. Is that correct? Shall we add a retry limit? 

This would already be the case currently, I believe we would then run into [the pipeline timeout](https://github.com/openshift/configuration-anomaly-detection/blob/1db358bf0f1efd3333d023f7f1a35c51cc3e5723/deploy/pipeline-trigger.yaml#L28).

pkg/aws/aws.go
```diff
@@ -491,20 +492,32 @@ func (c *SdkClient) defaultRouteTableForVpc(vpcId string) (*ec2v2types.RouteTabl
...
+	paginator := cloudtrailv2.NewLookupEventsPaginator(c.CloudtrailClient, in)
+	// FIXME: Decide if we should just always retrieve *all* events which could
+	// be wasteful
+	for paginator.HasMorePages() && len(events) < maxNumberEvents {
+		out, err := paginator.NextPage(context.TODO())
+		if err != nil {
+			return nil, err
+		}
+		for _, e := range out.Events {
+			logging.Infof("Event found: %s", *e.EventId)
```
typeid, line 516: This might become very noisy. Suggest making this debug.

bergmannf: Oh yep I only added those because it didn't work while testing until I realized RH-SRE events are always considered valid

pkg/aws/aws.go
typeid: You're correct, but then in the current logic we would never use the retry, as the lambda never returns `(false, nil)` correct?

Two easy possibilites:
- drop retries as a whole
- add a retry limit and update the logic to return `false, nil` when the AWS call fails, retry "blindly"

The harder but more complete solution:
- add a retry limit and update the logic to return `false, nil` when the AWS call fails + proper error handling to know when a retry makes sense

typeid (approved):
/lgtm
/approve

PR #347, opened 2025-02-03
[OSD-27978] Change investigation return type

bergmannf:
Change the return type of investigations, to indicate what actions were performed during the investigation.
Based on this information, metrics will now be set outside the investigation itself.


pkg/investigations/investigation.go
```diff
@@ -9,23 +9,36 @@ import (
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 )
 
+type InvestigationStep struct {
+	Performed bool
+	Labels    []string
+}
+
+type InvestigationResult struct {
+	InvestigationName  string
```
typeid, line 18: We should know the investigation name after calling `GetInvestigation` in `investigate.go`.

Would it be more logical to just pass that to `updateMetrics` instead of having it as pat of the `InvestigationResult`? That way, we can also remove the local variables for the investigation names in the investigation specific files, as those have no good reason to be there.

typeid (commented):
Great stuff @bergmannf! I added one question / possible suggestion.

cadctl/cmd/investigate/investigate.go
```diff
@@ -120,13 +120,18 @@ func run(_ *cobra.Command, _ []string) error {
 
 	customerAwsClient, err := managedcloud.CreateCustomerAWSClient(cluster, ocmClient)
 	if err != nil {
-		return ccam.Evaluate(cluster, err, ocmClient, pdClient, alertInvestigation.Name)
+		ccamResources := &investigation.Resources{Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: customerAwsClient, OcmClient: ocmClient, PdClient: pdClient, AdditionalResources: map[string]interface{}{"alertType": alertInvestigation.Name, "error": err}}
```
RaphaelBut, line 123: can we omit the alertType from the additionalResources here? Looks like its not used inside the ccam investigation

pkg/investigations/investigation.go
```diff
@@ -9,23 +9,35 @@ import (
...
-	ClusterDeployment *hivev1.ClusterDeployment
-	AwsClient         aws.Client
-	OcmClient         ocm.Client
-	PdClient          pagerduty.Client
+	Cluster             *cmv1.Cluster
+	ClusterDeployment   *hivev1.ClusterDeployment
+	AwsClient           aws.Client
+	OcmClient           ocm.Client
+	PdClient            pagerduty.Client
+	AdditionalResources map[string]interface{}
```
RaphaelBut, line 42: i think this can be generally nice to have ( this is added for the error to be passed to ccam? )

Are there other concerns with this besides remembering to do proper type assertion like you do above :D 
`bpError, ok := r.AdditionalResources["error"].(error)
if !ok {
	return result, fmt.Errorf("Missing required CCAM field 'error'")
}
`

bergmannf: No that's the main concern, but I don't like it much - we have a statically typed language and this just kind of opts-out of it, but it would allow 'pass whatever you want' - or I can just add a dedicated error for just CCAM and keep it `nil` in all other cases.

cadctl/cmd/investigate/investigate.go
bergmannf: Oh I must have missed that line when removing the usage.

typeid (approved):
/approve

@RaphaelBut lgtm from my side, feel free to add the label if this is ready for you as well.

pkg/investigations/investigation.go
RaphaelBut: I see, 
I think both is completely fine to go forward with for now :) 

i wouldn't worry too much as we want to change this further most likely, 
if we have an alert which we can handle without the awsclient/cloud, we might want ccam to only add the LS and still remediate the resize or whatever?

RaphaelBut: ah, this might get resolved when we properly chain the investigations 
the network verifier is called inside the chgm investigation for example instead of having its own investigation like ccam

but i feel like networkverifier and ccam are "basic checks" while chgm is alert -> basic check1, basic check2, nodes stopped? -> resolution/escalation

alert warrants investigation and always comes to a defined conclusion 
network verifier is informing but the chgm investigation decides what to do depending on network verifier output

these are non blocking comments, I am just thinking out loud haha^^

PR #349, opened 2025-02-06
Switch kube-api access to backplane remediations

RaphaelBut:
When creating k8s client, we now use backplane remediations and pass the remediation name

`k8scli, err := k8sclient.New(r.Cluster.ID(), r.OcmClient, "clustermonitoringerrorbudgetburn")`


go.mod
```diff
@@ -1,53 +1,61 @@
 module github.com/openshift/configuration-anomaly-detection
 
-go 1.21
+go 1.22.7
```
typeid, line 3: CI still runs 1.21 AFAIK, we can't just bump this.

pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
```diff
@@ -27,7 +27,7 @@ var uwmMisconfiguredSL = ocm.ServiceLog{
 func Investigate(r *investigation.Resources) (investigation.InvestigationResult, error) {
 	// Initialize k8s client
 	result := investigation.InvestigationResult{}
-	k8scli, err := k8sclient.New(r.Cluster.ID(), r.OcmClient)
+	k8scli, err := k8sclient.New(r.Cluster.ID(), r.OcmClient, "clustermonitoringerrorbudgetburn")
```
typeid, line 30: **thought:** would be nice to have this in a less error prone way than typing it out.

pkg/investigations/mapping/mapping.go
```diff
@@ -20,6 +21,8 @@ func GetInvestigation(alertTitle string) *investigation.Investigation {
 		return investigation.NewInvestigation(chgm.Investigate, "ClusterHasGoneMissing")
 	case strings.Contains(alertTitle, "ClusterProvisioningDelay -"):
 		return investigation.NewInvestigation(cpd.Investigate, "ClusterProvisioningDelay")
+	case strings.Contains(alertTitle, "InsightsOperatorDownSRE"):
+		return investigation.NewInvestigation(insightsoperatordown.Investigate, "InsightsOperatorDown")
```
typeid, line 25: Accidentally commited this? I don't see this existing in the PR.

test/set_stage_env.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
typeid, line 1: Nice!

pkg/investigations/mapping/mapping.go
RaphaelBut: ah jup ^^ coming in another PR

pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
RaphaelBut: yeah, definitely, will change this to use the investigation resources

there is a slight annoyance that backplane expects the directory to be the same name as the investigation name here
will need to document this

pkg/investigations/investigation.go
```diff
@@ -34,6 +34,7 @@ func NewInvestigation(investigationFn func(resources *Resources) (InvestigationR
 
 // Resources holds all resources/tools required for alert investigations
 type Resources struct {
+	Name                string
```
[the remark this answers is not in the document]
RaphaelBut: If you have no strong feeling about it, I would opt for not doing it at this point. 

We could sort that out with the way backplane finds the metadata file as well..

typeid (commented):
/lgtm
/approve

PR #354, opened 2025-02-11
Add intesigation interface

petrkotas:
**What this PR does:**

Adds the interface for investigations. Adds auto registrations for all investigations.

**Why:**

It simplifies the process of adding new investigations.


RaphaelBut:
hyped for this 
/honk

pkg/investigations/ccam/ccam.go
```diff
@@ -62,6 +64,22 @@ func Investigate(r *investigation.Resources) (investigation.InvestigationResult,
...
+func (c *CCAM) Name() string {
+	return "Cluster Credentials Are Missing (CCAM)"
+}
+
+func (c *CCAM) Description() string {
+	return "Detects missing cluster credentials"
+}
+
+func (c *CCAM) CheckActive(alert string) bool {
+	return false
```
RaphaelBut, line 76: is this special to ccam because it does not relate to an alert and thus always false?

pkg/investigations/registry.go
```diff
@@ -0,0 +1,27 @@
...
+	availableInvestigations = append(availableInvestigations, inv)
+}
+
+// GetInvestigation returns the first Investigation that applies to the given alert title.
+// This is a naive version that only returns the first matching investigation and ignores the rest.
+// Future improvement is to use the proper mapping that can return multiple investigations
+// linked to single alert type.
+func GetInvestigation(title string, experimental bool) Investigation {
+	for _, inv := range availableInvestigations {
+		if inv.CheckActive(title) && (experimental || !inv.IsExperimental()) {
```
RaphaelBut, line 22: non-blocking comment
naming things is hard but I will also give it a try :D 
inv.MatchesTitle(title) instead of CheckActive?

typeid (commented):
Personal opinion: I think it makes it makes it slightly more complex to add a new investigation. I'd like a few more people to weigh in on that. 

A few things that add additional cognitive load IMO:
- respecting the interface: there is no clear boundaries that enforce respecting the interface when creating a new investigation
- abstraction of the mapping: when writing a new investigation, it's not clear how an alert will map to an investigation, as we abstract it away. I think we could slightly improve this by renaming `CheckActive()`

What I like about this PR is that we're encapsulating things further into the respective `investigations` subfolders.

pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
```diff
@@ -69,6 +71,22 @@ func Investigate(r *investigation.Resources) (investigation.InvestigationResult,
 	return result, r.PdClient.EscalateIncidentWithNote(notes.String())
 }
 
+func (c *CMEBB) Name() string {
+	return "ClusterMonitoringErrorBudgetBurn"
+}
+
+func (c *CMEBB) Description() string {
```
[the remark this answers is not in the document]
petrkotas: Currently none, in the following implementations to get description for additional interfaces like webui.

petrkotas:
I did the changes to make the code little more understandable. I agree that naming is hard. Should we postpone the naming to the final phase of refactoring?

joshbranham:
I agree building out a simple interface to provide a contract to make new investigations is a good path forward. This will ensure folks have patterns to follow and implementation is as scoped to the business logic in question as possible.

> respecting the interface: there is no clear boundaries that enforce respecting the interface when creating a new investigation

@typeid with regard to the above comment, my answer to that would be testing. Each investigation should have unit tests on it's own in some form, and we could also maintain a factory to return all investigations and ensure they implement the interface.

interceptor/pkg/interceptor/pdinterceptor.go
```diff
@@ -106,7 +106,7 @@ func (pdi *PagerDutyInterceptor) Process(ctx context.Context, r *triggersv1.Inte
 		return interceptors.Failf(codes.InvalidArgument, "could not initialize pagerduty client: %v", err)
 	}
 
-	investigation := investigation_mapping.GetInvestigation(pdClient.GetTitle())
+	investigation := investigations.GetInvestigation(pdClient.GetTitle(), true)
```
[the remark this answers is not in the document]
typeid: @petrkotas this is still an issue here.

pkg/investigations/investigation/investigation.go
```diff
@@ -20,16 +20,12 @@ type InvestigationResult struct {
...
-type Investigation struct {
-	Run  func(resources *Resources) (InvestigationResult, error)
-	Name string
-}
-
-// NewInvestigation creates a new investigation
-func NewInvestigation(investigationFn func(resources *Resources) (InvestigationResult, error), name string) *Investigation {
-	return &Investigation{investigationFn, name}
+type Investigation interface {
+	Investigate(resources *Resources) (InvestigationResult, error)
```
typeid, line 24: Do you think calling this function `Run` is "cleaner" than `Investigate`? This reduces repetitiveness e.g. if you have 
`investigation.Investigate` vs `investigation.Run`

pkg/investigations/registry.go
typeid: `HandlesAlert`?

typeid (commented):
I'm happy with the new registration pattern from the latest push, it's similar to what we had before but encapsulates things better. :) 

If we address open comments & this is tested, it's is good to go for me. Thanks @petrkotas

typeid (approved):
/hold - to have a synced state with @RaphaelBut's presentation. 
/lgtm
/approve

PR #355, opened 2025-02-11
Cleanup k8s

RaphaelBut:
Add the cleanup call to backplane-api to delete investigation RBAC

After the investigation has ended we need to cleanup the created RBAC.
We will have garbage collection via managed-cluster-config in case this
fails. One can also delete it manually with `ocm backplane remediation delete <clusterid> <investigation>` .


typeid (approved):
/lgtm
/approve

PR #358, opened 2025-02-14
Improve generate_incident script

RaphaelBut:
Adds short delay to allow pagerduty to create the incident

It now creates the payload file for you


typeid (approved):
/lgtm
/approve

PR #359, opened 2025-02-14
Fix clustermonitoringerrorbudgetburn

RaphaelBut:


RaphaelBut:
we will have to not tie this to the investigation name ^^

PR #360, opened 2025-02-18
OSD-28248: Investigation bootstrap make target

MateSaary:
Adds bootstrap script (& make target) to create boilerplate code for new investigations, and appends it to investigations interface in `registry.go`


hack/bootstrap-investigation.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
```
typeid, line 1: Can we add the metadata file as part of this as well? Even if it's empty (no permissions).

hack/bootstrap-investigation.sh
```diff
@@ -0,0 +1,85 @@
...
+	notes := notewriter.New(r.Name, logging.RawLogger)
+
+	// TODO:
+	// Implement investigation logic here
+
+	return result, r.PdClient.EscalateIncidentWithNote(notes.String())
+}
+
+func (c *${STRUCT_NAME}) Name() string {
+	return "TODO" // TODO: Add name
```
typeid, line 59: Can this be retrieved as part of the bootstrap?

hack/bootstrap-investigation.sh
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (c *${STRUCT_NAME}) Description() string {
+	return "TODO" // TODO: Add description
+}
+
+func (c *${STRUCT_NAME}) ShouldInvestigateAlert(alert string) bool {
+	return strings.Contains(alert, "TODO") // TODO: Add alert string
```
typeid, line 67: Can this be retrieved as part of the bootstrap?

hack/bootstrap-investigation.sh
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (c *${STRUCT_NAME}) IsExperimental() bool {
+	return true // TODO: Modify experimental flag status
```
typeid, line 71: Can this be retrieved as part of the bootstrap?

hack/bootstrap-investigation.sh
MateSaary: Added following, let me know if I understood the question correctly;
```
touch "${INVESTIGATION_DIR}/metadata.yaml"
```

MateSaary:
@typeid thank you for the review!

I have added aforementioned aspects as a series of "please enter..." prompts, with the exception of `IsExperimental()`; should this condition not be true by default? I may be misunderstanding it's usage.

Is there anywhere else the above info can be sourced without manual prompting?

Concerning https://github.com/openshift/configuration-anomaly-detection/pull/362#discussion_r1967675640, I also added generation of a readme file as part of the script, let me know if this looks good or if there is anything else to add to this :)

hack/bootstrap-investigation.sh
```diff
@@ -0,0 +1,110 @@
...
+type ${STRUCT_NAME} struct{}
+
+func (c *${STRUCT_NAME}) Run(r *investigation.Resources) (investigation.InvestigationResult, error) {
+	result := investigation.InvestigationResult{}
+
+	// Initialize PagerDuty note writer
+	notes := notewriter.New(r.Name, logging.RawLogger)
+
+	// TODO:
+	// Implement investigation logic here
```
typeid, line 77:
```suggestion
	// TODO: Implement investigation logic here
```

typeid (commented):
Looking great, added some more minor suggestions. 

You're right on the experimental, we should just default to false and not even bother asking when a new investigation is bootstrapped.

MateSaary:
Thank you for the suggestions, all changed now.

I'm wondering if something along the lines of "Unique Alert Title/Identifier" would be already a bit more informative, otherwise I left the prompt as is for the time being. Thanks!

typeid:
@MateSaary sorry for the delay on this. Are you okay to incorporate changes from https://github.com/openshift/configuration-anomaly-detection/pull/369 into this PR?

MateSaary:
@typeid no worries! Should be changed accordingly now.

hack/bootstrap-investigation.sh
```diff
@@ -0,0 +1,119 @@
...
+
+read -p "Enter new investigation description: " INVESTIGATION_DESCRIPTION
+if [[ "${INVESTIGATION_DESCRIPTION}" == "" ]] ; then
+	INVESTIGATION_DESCRIPTION="TODO"
+fi
+
+read -p "Should Investigate Alert (y/n): " INVESTIGATE_ALERT_BOOL
+if [[ "${INVESTIGATE_ALERT_BOOL}" == "y" ]] ; then
+	read -p "Investigation alert string: " INVESTIGATION_ALERT_STRING
+	INVESTIGATION_ALERT="strings.Contains(alert, ${INVESTIGATION_ALERT_STRING})"
```
typeid, line 22:
```suggestion
	INVESTIGATION_ALERT="strings.Contains(alert, \"${INVESTIGATION_ALERT_STRING}\")"
```

hack/bootstrap-investigation.sh
```diff
@@ -0,0 +1,119 @@
...
+
+touch "${INVESTIGATION_DIR}/${INVESTIGATION_NAME}.go"
+touch "${INVESTIGATION_DIR}/metadata.yaml"
+touch "${INVESTIGATION_DIR}/README.md"
+
+# Create README.md file
+cat <<EOF > "${INVESTIGATION_DIR}/README.md"
+# ${INVESTIGATION_NAME} Investigation
+
+*TODO*
```
typeid, line 50:
```suggestion
${INVESTIGATION_DESCRIPTION}
```

PR #362, opened 2025-02-24
investigation metadata

RaphaelBut:
When adding investigations, i think it would be good to have some way of codifying the lifecycle

I will keep this here for discussion/placeholder until I can do more work on it

### DIFF
--- a/pkg/investigations/clustermonitoringerrorbudgetburn/metadata.yaml
+++ b/pkg/investigations/clustermonitoringerrorbudgetburn/metadata.yaml
@@ -10,3 +10,7 @@ rbac:
       resources:
         - clusteroperators
 customerDataAccess: false
+sop: https://github.com/openshift/ops-sop/blob/master/v4/troubleshoot/clusteroperators/monitoring.md
+issues:
+  - issue: https://issues.redhat.com/browse/OCPBUGS-33863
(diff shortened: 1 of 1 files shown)


pkg/investigations/clustermonitoringerrorbudgetburn/metadata.yaml
```diff
@@ -10,3 +10,7 @@ rbac:
       resources:
         - clusteroperators
 customerDataAccess: false
+sop: https://github.com/openshift/ops-sop/blob/master/v4/troubleshoot/clusteroperators/monitoring.md
+issues:
+  - issue: https://issues.redhat.com/browse/OCPBUGS-33863
+    removeAfter: 4.17
```
typeid, line 16: IMO this should be in a README of the investigation package. 

@MateSaary maybe we could generate a README as part of the boilerplate setup as well: https://github.com/openshift/configuration-anomaly-detection/pull/361 - thoughts?

RaphaelBut: i like the idea of having a readme for each investigation  and we start of from there
 once we implement some kind of lifecycle automation ( PR checks tracking the bugs) we can revisit this

PR #368, opened 2025-03-11
Add request unwrapping to interceptor

MateSaary:


interceptor/pkg/interceptor/pdinterceptor.go
```diff
@@ -90,15 +90,36 @@ func (pdi *PagerDutyInterceptor) executeInterceptor(r *http.Request) ([]byte, er
 	}
 	r.Body = io.NopCloser(bytes.NewReader(body.Bytes()))
 
+	var wrapper struct {
+		Body   string              `json:"body"`
+		Header map[string][]string `json:"header"`
+	}
+	if err := json.Unmarshal(body.Bytes(), &wrapper); err != nil {
+		return nil, badRequest(fmt.Errorf("failed to parse body as JSON: %w", err))
```
RaphaelBut, line 98: polish: failed to parse request body?

interceptor/pkg/interceptor/pdinterceptor.go
```diff
@@ -90,15 +90,36 @@ func (pdi *PagerDutyInterceptor) executeInterceptor(r *http.Request) ([]byte, er
 	}
 	r.Body = io.NopCloser(bytes.NewReader(body.Bytes()))
 
+	var wrapper struct {
```
RaphaelBut, line 93: naming things is hard :D I think this is okay, but can we highlight that this is the "original" request reaching the evenlistener from pagerduty we are unwrapping

or a comment on what its for, would achieve the same :)

interceptor/pkg/interceptor/pdinterceptor.go
```diff
@@ -90,15 +90,36 @@ func (pdi *PagerDutyInterceptor) executeInterceptor(r *http.Request) ([]byte, er
...
+		Body   string              `json:"body"`
+		Header map[string][]string `json:"header"`
+	}
+	if err := json.Unmarshal(body.Bytes(), &wrapper); err != nil {
+		return nil, badRequest(fmt.Errorf("failed to parse body as JSON: %w", err))
+	}
+
+	extractedRequest, err := http.NewRequestWithContext(ctx, r.Method, r.URL.String(), bytes.NewReader([]byte(wrapper.Body)))
+	if err != nil {
+		return nil, internal(fmt.Errorf("failed to create new request: %w", err))
```
RaphaelBut, line 103: thought: while this error message is accurate, i can imagine it might be confusing as there is no request happening, maybe something along the lines of, malformed request body/header ( in unwrapped request ) ?

PR #369, opened 2025-03-11
OSD-28725 - Rename investigation structures to standard name 'Investigation'

tnierman:
https://issues.redhat.com/browse/OSD-28725

---

This PR just updates the name of each investigation package's primary struct to a standard name, `Investigation`.

The goal is to create a consistent package standard, and to make object initialization more idiomatic. ie -

```
&clustermonitoringerrorbudgetburn.CMEBB{}
```
becomes
```
&clustermonitoringerrorbudgetburn.Investigation{}
```


typeid (approved):
/lgtm
/approve

PR #372, opened 2025-03-14
Hotfix dev env

RaphaelBut:
Improve dev environment setup and add utility to run backplane locally

Only require aws client for investigations that need it. 
This comes out of a need to run backplane-api locally, but its unable to get aws access.
Its also part of the bigger picture of only initializing required integrations.

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+	"github.com/openshift/configuration-anomaly-detection/pkg/aws"
 	investigations "github.com/openshift/configuration-anomaly-detection/pkg/investigations"
 	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/ccam"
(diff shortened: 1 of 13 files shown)


cadctl/cmd/investigate/investigate.go
```diff
@@ -129,18 +130,23 @@ func run(cmd *cobra.Command, _ []string) error {
...
 
-	customerAwsClient, err := managedcloud.CreateCustomerAWSClient(cluster, ocmClient)
-	if err != nil {
-		ccamResources := &investigation.Resources{Name: "ccam", Cluster: cluster, ClusterDeployment: clusterDeployment, AwsClient: customerAwsClient, OcmClient: ocmClient, PdClient: pdClient, AdditionalResources: map[string]interface{}{"error": err}}
-		inv := ccam.Investigation{}
-		result, err := inv.Run(ccamResources)
-		updateMetrics(alertInvestigation.Name(), &result)
-		return err
+	var customerAwsClient *aws.SdkClient
+	if alertInvestigation.RequiresAwsClient() {
```
typeid, line 134: **Question**: how would we test aws access + kube-api access short term? 
This is an interesting side effect of running the local backplane-api.

PR #384, opened 2025-03-21
Simplify local testing steps

RaphaelBut:
Remove the need to export pd token, by using cads own token. 
Redo the corresponding Readme steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,6 @@ To add a new alert investigation:
 > **Note:** When writing an investiation, you can use them right away.
 They are initialized for you and passed to the investigation via investigation.Resources.
 
-
 * [AWS](https://github.com/aws/aws-sdk-go) -- Logging into the cluster, retreiving instance info and AWS CloudTrail events.
     - See `pkg/aws`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,24 +89,28 @@ They are initialized for you and passed to the investigation via investigation.R
...
+   ```bash
+   ./test/generate_incident.sh <alertname> <clusterid>
+   ```
+
+2. Export the required env variables from vault
+
+   > **Note:** For information on the envs see [required env variables](#required-env-variables).
+
+   ```
+   source test/set_stage_env.sh
```
typeid, line 105: Doesn't this already set the env var, if we move this step before the generate incident, does it fix the issue without needing to touch the generation script?

RaphaelBut: mhh yeah i see what you mean, 

in generate_incident we use the  creds from `osd-sre/configuration-anomaly-detection/cad-testing`
in set_stage_env its stage pd creds  `osd-sre/configuration-anomaly-detection/pd/stg`

it results in kind of the same, so we could change the order 
but I feel like this is good for now, as the incident is usually created only once until you have a new cluster/alert title

setting stage envs is required more often, e.g each time you reopen your terminal/editor

typeid (approved):
/lgtm
/approve

PR #386, opened 2025-03-24
first draft of metadata for upgradeconfigsyncfailureover4hr

Mhodesty:
Adding metadata first so I can do local testing


pkg/investigations/upgradeconfigsyncfailureover4hr/metadata.yaml
```diff
@@ -0,0 +1,15 @@
...
+      rules:
+        - verbs:
+            - "get"
+            - "list"
+          apiGroups:
+            - ""
+          resources:
+            - "secrets"
+          resourceNames:
+            - "pull-secret"
```
joshbranham, line 14: Is this the only secret you need, by name?

pkg/investigations/upgradeconfigsyncfailureover4hr/metadata.yaml
```diff
@@ -0,0 +1,15 @@
+name: upgradeconfigsyncfailureover4hr
+rbac:
+  roles:
+    - namespace: "openshift-config"
+      rules:
+        - verbs:
+            - "get"
+            - "list"
```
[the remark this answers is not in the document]
Mhodesty: Sure, I'll remove the list.

pkg/investigations/upgradeconfigsyncfailureover4hr/metadata.yaml
Mhodesty: Yeah, that's what it looks like in the SOP. If I need to get more permissions I can add them later but I'd rather have to add another specific permission than have to remove extraneous permissions.

joshbranham (commented):
Thanks for scoping this down, much better place to be, and we can always add later if we need to

/lgtm

PR #387, opened 2025-03-24
Added metadata.yaml for CannotRetrieveUpdatesSRE for investigations

anispate:
[OSD-18645](https://issues.redhat.com/browse/OSD-18645) - For local testing

### DIFF
--- a/pkg/investigations/CannotRetrieveUpdatesSRE/metadata.yaml
+++ b/pkg/investigations/CannotRetrieveUpdatesSRE/metadata.yaml
@@ -1,0 +1,12 @@
+name: cannotretrieveupdatessre
+rbac:
+  roles: []
+  clusterRoleRules:
+    - verbs:
+        - "get"
+        - "list"
+      apiGroups:
+        - "config.openshift.io"
+      resources:
+        - clusteroperators
+customerDataAccess: false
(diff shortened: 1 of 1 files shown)


pkg/investigations/CannotRetrieveUpdatesSRE/metadata.yaml
```diff
@@ -0,0 +1,20 @@
...
+      resources:
+        - clusteroperators
+    - apiGroups:
+        - ""
+      resources:
+        - pods
+        - namespaces
+      verbs:
+        - get
+        - list
```
joshbranham, line 19: Does this actually need a ClusterRole and `get/list` on `pods/namespaces` for the _entire_ cluster? This should be scoped as tightly as we can right?

anispate: I had added it as per ref Alice talked about last week but ill perform a trial and error and if during the testing time, i will need it then ill add the get/list on pods/namespaces then

pkg/investigations/CannotRetrieveUpdatesSRE/metadata.yaml
```diff
@@ -0,0 +1,12 @@
...
+  roles: []
+  clusterRoleRules:
+    - verbs:
+        - "get"
+        - "list"
+      apiGroups:
+        - "config.openshift.io"
+      resources:
+        - clusterversions
+customerDataAccess: false
```
tnierman, line 12: Should this be `true`?

tnierman: Nevermind, this relates to lockbox to prevent investigations against objects/data that a customer would consider sensitive

PR #389, opened 2025-03-25
Fix interceptor e2e env issue & investigation name logging

MateSaary:
This PR aims to fix an issue with the interceptor e2e not adding the env signature variable, thus unreadable to the interceptor. It also corrects the investigation's name in logging.

### DIFF
--- a/interceptor/pkg/interceptor/pdinterceptor.go
+++ b/interceptor/pkg/interceptor/pdinterceptor.go
@@ -110,8 +110,6 @@ func (pdi *PagerDutyInterceptor) executeInterceptor(r *http.Request) ([]byte, er
 
 	var ireq triggersv1.InterceptorRequest
 
-	// logging request
-	logging.Debug("Unwrapped Request header: %v", extractedRequest.Header)
 	logging.Debug("Unwrapped Request body: ", originalReq.Body)
 
 	token, _ := os.LookupEnv("PD_SIGNATURE")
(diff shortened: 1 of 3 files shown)


interceptor/test/e2e.sh
```diff
@@ -82,7 +82,3 @@ test_interceptor "Q12WO44XJLR3H3" "$EXPECTED_RESPONSE_CONTINUE"
 echo "Test 2: unhandled alerts returns a 'continue: false' response"
 test_interceptor "Q3722KGCG12ZWD" "$EXPECTED_RESPONSE_STOP"
 
-# Test for an alert with invalid signature
```
typeid, line 85: This test is great, let's keep it but fix it. 
An idea would be to add a new parameter to `test_interceptor` that allows overriding the signature. This would be optional, we would only pass it for this test.

interceptor/pkg/interceptor/pdinterceptor.go
```diff
@@ -159,7 +157,7 @@ func (pdi *PagerDutyInterceptor) Process(ctx context.Context, r *triggersv1.Inte
 		}
 	}
 
-	logging.Infof("Incident %s is mapped to investigation '%s', returning InterceptorResponse `Continue: true`.", pdClient.GetIncidentID(), investigation.Name)
+	logging.Infof("Incident %s is mapped to investigation '%s', returning InterceptorResponse `Continue: true`.", pdClient.GetIncidentID(), investigation.Name())
```
typeid, line 160: There's some more `investigation.Name` occurences in the repo, could you fix those as well? :)

typeid (changes requested):

interceptor/pkg/interceptor/pdinterceptor.go
MateSaary: I found one other reference (cadtcl/cmd/investigate/investigate.go).

interceptor/test/e2e.sh
MateSaary: Thanks for the suggestion! Let me know your thoughts on the fix.

typeid (approved):
/lgtm
/approve

PR #395, opened 2025-03-27
OSD-28525 - Initial implementation for MachineHealthCheckUnterminatedShortCircuitSRE investigation

tnierman:
https://issues.redhat.com/browse/OSD-28525

---

These changes introduce the initial implementation for the `machineHealthCheckUnterminatedShortCircuitSRE` alert. The alert investigation takes the following steps:
- Retrieves the machines managed by any failling machinehealthcheck object. Machines & nodes that are failing which are otherwise un-managed by a machinehealthcheck that is short-circuited (ie - masters) are not investigated
    - Machines that are deleting will have their node checked to see if it's stuck draining. At the moment, this check is imperfect: once the ability to query metrics is introduced, we can more precisely confirm what workloads are blocking a node drain based on the `pods_preventing_node_drain` metric
    - Machines that are in a failed state will have their .Status.ErrorReason and .Status.ErrorMessages examined
    - For each of the machines that's examined and found to be problematic, a recommended action will be added to the investigation's `recommendations` map. This map is used to holistically evaluate the state of the cluster following the investigation
- Once all machines have been evaluated, the investigation moves on to the machines' nodes.
    - Nodes whose machines have already been identified as problematic are **not** evaluated: this cuts down on noise and prevents us from accidentally recommending two different courses of action for the same underlying compute instance
    - Currently, the node investigation is a bit bare, it primarily determines whether a specific node warrants further human investigation. Future iterations involving metrics may be able to expand on this aspect of the investigation
- Finally, once all objects have been evaluated, the investigation's `recommendations` are summarized. By compiling the summary at the end of the investigation, rather than during each iteration, a single action can be recommended for a number of different nodes & machines (ie - "send this one servicelog regarding machine A, B, and C" rather than "send servicelog regarding machine A" + "send servicelog regarding machine B" + "send servicelog regarding machine C", etc)

Additionally, recommended test methods are supplied in the `testing/` directory, along with a (somewhat) detailed README to help newcomers get started. This includes test objects that should be directly applicable to any OSD/ROSA staging cluster.

### DIFF
--- a/pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
+++ b/pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
@@ -1,0 +1,329 @@
+/*
+machinehealthcheckunterminatedshortcircuitsre defines the investigation logic for the MachineHealthCheckUnterminatedShortCircuitSRE alert
+*/
+package machinehealthcheckunterminatedshortcircuitsre
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+	"time"
+
+	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+
+	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/investigation"
+	machineutil "github.com/openshift/configuration-anomaly-detection/pkg/investigations/utils/machine"
+	nodeutil "github.com/openshift/configuration-anomaly-detection/pkg/investigations/utils/node"
+	k8sclient "github.com/openshift/configuration-anomaly-detection/pkg/k8s"
+	"github.com/openshift/configuration-anomaly-detection/pkg/logging"
+	"github.com/openshift/configuration-anomaly-detection/pkg/notewriter"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	alertname = "MachineHealthCheckUnterminatedShortCircuitSRE"
+	// remediationName must match the name of this investigation's directory, so it can be looked up via the backplane-api
+	remediationName = "machineHealthCheckUnterminatedShortCircuitSRE"
+)
+
+type Investigation struct {
+	// kclient provides access to on-cluster resources
+	kclient client.Client
+	// notes holds the messages that will be shared with Primary upon completion
+	notes *notewriter.NoteWriter
+	// recommendations holds the set of actions CAD recommends primary to take
+	recommendations investigationRecommendations
+}
+
+func (i *Investigation) setup(r *investigation.Resources) error {
+	// Setup investigation
+	k, err := k8sclient.New(r.Cluster.ID(), r.OcmClient, remediationName)
+	if err != nil {
+		return fmt.Errorf("failed to initialize kubernetes client: %w", err)
+	}
+	i.kclient = k
+	i.notes = notewriter.New(r.Name, logging.RawLogger)
+	i.recommendations = investigationRecommendations{}
+
+	return nil
+}
+
+// Run investigates the MachineHealthCheckUnterminatedShortCircuitSRE alert
+//
+// The investigation seeks to provide exactly one recommended action per affected machine/node pair.
+// The machine object is evaluated first, as it represents a lower-level object that could affect the health of the node.
+// If the investigation determines that the breakage is occurring at the machine-level, the corresponding node is *not* investigated.
+// After investigating all affected machines, potentially affected nodes are investigated.
+func (i *Investigation) Run(r *investigation.Resources) (investigation.InvestigationResult, error) {
+	ctx := context.Background()
+	result := investigation.InvestigationResult{}
+
+	// Setup & teardown
+	err := i.setup(r)
+	if err != nil {
+		return result, fmt.Errorf("failed to setup investigation: %w", err)
+	}
+	defer func(r *investigation.Resources) {
+		err := k8sclient.Cleanup(r.Cluster.ID(), r.OcmClient, remediationName)
+		if err != nil {
+			logging.Errorf("failed to cleanup investigation: %w", err)
+		}
+	}(r)
+
+	targetMachines, err := i.getMachinesFromFailingMHC(ctx)
+	if err != nil {
+		i.notes.AppendWarning("failed to retrieve one or more target machines: %v", err)
+	}
+	if len(targetMachines) == 0 {
+		i.notes.AppendWarning("no machines found for short-circuited machinehealthcheck objects")
+		return result, r.PdClient.EscalateIncidentWithNote(i.notes.String())
+	}
+
+	problemMachines, err := i.InvestigateMachines(ctx, targetMachines)
+	if err != nil {
+		i.notes.AppendWarning("failed to investigate machines: %v", err)
+	}
+
+	// Trim out the machines that we've already investigated and know have problems
+	//
+	// The purpose of this is to avoid re-investigating nodes whose machines were already investigated. Any node-level issue on a failing machine
+	// is most likely related to the machine itself, and providing duplicate/conflicting advice will only prove confusing to the responder
+	targetMachines = slices.DeleteFunc(targetMachines, func(targetMachine machinev1beta1.Machine) bool {
+		for _, problemMachine := range problemMachines {
+			if problemMachine.Name == targetMachine.Name {
+				return true
+			}
+		}
+		return false
+	})
+
+	// If one or more machines managed by the machinehealthcheck have not yet been identified as a problem, check on the machine's
+	// node to determine if there are node-level problems that need remediating
+	if len(targetMachines) > 0 {
+		targetNodes, err := machineutil.GetNodesForMachines(ctx, i.kclient, targetMachines)
+		if err != nil {
+			i.notes.AppendWarning("failed to retrieve one or more target nodes: %v", err)
+		}
+		if len(targetNodes) > 0 {
+			i.InvestigateNodes(targetNodes)
+		}
+	}
+
+	// Summarize recommendations from investigation in PD notes, if any found
+	if len(i.recommendations) > 0 {
+		i.notes.AppendWarning(i.recommendations.summarize())
+	} else {
+		i.notes.AppendSuccess("no recommended actions to take against cluster")
+	}
+
+	return result, r.PdClient.EscalateIncidentWithNote(i.notes.String())
+}
+
+func (i *Investigation) getMachinesFromFailingMHC(ctx context.Context) ([]machinev1beta1.Machine, error) {
+	healthchecks := machinev1beta1.MachineHealthCheckList{}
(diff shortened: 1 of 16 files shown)


pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
```diff
@@ -0,0 +1,487 @@
...
+		return result, fmt.Errorf("failed to setup investigation: %w", err)
+	}
+	defer func(r *investigation.Resources) {
+		err := i.cleanup(r)
+		if err != nil {
+			logging.Errorf("failed to cleanup investigation: %w", err)
+		}
+	}(r)
+
+	targetMachines, err := i.getTargetMachines(ctx)
```
bergmannf, line 86: Mostly my personal opinion, but I generally prefer to retrieve all external resources up-front, so we fail if there are issues getting `machines` or `nodes` and don't run half an investigation.
In this case this might even use less resources, if we just retrieve all machines we want in one call + all nodes we want in on-call as well, instead of one-by-one?

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
```diff
@@ -0,0 +1,487 @@
...
+	}
+	return role, nil
+}
+
+// getNodesForMachines retrieves the nodes for the given machines. Errors encountered are joined, but do not block the retrieval of other machines
+func (i *Investigation) getNodesForMachines(ctx context.Context, machines []machinev1beta1.Machine) ([]corev1.Node, error) {
+	nodes := []corev1.Node{}
+	var errs error
+	for _, machine := range machines {
+		node, err := i.getNodeForMachine(ctx, machine)
```
bergmannf, line 335: If we just run 2 kubernetes API calls (even though it might retrieve 'too many nodes' first) this might be faster still instead of running the loop that gets nodes one-by-one maybe?
Just get all machines we care about - then get all nodes for all of those in a single call and filter on the result.

bergmannf:
I think this looks really good already - I guess it could check more things, but I think the PR is already quite big and it's better to get this in and running in stage before adding even more checks.
(Something that happens and could be found by this is customers running out of available CIDRs for new nodes: machine = Running, Node never becoming ready, which would need to inspect the `network` configuration)

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
```diff
@@ -0,0 +1,487 @@
...
+	i.kclient = k
+	i.notes = notewriter.New(r.Name, logging.RawLogger)
+	i.recommendations = recommendations{}
+
+	return nil
+}
+
+func (i *Investigation) cleanup(r *investigation.Resources) error {
+	return k8sclient.Cleanup(r.Cluster.ID(), r.OcmClient, remediationName)
+}
```
typeid, line 62: What's the value in this function?

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
```diff
@@ -0,0 +1,487 @@
...
+	nodeRoleWorkerSuffix = "worker"
+)
+
+type Investigation struct {
+	// kclient provides access to on-cluster resources
+	kclient client.Client
+	// notes holds the messages that will be shared with Primary upon completion
+	notes *notewriter.NoteWriter
+	// recommendations holds the set of actions CAD recommends primary to take
+	recommendations recommendations
```
typeid, line 44: nit: confusing naming as it is also a type

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
typeid, line 86: **Suggestion:** find a good name for `getTargetMachines`. The name is not self-explanatory IMO. What is a target machine? I would have to read through the function to findout that we're looking for machines where a remediation is allowed by MHC.

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
```diff
@@ -0,0 +1,487 @@
...
+	case machinev1beta1.DeleteMachineError:
+		notes := fmt.Sprintf("the machine's node could not be gracefully terminated automatically: %q", errorMsg)
+		i.recommendations.addRecommendation(recommendationInvestigateMachine, machine.Name, notes)
+	case machinev1beta1.InsufficientResourcesMachineError:
+		notes := fmt.Sprintf("a servicelog should be sent because there is insufficient quota to provision the machine: %q", errorMsg)
+		i.recommendations.addRecommendation(recommendationQuotaServiceLog, machine.Name, notes)
+	default:
+		notes := "no .Status.ErrorReason found for machine"
+		i.recommendations.addRecommendation(recommendationInvestigateMachine, machine.Name, notes)
+	}
```
typeid, line 258: **Suggestion:** Instead of the big switch here, you could centralize the mapping of ErrorReason to recommendation logic, e.g.:
```
var machineFailureHandlers = map[machinev1beta1.MachineStatusError]func(machinev1beta1.Machine) (recommendedAction, string){
	machinev1beta1.IPAddressInvalidReason: func(m machinev1beta1.Machine) (recommendedAction, string) {
		return recommendationDeleteMachine, fmt.Sprintf("invalid IP: %q", *m.Status.ErrorMessage)
	},
	// etc...
}
```

Then you could replace the long switch case:
```
handler, ok := machineFailureHandlers[errorReason]
if !ok {
	i.notes.AppendInfo("Unknown error reason %s on failed machine %s", errorReason, machine.Name)
	return nil
}

action, note := handler(machine)
i.recommendations.addRecommendation(action, machine.Name, note)
return nil
```

This would decouple things and should be easier on the eyes and decouple things:)

typeid (commented):
Added some suggestions/nit comments.

Additional suggestions/thoughts (non-blocking):
I generally find it a bit odd we're going through all machines and looking for MHC available remediation. Would it be more logical to look at the MHCs and triaging by everything that has `unhealthyCount > maxUnhealthy` and to focus on those? 

As the file grew quite large, we could also consider splitting up the large file here into something like the following:
- `machine_investigation.go`: `InvestigateMachines`, `investigateFailingMachine`, `investigateDeletingMachine`, `getMachineRole`
- `node_investigation.go`: `InvestigateNode`, `InvestigateNodes`, `findNodeReadyCondition`, `checkForStuckDrain`
- `utils.go`: `findNoScheduleTaint`, `getNodeRole`
- `recommendations.go`: `recommendations`, `investigationResult`, and `String()` logic

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
tnierman: Good call, that would save us (quite a few) API calls

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
tnierman: Honestly, I was anticipating more cleanup post-install, but that didn't turn out to be necessary

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
tnierman: renamed `getTargetMachines` -> `getMachinesFromFailingMHC`

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
tnierman: Fair enough, renamed `recommendations` type -> `investigationRecommendations`

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
tnierman: Removed extraneous function

tnierman:
@typeid - thanks for the recommendations!

Specifically regarding 
> I generally find it a bit odd we're going through all machines and looking for MHC available remediation. Would it be more logical to look at the MHCs and triaging by everything that has unhealthyCount > maxUnhealthy and to focus on those?

It sounds like you're suggesting we remediate only the machines owned by a failing machineHealthCheck, correct? 

That _was_ the goal of `getMachinesFromFailingMHC()` (previously `getTargetMachines()`), though upon re-examining that logic, I had [mismatched the return value to the name](https://github.com/openshift/configuration-anomaly-detection/pull/395/files#diff-68fb4ef790ff1538774f70862609dc8f06dd59e26b5b400081bf8462105ddf7eR158), and then misused the [value to still get the right set of machines](https://github.com/openshift/configuration-anomaly-detection/pull/395/files#diff-68fb4ef790ff1538774f70862609dc8f06dd59e26b5b400081bf8462105ddf7eR140) (realistically, I probably changed the name of that function, and forgot to change the behavior :grimacing:). I pushed a fix for that now, but let me know if your concern extends beyond that error

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
tnierman: After thinking on it for a bit, I personally prefer the big-switch: it's easier to tell at-a-glance what recommendation we're making for each machine failure-state, and it feels like less complexity for the same end-result.

If this is something that you or others feel strongly about, I'm happy to make the switch, however.

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
tnierman: Updated this logic in the latest commit to retrieve all nodes first, then filter out the unneeded ones before returning

pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
tnierman: I added newlines between the `case` statements in an effort to improve readability

typeid (approved):
/lgtm
/approve

PR #397, opened 2025-03-31
OSD-28224 - cad should always post failures to the incident notes

MateSaary:
This PR aims to add handling to ensure that investigation failures (and notes collected up until failure) are always posted to PagerDuty, as currently some exceptions/errors result in empty notes.

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -76,6 +76,14 @@ func run(cmd *cobra.Command, _ []string) error {
 
 	logging.Infof("Incident link: %s", pdClient.GetIncidentRef())
 
+	var investigationResources *investigation.Resources
+
+	defer func() {
+		if err != nil {
+			handleCADFailure(err, investigationResources)
+		}
+	}()
+
 	_, cadExperimentalEnabled := os.LookupEnv("CAD_EXPERIMENTAL_ENABLED")
 	alertInvestigation := investigations.GetInvestigation(pdClient.GetTitle(), cadExperimentalEnabled)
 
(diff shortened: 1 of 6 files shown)


cadctl/cmd/investigate/investigate.go
```diff
@@ -121,29 +133,49 @@ func run(cmd *cobra.Command, _ []string) error {
...
+
+	return investigationErr
+}
+
+func handleCADFailure(err error, resources *investigation.Resources) {
+	logging.Errorf("CAD investigation failed: %v", err)
+
+	notes := resources.Notes
+
+	notes.AppendWarning("CAD investigation failed, CAD team has been notified. Please investigate manually.")
```
RaphaelBut, line 172: thought: Do we want to add the error to the notes as well? 
On the one hand it could be confusing to see some error which is possibly completely unrelated to the alert at hand, but I can imagine it being a clue in some cases

wdyt?

cadctl/cmd/investigate/investigate.go
```diff
@@ -76,6 +76,15 @@ func run(cmd *cobra.Command, _ []string) error {
 
 	logging.Infof("Incident link: %s", pdClient.GetIncidentRef())
 
+	var investigationErr error
```
RaphaelBut, line 79: Question: Why do we need to remap err to investigationErr?

I suspected the defer function needs to throw an extra error which would overwrite the original one, along the lines of how its used here https://github.com/openshift/configuration-anomaly-detection/blob/main/pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go#L42

Just realized, the other defer statement has a bug. [This line](https://github.com/openshift/configuration-anomaly-detection/blob/2285ffcc45d2312939815cac39b4efeaa46fc861/pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go#L46) will not do anything, as we are missing named return parameters in the surrounding function haha. https://go.dev/ref/spec#Defer_statements  

It will log the defer error fine, but the joining wont work. We need to switch the surrounding function to have (err error). Then the defer function will be able to access it.

Not sure if we should do the same thing here, meaning joining the "original" error and the pdError from handleCADFailure

MateSaary: Great points, thanks for the comment.

I remapped err to investigationErr to differentiate between the investigation errors as opposed to others, though on second thought this doesn't seem necessary (I also think it might have served more purpose in a previous version of my code), so I've reverted the errors back. 

I've also named the return parameters for the function in`clustermonitoringerrorbudgetburn.go`, so the join apply properly now!

I'm not certain on whether we should do the same thing here, not sure if I'm clear on what instance would this be beneficial? :smile:

cadctl/cmd/investigate/investigate.go
```diff
@@ -121,29 +133,49 @@ func run(cmd *cobra.Command, _ []string) error {
...
+
+func handleCADFailure(err error, resources *investigation.Resources) {
+	logging.Errorf("CAD investigation failed: %v", err)
+
+	notes := resources.Notes
+
+	notes.AppendWarning("CAD investigation failed, CAD team has been notified. Please investigate manually.")
+	pdErr := resources.PdClient.EscalateIncidentWithNote(notes.String())
+	if pdErr != nil {
+		logging.Errorf("Failed to escalate notes to PagerDuty")
```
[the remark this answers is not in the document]
MateSaary: Good spot, thanks!

cadctl/cmd/investigate/investigate.go
MateSaary: I changed the error message slightly to surround it with "🚨" to match other CAD-related info/error messages (i.e. "🤖 No automation implemented for this alert; escalated to SRE. 🤖"), I think this might be a decent compromise between the two! It might make it a bit clearer that this error message is not related to the notes of the investigation. Let me know what you think!

cadctl/cmd/investigate/investigate.go
RaphaelBut: Nice, thanks for fixing the other defer! 
Yeah, I think here its better to just let the original error bubble up. 
The goal is to still add the investigation notes in case of error. So when we look at a failed pipeline it makes more sense to see the error that caused the failure instead of a possible failure with adding the notes after. ( which we log as well so that's fine! )

PR #398, opened 2025-04-01
OSD: Fixed pipeline validation errors

rpodishe:
This PR fixes the validation errors mentioned in the ticket OSD-27590.


deploy/pipeline-trigger.yaml
```diff
@@ -32,6 +32,17 @@ spec:
         params:
           - name: payload
             value: $(tt.params.payload)
+        podTemplate:
```
typeid, line 35: Could this be done via `ComputeResources`? https://tekton.dev/docs/pipelines/pipelineruns/#specifying-task-level-computeresources

RaphaelBut: I am a bit confused by the tekton concepts here. We have specified request and limits on the Task Level ( which defines the pod? ) https://github.com/openshift/configuration-anomaly-detection/blob/main/openshift/template.yaml#L324
```
name: step-check-infrastructure
    resources:
      limits:
        cpu: 100m
        memory: 256Mi
      requests:
        cpu: 10m
        memory: 64Mi
 
```
Aha, got it. Tekton also adds init containers to the pod and those are missing it.
```
initContainers:
  - command:
    - /tekton/bin/entrypoint
    - step-check-infrastructure
    name: prepare
    resources: {}
```

deploy/pipeline-trigger.yaml
```diff
@@ -32,6 +32,17 @@ spec:
...
+          spec:
+            containers:
+              - name: step-container
```
RaphaelBut, line 38: If the above comment makes sense, then this should target the inticontainer, meaning "prepare" 

i am not sure what the init container needs in terms of resources. ( i am also not sure what it does haha )
But I suspect it should need less than the step container?

deploy/pipeline-trigger.yaml
rpodishe: > Could this be done via `ComputeResources`? https://tekton.dev/docs/pipelines/pipelineruns/#specifying-task-level-computeresources

@typeid yes this can be via ComputeResources as well. I used podTemplate is because the issue is related to the pod but not the task. pod can contain any number of containers/init containers, this configuration applies to all containers created as part of the pipeline run.
image

deploy/pipeline-trigger.yaml
rpodishe: @RaphaelBut updated the code.

typeid (approved):
/lgtm
/approve

PR #400, opened 2025-04-03
fix stage env script after removing aws creds from vault

RaphaelBut:
Followup from https://github.com/openshift/configuration-anomaly-detection/pull/394


typeid (approved):
woopsie, thanks @RaphaelBut 


/lgtm
/approve

PR #403, opened 2025-04-04
remove broken dependabot config, see OSD-28972

RaphaelBut:


typeid:
@RaphaelBut OSD-28972 is meant to help here, the implementation that landed on the main branch is a workflow which lints the interceptor in case it's needed. To my understanding we can keep the current configuration for dependabot as is. Keep me honest on this.

PR #404, opened 2025-04-04
OSD-18645 -  Initial implementation for CannotRetrieveUpdatesSRE

anispate:
[OSD-18645](https://issues.redhat.com/browse/OSD-18645) - CAD implementation for CannotRetrieveUpdatesSRE

Sample ticket: https://redhat.pagerduty.com/incidents/Q1S45W54TK1QKU#:~:text=%E2%9A%A0%EF%B8%8F%20ClusterVersion%20error%20detected,primary%20for%20review 

Updated sample ticket: https://redhat.pagerduty.com/incidents/Q2UVRI8YGLPP3G

Updated sample ticket: https://redhat.pagerduty.com/incidents/Q01H4ZQHM90EML


test/generate_incident.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/zsh
```
joshbranham, line 1: This should be
```suggestion
#!/usr/bin/env bash
```

pkg/investigations/registry.go
```diff
@@ -1,6 +1,7 @@
 package investigations
 
 import (
+	CannotRetrieveUpdatesSRE "github.com/openshift/configuration-anomaly-detection/pkg/investigations/CannotRetrieveUpdatesSRE"
```
joshbranham, line 4: Let's keep the existing pattern and not import this with a custom capitalized package name

pkg/investigations/CannotRetrieveUpdatesSRE/CannotRetrieveUpdatesSRE.go
```diff
@@ -0,0 +1,153 @@
...
+			if condition.Type == "RetrievedUpdates" &&
+				condition.Status == "False" &&
+				condition.Reason == "VersionNotFound" &&
+				strings.Contains(condition.Message, "Unable to retrieve available updates") {
+				i.notes.AppendWarning("ClusterVersion error detected: %s\nThis indicates the current version %s is not found in the specified channel %s",
+					condition.Message, versionCv.Status.Desired.Version, versionCv.Spec.Channel)
+			}
+		}
+		fmt.Printf("Cluster version: %s\n", versionCv.Status.Desired.Version)
+	}
```
joshbranham, line 125: This would be cleaner without a switch block. Just perform conditional checks for your each of your failures before doing the `default` behavior

anispate: I tried to do without a switch block but then one of the test was failing and it recommended me to use the switch case that time.

joshbranham: I don't follow, what do you mean "it recommended me to use the switch case that time"?

I think this is pointing to a broader issue that your `Run` here has a lot of business logic in one function, including performing multiple "validations" and needing to never return early.

If you break each "validation" into private functions, and call each in order from `Run` then you can handle the failure modes independent and it might be cleaner. Here is some pseudocode to work from. Let me know what you think and if you have any questions 👍 

```golang
func Run() {
	err := checkClusterVersion()
	if err != nil {
		// log warning
		// write warning note for pagerduty
	}
	
	err = runNetworkVerifier()
	if err != nil {
		// log warning
		// write warning note for pagerduty
	}
	
	// cleanup
}

func checkClusterVersion() error {
	cvList, err := getCvs()
	if err != nil {
		return err
	}
	
	if len(cvList) > 1 {
		return errors.New("More than one cluster version found")
	}
}

func runNetworkVerifier() error {
	result, err := runVerifier()
	if err != nil || result.Failed {
		return err
	}

	return nil
}
```

pkg/investigations/cannotretrieveupdatesre/cannotRetrieveUpdateSRE.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}(r)
+
+	if err := i.checkClusterValidity(r); err != nil {
+		logging.Errorf("Cluster validation failed: %v", err)
+		return result, r.PdClient.EscalateIncidentWithNote(i.notes.String())
+	}
+
+	if err := i.runNetworkVerifier(r, &result); err != nil {
+		logging.Errorf("Network verification failed: %v", err)
```
tnierman, line 54: We should probably handle this case when the network-verifier fails, right?

ie - add a note to the PD incident, prepare an SL (eventually), return a `result` and escalate to on-call, etc

pkg/investigations/cannotretrieveupdatessre/testing/README.md
tnierman, line 1: Looks like this is still the boilerplate document from running the `bootstrap-new-investigation` make target.

Can you revise this to include detailed steps for another SRE to test your work? Additional objects, scripts, etc can also be added to this `testing/` directory

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,164 @@
...
+			logging.Errorf("Failed to cleanup Kubernetes client: %v", cleanupErr)
+		} else {
+			logging.Infof("Cleanup completed successfully for cluster %s", r.Cluster.ID())
+		}
+	}(r)
+
+	verifierResult, failureReason, err := networkverifier.Run(r.Cluster, r.ClusterDeployment, r.AwsClient)
+	if err != nil {
+		logging.Error("Network verifier ran into an error: %s", err.Error())
+		notes.AppendWarning("NetworkVerifier failed to run:\n\t %s", err.Error())
```
tnierman, line 53: I don't believe we need to both log and note this - using `notes.AppendWarning` should be sufficient to add this messages to the logs as well

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,164 @@
...
+
+		err = r.PdClient.AddNote(notes.String())
+		if err != nil {
+			// We do not return as we want the alert to be escalated either no matter what.
+			logging.Error("could not add failure reason incident notes")
+		}
```
tnierman, line 59: Adding notes here without returning will mean that the network verifier results potentially get added to the incident twice: once here and again below, no?

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,164 @@
...
+	for _, condition := range clusterVersion.Status.Conditions {
+		logging.Debugf("Checking ClusterVersion condition: Type=%s, Status=%s, Reason=%s, Message=%q",
+			condition.Type, condition.Status, condition.Reason, condition.Message)
+		if condition.Type == "RetrievedUpdates" && condition.Status == "False" {
+			if (condition.Reason == "VersionNotFound" || condition.Reason == "RemoteFailed") &&
+				strings.Contains(strings.TrimSpace(condition.Message), "Unable to retrieve available updates") {
+				logging.Warnf("Detected ClusterVersion error: Reason=%s, Message=%s", condition.Reason, condition.Message)
+				return "", fmt.Sprintf("ClusterVersion error detected: %s. Current version %s not found in channel %s",
+						condition.Message, clusterVersion.Status.Desired.Version, clusterVersion.Spec.Channel),
+					fmt.Errorf("clusterversion validation failed: %s", condition.Reason)
```
tnierman, line 143: I find it a tad bit confusing that we're returning an error here: it's not the state we want to see the `clusterversion` in, but I'd expect an error return to mean that we _couldn't_ check the `clusterversion`, not that we were able to check the clusterversion but didn't like the state we saw it in

pkg/investigations/cannotretrieveupdatessre/testing/README.md
```diff
@@ -0,0 +1,5 @@
+# Testing CannotRetrieveUpdatesSRE Investigation
+
+TODO:
+- Add a test script or test objects to this  directory for future maintainers to use
+- Edit this README file and add detailed instructions on how to use the script/objects to recreate the conditions for the investigation. Be sure to include any assumptions or prerequisites about the environment (disable hive syncsetting, etc)
```
tnierman, line 5: This README still contains just the default text. Can we populate it with instructions on how to test this investigation?

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Run network verifier
+	verifierResult, failureReason, err := networkverifier.Run(r.Cluster, r.ClusterDeployment, r.AwsClient)
+	if err != nil {
+		notes.AppendWarning("NetworkVerifier failed to run:\n\t %s", err.Error())
+	} else {
+		switch verifierResult {
+		case networkverifier.Failure:
+			logging.Infof("Network verifier reported failure: %s", failureReason)
+			// XXX: metrics.Inc(metrics.ServicelogPrepared, investigationName)
```
tnierman, line 52:
```suggestion

```

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,124 @@
...
+	}()
+
+	// Run network verifier
+	verifierResult, failureReason, err := networkverifier.Run(r.Cluster, r.ClusterDeployment, r.AwsClient)
+	if err != nil {
+		notes.AppendWarning("NetworkVerifier failed to run:\n\t %s", err.Error())
+	} else {
+		switch verifierResult {
+		case networkverifier.Failure:
+			logging.Infof("Network verifier reported failure: %s", failureReason)
```
tnierman, line 51:
```suggestion

```
We probably can exclude this logging line since we're noting the failure below too?

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
anispate: I was going to remove it but it exists on other cpd.go and insightoperatordown.go file so, i kept it.

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,124 @@
...
+	logging.Infof("ClusterVersion found: %s", clusterVersion.Status.Desired.Version)
+
+	for _, condition := range clusterVersion.Status.Conditions {
+		if condition.Type == "RetrievedUpdates" && condition.Status == "False" {
+			if (condition.Reason == "VersionNotFound" || condition.Reason == "RemoteFailed") &&
+				strings.Contains(strings.TrimSpace(condition.Message), "Unable to retrieve available updates") {
+				logging.Warnf("Detected ClusterVersion issue: Reason=%s, Message=%s", condition.Reason, condition.Message)
+				return "", fmt.Sprintf("ClusterVersion issue detected: %s. Current version %s not found in channel %s",
+						condition.Message, clusterVersion.Status.Desired.Version, clusterVersion.Spec.Channel),
+					fmt.Errorf("clusterversion has undesirable state: %s", condition.Reason)
```
tnierman, line 103: The way this function is written and called [above](https://github.com/openshift/configuration-anomaly-detection/pull/404/files#diff-761565fefa17061f704f9cb537f6fbf1a357cc959e9fef51467520bb78f8132bR61-R71), returning an error here won't allow the note to be added to the PD incident.

I'd prefer if we didn't return an error at all, honestly, because that should be reserved for cases where we couldn't check the state of the cluster, rather than cases where we did check the state of the cluster, but didn't like what we saw.

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+		}
+	}
```
tnierman, line 106:
```suggestion
	for _, condition := range clusterVersion.Status.Conditions {
		if condition.Type == "RetrievedUpdates" {
		    note, err := checkRetrievedUpdatesCondition(condition)
		    return clusterVersion.Status.DesiredVersion, note, err
		}
	}
```
It looks like we're only looking to analyze the state of one condition here, correct? Would it make more sense to break some of this out into it's own function? 

Something like the following may make the nested conditionals more readable:

```
func checkRetrievedUpdatesCondition(condition corev1.StatusCondition) (string, err) {
    if condition.Status == corev1.ConditionFalse {
        return ...
    }
    if (condition.Reason == ...) {
        return ...
    }
}
```

Additionally, we won't waste CPU cycles continuing to loop through conditions we don't need to analyze once we find the one we're after.

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
tnierman: I think we should remove it for now :slightly_smiling_face:  If/when we add metrics to CAD, we'll have to scan the codebase for instances of servicelogs being sent anyway

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,136 @@
...
+	remediationName = "CannotRetrieveUpdatesSRE"
+)
+
+type Investigation struct{}
+
+// Run executes the investigation for the CannotRetrieveUpdatesSRE alert
+func (c *Investigation) Run(r *investigation.Resources) (investigation.InvestigationResult, error) {
+	result := investigation.InvestigationResult{}
+	notes := notewriter.New("CannotRetrieveUpdatesSRE", logging.RawLogger)
+	k8scli, err := k8sclient.New(r.Cluster.ID(), r.OcmClient, remediationName)
```
typeid, line 29: **issue:** I think this will fail as backplane will look for the `metadata.yaml` in the path `pkg/investigations/CannotRetrieveUpdatesSRE/metadata.yaml`, but the path is lowercase.

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,136 @@
...
+	notes.AppendWarning("Alert escalated to on-call primary for review.")
+	logging.Infof("Escalating incident with notes for cluster %s", r.Cluster.ID())
+	err = r.PdClient.EscalateIncidentWithNote(notes.String())
+	if err != nil {
+		logging.Errorf("Failed to escalate incident to PagerDuty: %v", err)
+		return result, fmt.Errorf("failed to escalate incident: %w", err)
+	}
+	logging.Infof("Investigation completed and escalated successfully for cluster %s", r.Cluster.ID())
+
+	return result, nil
```
typeid, line 83: We don't need to add in the PD note that we escalated the alert to SRE, it should be obvious if the SRE has the alert on their board ;) 

There might be a few superfluous logs here as well, the following should suffice: 

```suggestion
	return result, r.PdClient.EscalateIncidentWithNote(notes.String())
```

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// checkClusterVersion retrieves the cluster version
+func checkClusterVersion(k8scli client.Client, clusterID string) (version string, note string, err error) {
```
typeid, line 87: This function does more than the name suggests:
- it retrieves the cluster version
- checks if the updates could be retrieved
- pre-formats a note

Could be separate the logic here for the functions to be more clear cut on what they are doing? 

E.g.

`func getClusterVersion()` -> no logs, just gets the cluster version object
`func getUpdateRetrievalFailures(clusterversion)` -> looks for update retrieval failures in the clusterversion 

Logs and note creation should be outside of this logic, ideally in the main investigation function logic. This way, we could even move the above functions in common packages as they are re-usable.

typeid (changes requested):

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
anispate: But, i have updated the path as part of the PR to the lowercase for the metadata.yaml as part of the PR so, would it still cause an issue?

typeid: `remediationName` uses camel case in your PR, correct? And the path is not camel case in this PR, so it's a mismatch

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,117 @@
...
+	if condition.Type != "RetrievedUpdates" {
+		return "", false
+	}
+	if condition.Status == configv1.ConditionFalse {
+		return fmt.Sprintf("(Reason: %s). %s", condition.Reason, condition.Message), true
+	}
+	return "", false
+}
+
+func (i *Investigation) Name() string {
```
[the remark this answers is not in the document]
tnierman: They're required methods for an `Investigation` to satisfy [this interface](https://github.com/openshift/configuration-anomaly-detection/blob/main/pkg/investigations/investigation/investigation.go#L24-L32), yeah

pkg/investigations/cannotretrieveupdatessre/cannotretrieveupdatessre.go
```diff
@@ -0,0 +1,117 @@
...
+			logging.Warnf("Detected ClusterVersion issue: %s", failureReason)
+			notes.AppendWarning("ClusterVersion related issue detected: %s. Current version %s not found in channel %s",
+				failureReason, clusterVersion.Status.Desired.Version, clusterVersion.Spec.Channel)
+		}
+	}
+	notes.AppendWarning("Alert escalated to on-call primary for review and please check the ClusterVersion.")
+	return result, r.PdClient.EscalateIncidentWithNote(notes.String())
+}
+
+func getClusterVersion(k8scli client.Client) (*configv1.ClusterVersion, error) {
```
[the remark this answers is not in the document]
tnierman: +1 - fwiw other CAD maintainers have expressed a desire to subdivide the [`pkg/investigation/utils/` packages](https://github.com/openshift/configuration-anomaly-detection/tree/main/pkg/investigations/utils) into focused units. I think currently, we only have `node` and `machine` utilities packages, so I'd recommend creating a new one (maybe called `version`?)

PR #421, opened 2025-04-10
Make tekton pruner image configurable via app-interface

RaphaelBut:
https://issues.redhat.com/browse/OSD-29266
We have not updated this image's version in a long time. This makes it configurable in the template, so we can easier keep track, possibly adding it to the promotion process.


typeid (approved):
/lgtm
/approve

PR #422, opened 2025-04-10
Remove tekton pipeline pruner

RaphaelBut:
We can manage the tekton config via app-interface, the defaults however should be fine for current usage.

Allows us to not have to worry about the tekton pruner image, see 
https://issues.redhat.com/browse/OSD-29264
https://issues.redhat.com/browse/OSD-29266


RaphaelBut:
updated the image here
I see there will be updates to the tekton pruner soon ( 1.19 ) which should allow us to configure the pruning more easily on a resource level ( PipelineRuns ). Will revisit this then.

PR #424, opened 2025-04-11
OSD-29424 - Update PD incident title after investigation

tnierman:
https://issues.redhat.com/browse/OSD-29424

---

These changes update the PagerDuty incident title after an investigation completes to more visibly indicate when CAD has run. The goal is to reduce the potential for duplicated work as a result of on-call not seeing the investigation notes CAD attaches to the incident.

The title is only updated once: subsequent CAD runs will not re-attach the `[CAD Investigated]` prefix to the incident, in order to keep titles useful across multiple investigation attempts

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 	investigations "github.com/openshift/configuration-anomaly-detection/pkg/investigations"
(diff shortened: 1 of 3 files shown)


cadctl/cmd/investigate/investigate.go
```diff
@@ -231,3 +237,16 @@ func updateMetrics(investigationName string, result *investigation.Investigation
 		metrics.Inc(metrics.LimitedSupportSet, append([]string{investigationName}, result.LimitedSupportSet.Labels...)...)
 	}
 }
+
+func updateIncidentTitle(pdClient *pagerduty.SdkClient) error {
+	currentTitle := pdClient.GetTitle()
+	if strings.Contains(currentTitle, pagerdutyTitlePrefix) {
+		return nil
+	}
+	newTitle := fmt.Sprintf("[CAD Investigated] %s", pdClient.GetTitle())
```
joshbranham, line 246: Let's just use the existing `currentTitle` value here to save an API call
```suggestion
	newTitle := fmt.Sprintf("[CAD Investigated] %s", currentTitle)
```

tnierman: Whoops, good catch

typeid (approved):
/lgtm
/approve

PR #425, opened 2025-04-14
OSD-29424 - Bugfix to use `pagerdutyTitlePrefix` instead of hardcoded string when updating incident title

tnierman:
Fixes minor bug where the `pagerDutyTitlePrefix` const wasn't being used to prefix the PagerDuty incident title upon investigation completion.


typeid (approved):
/lgtm
/approve

PR #426, opened 2025-04-16
Maint: handle degraded api when creating k8sclient

typeid:
This PR adds handling for degraded api errors we can receive when calling backplane to create our kubeconfig/permission mappings. I also split up the k8sclient package for separation of concerns.

### DIFF
--- a/pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
+++ b/pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
@@ -36,6 +36,10 @@ func (c *Investigation) Run(r *investigation.Resources) (result investigation.In
 	// patching the existing RBAC etc...
 	k8scli, err := k8sclient.New(r.Cluster.ID(), r.OcmClient, r.Name)
 	if err != nil {
+		if errors.Is(k8sclient.ErrAPIServerUnavailable, err) {
+			return result, r.PdClient.EscalateIncidentWithNote("CAD was unable to  cluster's kube-api. Please investigate manually.")
+		}
+
 		return result, fmt.Errorf("unable to initialize k8s cli: %w", err)
 	}
 	defer func() {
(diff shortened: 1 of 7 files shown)


pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
```diff
@@ -36,6 +36,10 @@ func (c *Investigation) Run(r *investigation.Resources) (result investigation.In
 	// patching the existing RBAC etc...
 	k8scli, err := k8sclient.New(r.Cluster.ID(), r.OcmClient, r.Name)
 	if err != nil {
+		if errors.Is(k8sclient.ErrAPIServerUnavailable, err) {
+			return result, r.PdClient.EscalateIncidentWithNote("CAD was unable to  cluster's kube-api. Please investigate manually.")
```
RaphaelBut, line 40: access :D ?

pkg/k8s/errors.go
```diff
@@ -0,0 +1,17 @@
...
+	"errors"
+	"strings"
+)
+
+var ErrAPIServerUnavailable = errors.New("kubernetes API server unavailable")
+
+// isAPIServerUnavailable detects common symptoms of an unreachable API server.
+func isAPIServerUnavailable(err error) bool {
+	errStr := err.Error()
+	return strings.Contains(errStr, "TLS handshake timeout") ||
```
RaphaelBut, line 13: thought: wondering if we can at all infer anything from the different errors in terms of what might be the issue?

jbpratt: random drive-by ;) you can also use apimachinery to check errors https://github.com/openshift/osde2e-common/blob/7d13c0222fee1854e29e4e15d4ada896a1dfbaf4/pkg/clients/openshift/errors.go#L10

typeid: Backplane doesn't forward the errors in their original state, changed this PR to just use `The cluster could be down or under heavy load` which is the general backplane string for unreachable api.

PR #427, opened 2025-04-23
OSD-29208 - Remove CAD template file generation

MateSaary:
This PR deprecates CAD's template file generation, in favour of allowing for manual direct editing of the template.yaml file.
The PR also removes PD secrets from the template as it is understood that these are redundant, as they are generated by app-interface with the vault provider resource.


RaphaelBut:
This looks good to me. I wonder if we should remove the whole deploy directory now and just rely on the templates in the openshift folder?

MateSaary:
@RaphaelBut That's alright to me! I've done so now.

PR #428, opened 2025-04-25
Initial commit for AI POC

RaphaelBut:
Testing on this incident here https://redhat.pagerduty.com/incidents/Q3C8N13COTT47K
Currently the output looks something like this 
```
🤖 Based on the provided `ClusterOperator` status and events, here are some key points to investigate:

1. **Check AlertManager and Prometheus Status:**
   - Ensure that the `AlertManager` and `Prometheus` instances are running and healthy. Look for any errors or restarts in the events or operator logs.

2. **Review ServicMonitors and PodMonitors:**
   - Verify that `Servicemonitors` and `Podmonitors` are correctly configured and that the Prometheus instances are scraping the endpoints as expected.

3. **Investigate PrometheusRules:**
   - Check the `Prometheusrules` to ensure there are no syntax errors or misconfigured rules that might be causing alerts to fire incorrectly.

4. **Examine ThanosRuler:**
   - If `ThanosRuler` is in use, verify its configuration and status. Ensure it is properly integrated with Prometheus.

5. **Check Conditions:**
   - The conditions show `Degraded` as `False`, but it's essential to review the full logs and events around the period when the error budget burn alert was triggered.

6. **Review Recent Changes:**
   - Look for any recent changes in the configuration or updates to the monitoring stack that might have led to the error budget burn.

7. **operator not updated:**
  - The operator has not been updated recently - update if possible and monitor.

To gather more information, you can look into the corresponding logs from the `openshift-monitoring` and `openshift-user-workload-monitoring` namespaces. Check for any errors or warnings in the logs of Pods and Operator.
```

### DIFF
--- a/pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
+++ b/pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
@@ -54,13 +54,6 @@ func (c *Investigation) Run(r *investigation.Resources) (result investigation.In
 	notes := notewriter.New(r.Name, logging.RawLogger)
 	defer func() { r.Notes = notes }()
 
-	// AI POC
-	aiOutput, err := r.AIClient.Ask(fmt.Sprintf("How do I react to this alert: %s", r.Name))
-	if err != nil {
-		notes.AppendWarning("Failed to get AI answer: %s", err)
-	}
-	notes.AppendAutomation(aiOutput)
-
 	// List the monitoring cluster operator
 	coList := &configv1.ClusterOperatorList{}
 	listOptions := &client.ListOptions{FieldSelector: fields.SelectorFromSet(fields.Set{"metadata.name": "monitoring"})}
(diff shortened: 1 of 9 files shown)


pkg/investigations/clustermonitoringerrorbudgetburn/clustermonitoringerrorbudgetburn.go
```diff
@@ -82,6 +82,13 @@ func (c *Investigation) Run(r *investigation.Resources) (result investigation.In
 		return result, r.PdClient.SilenceIncidentWithNote(notes.String())
 	}
 
+	// AI POC
```
typeid, line 85: Thought: If we merge this and promote, we'll use this on production and bloat the notes for this alert. Maybe adding this POC to a new investigation that we toggle via the experimental flag is a better idea for now?

RaphaelBut: Yes, let me convert this to a draft for now
I think we might want to go in a direction of running this blindly against a bunch of alerts and put in the context we have available instead of doing specific questions in investigations, something along those lines

RaphaelBut:
superseded by https://github.com/openshift/configuration-anomaly-detection/pull/439

PR #429, opened 2025-04-30
OSD-28525 - Additional permission for CAD to retrieve machinehealthchecks on-cluster

tnierman:
Adds an additional object to the `machinehealthcheckUnterminatedShortCircuit` investigation's `openshift-machine-api` role, so that the investigation can retrieve the machinehealthcheck object it's investigating


joshbranham (commented):
/lgtm
/approve

PR #430, opened 2025-05-01
Create CAD remediation for UpgradeConfigSyncFailureOver4HrSRE

Mhodesty:
This PR adds an investigation for the UpgradeConfigSyncFailureOver4HrSRE alert. It is currently read only.

This investigation checks if a user on a cluster is banned and compares the pull secret in OCM vs the pull-secret on the cluster to confirm they are the same. It also includes a README on how to integration test this functionality and custom mocking for the getClusterPullSecret functionality.


pkg/investigations/upgradeconfigsyncfailureover4hr/upgradeconfigsyncfailureover4hr.go
```diff
@@ -0,0 +1,124 @@
...
+		return "", "Cluster pull secret does not contain the necessary .dockerconfigjson", err
+	}
+
+	dockerConfigJson, err := v1.UnmarshalAccessToken(secretValue)
+	if err != nil {
+		return "", "", err
+	}
+	_, exists = dockerConfigJson.Auths()["cloud.openshift.com"]
+	if !exists {
+		return "", "cloud.openshift.com value not found in clusterPullSecret. This almost certainly means there is an issue with the pull secret on the cluster.", err
```
tnierman, line 99:
```suggestion
		return "", "cloud.openshift.com value not found in clusterPullSecret. This means there is an issue with the pull secret on the cluster.", err
```

I think our automation should be assertive about its conclusions :slightly_smiling_face: For this case specifically, I don't see there being any reason why a default domain would be missing, so we can safely conclude it's misconfigured

pkg/investigations/upgradeconfigsyncfailureover4hr/upgradeconfigsyncfailureover4hr_test.go
```diff
@@ -0,0 +1,67 @@
...
+					Namespace: "openshift-config",
+				},
+				Type: corev1.DockerConfigJsonKey,
+				Data: map[string][]byte{
+					".dockerconfigjson": []byte(tt.data),
+				},
+			}
+			k8scli := fake.NewClientBuilder().WithObjects(secretTest).Build()
+			result, note, _ := getClusterPullSecret(k8scli)
+			fmt.Printf("Note is %s", note)
```
tnierman, line 56: minor nit: we report the `note` and `expectedNote` below if the test fails, so we may be able to omit this line?

pkg/ocm/ocm.go
```diff
@@ -310,6 +310,21 @@ func (c *SdkClient) IsAccessProtected(cluster *cmv1.Cluster) (bool, error) {
...
+	user, err := GetCreatorFromCluster(ocmClient.GetConnection(), cluster)
+	if err != nil {
+		return false, "encountered an issue when checking if the cluster owner is banned. Please investigate.", err
+	}
+
+	if user.Banned() {
+		noteMessage := fmt.Sprintf("User is banned %s. Ban description %s.\n Please open a proactive case, so that MCS can resolve the ban or organize a ownership transfer.", user.BanCode(), user.BanDescription())
+		logging.Warnf(noteMessage)
+		return true, noteMessage, nil
+	} else {
```
tnierman, line 323: minor nit: we could omit the `else` here :slightly_smiling_face: Just highlighting this because some linters are picky about that

pkg/investigations/upgradeconfigsyncfailureover4hr/upgradeconfigsyncfailureover4hr_test.go
Mhodesty: Absolutely. This must have been from a test that I was running and forgot to clean up. Nice catch.

PR #432, opened 2025-05-02
[WIP] SREP-11 - Enforce informing mode on new investigations

MateSaary:


pkg/investigations/ccam/ccam.go
```diff
@@ -80,6 +80,11 @@ func (c *Investigation) IsExperimental() bool {
 	return false
 }
 
+func (c *Investigation) InformingMode(flag bool) bool {
+	informingOnly := false
+	return informingOnly || flag
```
typeid, line 85: Why does this take a parameter? Why not just 
```
func (c *Investigation) InformingMode() bool {
	return false
```

typeid (commented):
If we want to enforce informing, it would need to be on:
- kube api operations
- ocm write operations
- aws write operations

As is, IMO this PR adds a lot of unnecessary complexity by having to pass the informing mode through everything. An alternative I can think of here would be to add some kind of make target linting for investigations initially being created as informing.

Maybe we're overengineering here though: we have reviewers for a reason. The guidelines should be good enough.

MateSaary:
@typeid Thanks for the review :slightly_smiling_face: Great points, a lot of this does overlap with reviews. Do we think it's worth looking into implementing guidelines elsewhere (e.g. PR template akin to MR in App-interface)?
For kube-api, I wrote a short make target script to "lint" investigations and their corresponding RBAC to ensure investigations set to informing cannot conduct write operations. Let me know your thoughts. Thanks again!

pkg/investigations/cpd/cpd.go
```diff
@@ -65,14 +65,16 @@ func (c *Investigation) Run(r *investigation.Resources) (investigation.Investiga
 				logging.Error(err)
 			}
 			if !isValid {
-				if err := r.OcmClient.PostServiceLog(r.Cluster.ID(), byovpcRoutingSL); err != nil {
+				if err := r.OcmClient.PostServiceLog(r.Cluster.ID(), byovpcRoutingSL, c.InformingMode()); err != nil {
```
RaphaelBut, line 68: thought: Can we somehow, make the investigation code agnostic to the informing mode?
it would be cool to avoid adding the condition L75-77 below.

I like the extra parameter we pass to functions in the ocm package here, https://github.com/openshift/configuration-anomaly-detection/pull/432/files#diff-83e5cc2dfa3554ded718960969e8a817d6791f7e3c9c769a1136e6a0a3f94820R211 
but suspect we do not want to add this logic to the ocm package

something along the lines of having a layer where the note posting to PD and actions taken are evaluated against the informing mode.

RaphaelBut: I am not too sure however if this approach is the best idea

How would we deal with changes to the investigation? 
I dont think we would want to fallback to informing mode for the whole investigation, but if the newly added condition is hit, would we like SRE to verify?

pkg/investigations/chgm/chgm.go
```diff
@@ -450,7 +454,7 @@ func extractUserDetails(cloudTrailEvent *string) (CloudTrailEventRaw, error) {
 
 // postStoppedInfraLimitedSupport will put the cluster on limited support because the user has stopped instances
 func postStoppedInfraLimitedSupport(clusterID string, ocmCli ocm.Client, pdCli pagerduty.Client) error {
-	err := ocmCli.PostLimitedSupportReason(&stoppedInfraLS, clusterID)
+	err := ocmCli.PostLimitedSupportReason(&stoppedInfraLS, clusterID, false)
```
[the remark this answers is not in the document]
MateSaary: Great spot, thanks!

MateSaary:
Closing PR following discussion https://redhat-internal.slack.com/archives/C081RTBAPA9/p1751354368278599

PR #436, opened 2025-05-07
Fix: ignore non-existant clusters

typeid:
This PR gets rid of all the non-0 exiting pipelines for clusters that don't exist (e.g. integration cluster ids in staging). This is mostly to suppress failed pipeline noise, should not happen on production.


rafael-azevedo (approved):
/lgtm

PR #437, opened 2025-05-07
Simplify the configuration loading

petrkotas:
The code simplifies the ocm config loading. It also moves the env variables right to the start, 
so it is obvious when they are used.

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -19,7 +19,6 @@ package investigate
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 	"strings"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
@@ -35,12 +34,82 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var debugFlag = false
+var experimentalFlag = false
+var cadOcmClientID string
+var cadOcmClientSecret string
+var cadOcmURL string
+
 // InvestigateCmd represents the entry point for alert investigation
 var InvestigateCmd = &cobra.Command{
 	Use:          "investigate",
 	SilenceUsage: true,
 	Short:        "Filter for and investigate supported alerts",
-	RunE:         run,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		debugFlag, err := cmd.Flags().GetBool("debug")
+		if err != nil {
+			return fmt.Errorf("failed to get debug flag: %w", err)
+		}
+
+		if !debugFlag {
+			// check environment variable
+			debug, exist := os.LookupEnv("CAD_DEBUG")
+			if exist && debug == "true" {
+				debugFlag = true
+			}
+		}
+
+		if debugFlag && !ocm.HasConfigFile() {
+			return fmt.Errorf("debug flag is set, but no OCM config file found. When using debug mode the OCM config file must be set")
+		}
+
+		var exist bool
+		cadOcmURL, err = cmd.Flags().GetString("ocm-url")
+		if err != nil {
+			return fmt.Errorf("failed to get ocm-url flag: %w", err)
+		}
+		if cadOcmURL == "" {
+			cadOcmURL, exist = os.LookupEnv("CAD_OCM_URL")
+			if !exist && !debugFlag {
+				return fmt.Errorf("ocm-url flag or CAD_OCM_URL environment variable must be set")
+			}
+		}
+
+		cadOcmClientID, err = cmd.Flags().GetString("client-id")
+		if err != nil {
+			return fmt.Errorf("failed to get client-id flag: %w", err)
+		}
+		if cadOcmClientID == "" {
+			cadOcmClientID, exist = os.LookupEnv("CAD_OCM_CLIENT_ID")
+			if !exist && !debugFlag {
+				return fmt.Errorf("client-id flag or CAD_OCM_CLIENT_ID environment variable must be set")
+			}
+		}
+		cadOcmClientSecret, err = cmd.Flags().GetString("client-secret")
+		if err != nil {
+			return fmt.Errorf("failed to get client-secret flag: %w", err)
+		}
+		if cadOcmClientSecret == "" {
+			cadOcmClientSecret, exist = os.LookupEnv("CAD_OCM_CLIENT_SECRET")
+			if !exist && !debugFlag {
+				return fmt.Errorf("client-secret flag or CAD_OCM_CLIENT_SECRET environment variable must be set")
+			}
+		}
+
+		experimentalFlag, err = cmd.Flags().GetBool("experimental")
+		if err != nil {
+			return fmt.Errorf("failed to get experimental flag: %w", err)
+		}
+		if !experimentalFlag {
+			experimental, exist := os.LookupEnv("CAD_EXPERIMENTAL_ENABLED")
+			if exist && experimental == "true" {
+				experimentalFlag = true
+			}
+		}
+
+		return nil
+	},
+	RunE: run,
 }
 
 var (
@@ -53,6 +122,11 @@ const pagerdutyTitlePrefix = "[CAD Investigated]"
 func init() {
 	InvestigateCmd.Flags().StringVarP(&payloadPath, "payload-path", "p", payloadPath, "the path to the payload")
 	InvestigateCmd.Flags().StringVarP(&logging.LogLevelString, "log-level", "l", "", "the log level [debug,info,warn,error,fatal], default = info")
+	InvestigateCmd.Flags().BoolP("debug", "d", false, "enable debugging mode, if not present CAD_DEBUG env is checked")
+	InvestigateCmd.Flags().StringP("ocm-url", "u", "", "the OCM URL, when not set the CAD_OCM_URL environment is used.")
+	InvestigateCmd.Flags().StringP("client-id", "i", "", "the OCM client ID, when not set the CAD_OCM_CLIENT_ID environment is used.")
+	InvestigateCmd.Flags().StringP("client-secret", "s", "", "the OCM client secret, when not set the CAD_OCM_CLIENT_SECRET environment is used.")
+	InvestigateCmd.Flags().BoolP("experimental", "e", false, "enable experimental features")
 
 	err := InvestigateCmd.MarkFlagRequired("payload-path")
 	if err != nil {
@@ -87,8 +161,7 @@ func run(cmd *cobra.Command, _ []string) error {
 		}
 	}()
 
-	_, cadExperimentalEnabled := os.LookupEnv("CAD_EXPERIMENTAL_ENABLED")
-	alertInvestigation := investigations.GetInvestigation(pdClient.GetTitle(), cadExperimentalEnabled)
+	alertInvestigation := investigations.GetInvestigation(pdClient.GetTitle(), experimentalFlag)
 
 	// Escalate all unsupported alerts
 	if alertInvestigation == nil {
@@ -109,8 +182,14 @@ func run(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	ocmClient, err := GetOCMClient()
-	if err != nil {
+	var ocmClient *ocm.SdkClient
+	var ocmErr error
+	if debugFlag {
+		ocmClient, ocmErr = ocm.NewFromConfig()
+	} else {
+		ocmClient, ocmErr = ocm.NewFromClientKeyPair(cadOcmURL, cadOcmClientID, cadOcmClientSecret)
+	}
+	if ocmErr != nil {
 		return fmt.Errorf("could not initialize ocm client: %w", err)
 	}
 
@@ -182,22 +261,6 @@ func handleCADFailure(err error, resources *investigation.Resources, pdClient *p
 	} else {
 		logging.Errorf("Failed to obtain PagerDuty client, unable to escalate CAD failure to PagerDuty notes.")
 	}
-}
-
-// GetOCMClient will retrieve the OcmClient from the 'ocm' package
-func GetOCMClient() (*ocm.SdkClient, error) {
-	cadOcmFilePath := os.Getenv("CAD_OCM_FILE_PATH")
-
-	_, err := os.Stat(cadOcmFilePath)
-	if os.IsNotExist(err) {
-		configDir, err := os.UserConfigDir()
-		if err != nil {
-			return nil, err
-		}
-		cadOcmFilePath = filepath.Join(configDir, "/ocm/ocm.json")
-	}
-
-	return ocm.New(cadOcmFilePath)
 }
 
 // Checks pre-requisites for a cluster investigation:
--- a/cadctl/cmd/root.go
+++ b/cadctl/cmd/root.go
@@ -17,16 +17,21 @@ limitations under the License.
 package cmd
 
 import (
+	"fmt"
+	"os"
+
 	investigate "github.com/openshift/configuration-anomaly-detection/cadctl/cmd/investigate"
 	"github.com/openshift/configuration-anomaly-detection/pkg/logging"
 	"github.com/openshift/configuration-anomaly-detection/pkg/metrics"
+	ocm "github.com/openshift/configuration-anomaly-detection/pkg/ocm"
 	"github.com/spf13/cobra"
 )
 
 // rootCmd represents the base command when called without any subcommands
 var rootCmd = &cobra.Command{
-	Use:   "cadctl",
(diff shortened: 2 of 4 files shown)


cadctl/cmd/investigate/investigate.go
```diff
@@ -35,12 +34,82 @@ import (
...
+			// check environment variable
+			debug, exist := os.LookupEnv("CAD_DEBUG")
+			if exist && debug == "true" {
+				debugFlag = true
+			}
+		}
+
+		if debugFlag && !ocm.HasConfigFile() {
+			return fmt.Errorf("debug flag is set, but no OCM config file found. When using debug mode the OCM config file must be set")
+		}
```
bergmannf, line 64: If this has a config-file shouldn't all the other variables be retrieved from the config-file?

This would make it a bit faster, by short-circuiting the rest of the flag-handling by never even doing it with `--debug` being set + `config-file` exists.

pkg/ocm/ocm.go
```diff
@@ -57,51 +55,48 @@ type SdkClient struct {
...
 // if no path is provided, it will assume it in the default path
-func New(ocmConfigFile string) (*SdkClient, error) {
+func NewFromClientKeyPair(ocmURL string, clientID string, clientSecret string) (*SdkClient, error) {
 	var err error
 	client := SdkClient{}
 
-	// The debug environment variable ensures that we will never use
-	// an ocm config file on a cluster deployment. The debug environment variable
-	// is only for local cadctl development
-	debugMode := os.Getenv("CAD_DEBUG")
```
bergmannf, line 67:
```suggestion
	// The debug environment variable ensures that we will never use
	// an ocm config file on a cluster deployment. The debug environment variable
	// is only for local cadctl development
	debugMode := os.Getenv("CAD_DEBUG")
```
Probably a good idea to keep at least this comment somewhere, because I initially expected the `CAD_DEBUG` to be more something along the lines of a debug-level logging setting, not something that changes behaviour that much.

pkg/ocm/ocm_config.go
```diff
@@ -9,6 +9,10 @@ import (
 	sdk "github.com/openshift-online/ocm-sdk-go"
 )
 
+// ocmConfigFile is the path refers to existing ocm config file
+// set by the ocm command by calling ocm login
+var ocmConfigFile string
```
bergmannf, line 14: Based on the deleted comment that this should only be used locally this is probably ok.

However if this ever gets used in production with a long-running CAD process this could potentially lead to weird race-conditions if multiple investigations try to set this to different values and might need to be wrapped in a `sync.Mutex` to change.

(Or even better, not be a global variable but be set per-investigation)

petrkotas: The config should be used only locally not in prod. So this is not an issue.

cadctl/cmd/investigate/investigate.go
```diff
@@ -35,12 +34,82 @@ import (
...
+var cadOcmClientSecret string
+var cadOcmURL string
+
 // InvestigateCmd represents the entry point for alert investigation
 var InvestigateCmd = &cobra.Command{
 	Use:          "investigate",
 	SilenceUsage: true,
 	Short:        "Filter for and investigate supported alerts",
-	RunE:         run,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
petrkotas: I extracted it to minimum that makes sense here.

cadctl/cmd/investigate/investigate.go
```diff
@@ -40,19 +41,42 @@ var InvestigateCmd = &cobra.Command{
...
+			if _, err = config.Config.GetCadOcmURL(cmd); err != nil {
+				return fmt.Errorf("failed to get cadOcmURL: %w", err)
+			}
+		}
+
+		if _, err = config.Config.GetExperimental(cmd); err != nil {
+			return fmt.Errorf("failed to get experimental flag: %w", err)
+		}
+
+		return nil
```
typeid, line 65: Why not just a `config, err := config.Load()` wrapping all of this? We're also introducing a new antipattern (global `config` variable) here, can we instead use a properly scoped variable here? 

If we want to specifically load only what we need because we think it will be different for manually triggered stuff, we could have two configuration types that could be loaded.

cadctl/config/config.go
```diff
@@ -0,0 +1,157 @@
...
+		// Location() checks for file existence
+		return "", fmt.Errorf("failed to get ocmconfig file location: %w", err)
+	}
+
+	c.ConfigFile = ocmConfigPath
+	return ocmConfigPath, nil
+}
+
+func (c *config) GetLogLevel(cmd *cobra.Command) (string, error) {
+	logLevel, err := cmd.Flags().GetString("log-level")
```
typeid, line 82: This flag doesn't exist, this causes things to error out if you run CAD:
```
/bin/cadctl investigate --payload-path payload 
Warning: Unable to retrieve the pipeline ID on logger creation. Continuing with empty value.
Warning: Unable to retrieve the pipeline ID on logger creation. Continuing with empty value.
{"level":"fatal","timestamp":"2025-05-14T10:38:59+02:00","caller":"logging/logging.go:68","msg":"failed to get log-level flag: flag accessed but not defined: log-level","cluster_id":"","pipeline_name":""}
```

cadctl/cmd/investigate/investigate.go
```diff
@@ -40,19 +41,42 @@ var InvestigateCmd = &cobra.Command{
...
 var (
-	logLevelFlag = ""
-	payloadPath  = "./payload.json"
+	payloadPath = "./payload.json"
 )
 
 const pagerdutyTitlePrefix = "[CAD Investigated]"
 
 func init() {
+	logging.RawLogger = logging.InitLogger(config.Config.LogLevel, "")
```
typeid, line 75: We're initializing the `logging.RawLogger` variable twice now: once in `logging.go` and once here.

interceptor/main.go
```diff
@@ -71,6 +71,18 @@ func main() {
 	logger.Infof("Server exiting")
 }
 
+func init() {
+	// read log level from environment variable
+	logLevel := os.Getenv("LOG_LEVEL")
+	if logLevel != "" || logLevel == "info" {
```
typeid, line 77: Is it supposed to be `logLevel == "" || logLevel == "info"`? 

Either way I think what we want is two different loggers to load:
- one initialized on the cluster id
- one initialized without cluster id
Both of those should be able to receive a log level. Shall we move this logic into `logging.go`, and we'll be able to re-use accross interceptor, investigate and everything else?

cadctl/cmd/investigate/investigate.go
```diff
@@ -109,8 +126,17 @@ func run(cmd *cobra.Command, _ []string) error {
...
+	var ocmErr error
+	// Initialize the OCM client
+	// If the config file is set, use it to initialize the OCM client
+	// use config file only when developing
+	if config.Config.ConfigFile != "" {
+		ocmClient, ocmErr = ocm.NewFromConfig(config.Config.ConfigFile)
+	} else {
+		ocmClient, ocmErr = ocm.NewFromClientKeyPair(config.Config.CadOcmURL, config.Config.CadOcmClientID, config.Config.CadOcmClientSecret)
+	}
+	if ocmErr != nil {
```
typeid, line 139: **Nit:** this is probably logic we could hide and only pass encessary config to the ocm client init. I see us copy pasting this piece of code if we don't do it. Thoughts?

cadctl/cmd/investigate/investigate.go
```diff
@@ -122,13 +148,8 @@ func run(cmd *cobra.Command, _ []string) error {
...
 
-	// re-initialize logger for the internal-cluster-id context
-	// if log-level flag is set, take priority over env + default
-	if cmd.Flags().Changed("log-level") {
-		logging.RawLogger = logging.InitLogger(logLevelFlag, internalClusterID)
-	} else {
-		logging.RawLogger = logging.InitLogger(logging.LogLevelString, internalClusterID)
-	}
+	// add the internal-cluster-id context to the logger
+	logging.RawLogger = logging.RawLogger.With(zap.String("cluster_id", internalClusterID))
```
typeid, line 152: Does this keep the log level? Neat!

cadctl/cmd/root.go
```diff
@@ -41,4 +45,22 @@ func Execute() {
...
+	rootCmd.PersistentFlags().StringP("ocmconfig", "c", "", "Path to the OCM config file to use for CLI requests. When not set the OCM_CONFIG environment variable will be used. If that is not set, the default OCM config locations will be file will be used.")
+}
+
+func loadCommon(cmd *cobra.Command, args []string) error {
+	var err error
+
+	_, err = config.Config.GetLogLevel(cmd)
+	if err != nil {
+		logging.Fatal(err)
+		return err
```
typeid, line 57: Doesn't fatal already return? https://pkg.go.dev/go.uber.org/zap#Logger.Fatal 

We can re-write this without the error passing and fatal only instead.

cadctl/config/config.go
```diff
@@ -0,0 +1,157 @@
+/*
+Copyright © 2022 Red Hat, Inc.
```
typeid, line 2: hmm.. 2025? Or remove the licensing on the file as a whole?

cadctl/config/config.go
```diff
@@ -0,0 +1,157 @@
...
+		return "", fmt.Errorf("client-secret flag or CAD_OCM_CLIENT_SECRET environment variable must be set")
+	}
+
+	c.CadOcmClientSecret = cadOcmClientSecret
+	return cadOcmClientSecret, nil
+}
+
+// GetCadOcmURL retrieves the OCM URL passed as an environment
+// CAD_OCM_URL variable.
+func (c *config) GetCadOcmURL(cmd *cobra.Command) (string, error) {
```
typeid, line 149: `cmd` is not used

cadctl/config/config.go
```diff
@@ -0,0 +1,157 @@
...
+		return "", fmt.Errorf("client-id flag or CAD_OCM_CLIENT_ID environment variable must be set")
+	}
+
+	c.CadOcmClientID = cadOcmClientID
+	return cadOcmClientID, nil
+}
+
+// GetCadOcmClientSecret retrieves the OCM client secret passed as
+// an environment CAD_OCM_CLIENT_SECRET variable.
+func (c *config) GetCadOcmClientSecret(cmd *cobra.Command) (string, error) {
```
typeid, line 137: `cmd` is not used

cadctl/config/config.go
```diff
@@ -0,0 +1,157 @@
...
+		return experimentalFlag, nil
+	}
+
+	c.Experimental = false
+	return experimentalFlag, nil
+}
+
+// GetCadOcmClientID retrieves the OCM cliend ID passed as
+// an CAD_OCM_CLIENT_ID environment variable.
+func (c *config) GetCadOcmClientID(cmd *cobra.Command) (string, error) {
```
typeid, line 125: `cmd` is not used

cadctl/config/config.go
```diff
@@ -0,0 +1,157 @@
...
+
+	c.LogLevel = "info"
+	return logLevel, nil
+}
+
+// GetExperimental retrieves the experimental flag passed as a command line argument.
+// If not set, it checks the CAD_EXPERIMENTAL_ENABLED environment variable.
+// Defaults to false if neither is set.
+func (c *config) GetExperimental(cmd *cobra.Command) (bool, error) {
+	experimentalFlag, err := cmd.Flags().GetBool("experimental")
```
typeid, line 104: **issue:** this would break if we run it, because `experimental` does not exist as flag.

typeid (changes requested):

cadctl/cmd/investigate/investigate.go
petrkotas: Yes, it only adds fields.

cadctl/config/config.go
petrkotas: ah I removed it 🤦 and than forgot.

cadctl/cmd/investigate/investigate.go
typeid: We're now initializing it twice before initializing it with the cluster_id, so a third time

cadctl/config/config.go
typeid: I think `GetLogLevel` has the same issue, somehow my comment was lost

cadctl/cmd/investigate/investigate.go
petrkotas: I have a followup on this that focuses entirely on logging.

cadctl/cmd/investigate/investigate.go
petrkotas: I am not sure. I do not want to make the configuration imported in pkg. This is the only place where I think this is going to be used. The other command that I will be adding will solely use the config file, no variables.

cadctl/cmd/root.go
petrkotas: I removed this part of code.

cadctl/cmd/root.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package cmd
 
 import (
+
+	// "github.com/openshift-online/ocm-cli/pkg/config"
```
RaphaelBut, line 21: likely a left over comment? :)

cadctl/cmd/investigate/investigate.go
```diff
@@ -109,8 +116,17 @@ func run(cmd *cobra.Command, _ []string) error {
...
+	var ocmErr error
+	// Initialize the OCM client
+	// If the config file is set, use it to initialize the OCM client
+	// use config file only when developing
+	if c.ConfigFile != "" {
+		ocmClient, ocmErr = ocm.NewFromConfig(c.ConfigFile)
+	} else {
+		ocmClient, ocmErr = ocm.NewFromClientKeyPair(c.CadOcmURL, c.CadOcmClientID, c.CadOcmClientSecret)
+	}
+	if ocmErr != nil {
```
typeid, line 129: **issue:** trying to run CAD locally with these code changes, it breaks for me.

The logic here seems to have changed changed, we're taking config file over the ENVs. The logic in ocm itself, CAD, and every other tool is to prioritize ENVs over the config file. 

E.g. AWS cli will take envs over your config file.

cadctl/config/config.go
```diff
@@ -0,0 +1,182 @@
...
+		return "", fmt.Errorf("client-secret flag or CAD_OCM_CLIENT_SECRET environment variable must be set")
+	}
+
+	c.CadOcmClientSecret = cadOcmClientSecret
+	return cadOcmClientSecret, nil
+}
+
+// getCadOcmURL retrieves the OCM URL passed as an environment
+// CAD_OCM_URL variable.
+func (c *Config) getCadOcmURL() (string, error) {
```
typeid, line 174: **thought:** All these getters are also setters. Would the following make more sense? 
```
func  getCadOcmURL() (string, error) {
	cadOcmURL, exist := os.LookupEnv("CAD_OCM_URL")
	if !exist {
		return "", fmt.Errorf("ocm-url flag or CAD_OCM_URL environment variable must be set")
	}

	return cadOcmURL, nil
}
```

```
func LoadConfig() {

c := &Config{}

c.CadOcmUrl, err := getCadOcmUrl()
...
...
```

interceptor/main.go
```diff
@@ -71,6 +72,17 @@ func main() {
 	logger.Infof("Server exiting")
 }
 
+func init() {
+	// read log level from environment variable
+	logLevel := os.Getenv("LOG_LEVEL")
+	if logLevel != "" {
+		logLevel = "info"
+	}
```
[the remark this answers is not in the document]
petrkotas: good point, this is how AI completion sneaks upon you. Thanks for spotting this.

cadctl/cmd/investigate/investigate.go
petrkotas: In case you are not running it on fresh environment, you are right. It will prefere the local config over the env variables. Which is not entirely the same thing as we want for the PD investigate functionality. 

I am removing this as of now and will add it to the second command that does the local investigation. That one will not use the environment variables but instead rely entirely on logged in ocm client.

interceptor/main.go
```diff
@@ -22,13 +23,23 @@ const (
 	idleTimeout  = 60 * time.Second
 )
 
-var logger = logging.InitLogger(logging.LogLevelString, "")
+var logger *zap.SugaredLogger
```
bergmannf, line 26: Does this logger actually work?
The examples from zap all initialize the logger like this:
```
logger, _ := zap.NewProduction()
defer logger.Sync() // flushes buffer, if any
sugar := logger.Sugar()
```

PR #447, opened 2025-05-17
Fix: pass AWS proxy explicitly

typeid:
Running CAD locally with the latest backplane-cli results in issues, as the AWS client will not run with the proxy.  The latest backplane-cli introduced enforcement on IPs that interact with the AWS account.

This PR does multiple things:
- it refactors the way we initialize the aws sdk, removes some duplicate code
- it adds the ENV variable to the manual development test script
- it explicitly initializes the AWS SDK with the proxy
- it refactors our network verifier usage to properly pass our config vs re-building and dropping some fields


pkg/networkverifier/networkverifier.go
```diff
@@ -97,8 +96,13 @@ func Run(cluster *v1.Cluster, clusterDeployment *hivev1.ClusterDeployment, awsCl
 		return Undefined, "", fmt.Errorf("failed to initialize validateEgressInput: %w", err)
 	}
 
-	credentials := awsClient.GetAWSCredentials()
-	awsVerifier, err := verifierAws.NewAwsVerifier(credentials.AccessKeyID, credentials.SecretAccessKey, credentials.SessionToken, cluster.Region().ID(), "", false)
```
typeid, line 101: We would drop the proxy config here as well..

pkg/networkverifier/networkverifier.go
```diff
@@ -97,8 +96,13 @@ func Run(cluster *v1.Cluster, clusterDeployment *hivev1.ClusterDeployment, awsCl
...
+	// The network verifier requires a very specific type of logger (ocm logger)
+	logger, err := ocmlog.NewStdLoggerBuilder().Build()
```
typeid, line 100: This might be worth a followup: we're not able to use our structured logger because of this.

typeid:
/unhold

Tested locally, should work on stage as well as the proxy is not set there, thus it's not passed either.

pkg/networkverifier/networkverifier.go
```diff
@@ -76,7 +76,6 @@ func InitializeValidateEgressInput(cluster *v1.Cluster, clusterDeployment *hivev
 	}
 
 	return &verifier.ValidateEgressInput{
-		Timeout:      5 * time.Second,
```
typeid, line 79: The upstream default is 5 sec as well, we ideally don't want to manually override this and inherit it from upstream directly.

pkg/networkverifier/networkverifier.go
joshbranham: This is on our radar, probably a roadmap improvement alongside passing a context from the caller

PR #460, opened 2025-05-28
SREP-207: Naming cad-interceptor-service service port so it can be referenced by the ServiceMonitor

Nikokolas3270:
Unlike what I thought: service monitors cannot handle the numeric value of a port.


rafael-azevedo (approved):
/lgtm

PR #476, opened 2025-06-09
OSD-30036: To create E2E Test for CAD - Cluster has gone missing - UpgradeConfigSyncFailureOver4Hr 

ratnam915:
https://github.com/openshift/configuration-anomaly-detection/pull/468

Continuing from above PR

With the addition of this test case, possible scenarios for E2E tests in CAD has been taken care of.

Changes from previous pull request, handled the dependency from the CAD_OCM_FILE_PATH variable and established the connection to OCM using OCM Adapter in utils.go.

Handled all the dependencies in making the OCM connection in the utils.go file.


test/e2e/configuration_anomaly_detection_test.go
```diff
@@ -672,4 +668,135 @@ var _ = Describe("Configuration Anomaly Detection", Ordered, func() {
...
+		// Get cluster information from OCM
+		response, err := ocme2eCli.ClustersMgmt().V1().Clusters().Cluster(clusterID).Get().Send()
+		Expect(err).ToNot(HaveOccurred(), "failed to get cluster from OCM")
+		cluster := response.Body()
+		Expect(cluster).ToNot(BeNil(), "received nil cluster from OCM")
+
+		lsResponseBefore, err := utils.GetLimitedSupportReasons(ocme2eCli, clusterID)
+		var lsReasonsBefore int
+		if err != nil {
+			ginkgo.GinkgoWriter.Printf("Could not get limited support reasons before test: %v\n", err)
```
bergmannf, line 681: If we can't compare to the LS reasons before the test, should we even run it or just [Skip](https://onsi.github.io/ginkgo/#skipping-specs) it in that case?

This would not fail the suite but still be reported as skipped at least.

test/e2e/configuration_anomaly_detection_test.go
```diff
@@ -672,4 +668,135 @@ var _ = Describe("Configuration Anomaly Detection", Ordered, func() {
...
+		}
+
+		// Check if user is banned (part of the investigation logic)
+		ginkgo.GinkgoWriter.Printf("Checking if cluster owner is banned...\n")
+		userBannedStatus, userBannedNotes, err := utils.IsUserBanned(ocme2eCli, cluster)
+		Expect(err).NotTo(HaveOccurred(), "Failed to check if user is banned")
+
+		if userBannedStatus {
+			ginkgo.GinkgoWriter.Printf("User is banned: %s\n", userBannedNotes)
+			_, err = testPdClient.TriggerIncident("UpgradeConfigSyncFailureOver4HrSRE", clusterID)
```
bergmannf, line 695: If the user is banned this triggers and incident and later on triggers the same incident again.

test/e2e/configuration_anomaly_detection_test.go
```diff
@@ -656,8 +650,10 @@ var _ = Describe("Configuration Anomaly Detection", Ordered, func() {
 			_, err = testPdClient.TriggerIncident("InsightsOperatorDown", clusterID)
 			Expect(err).NotTo(HaveOccurred(), "Failed to trigger silent PagerDuty alert")
 
+			time.Sleep(2 * time.Minute)
```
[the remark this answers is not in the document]
ratnam915: Hey @bergmannf : For this we need a get method to get the information from PagerDuty and as discussed since it is not possible, this is the best way to make sure the incident is created and CAD has enough time to react

test/e2e/configuration_anomaly_detection_test.go
RaphaelBut: Should we mark the test failed if we cannot get the limited support reason before?

PR #481, opened 2025-06-16
Fix validate PR check

MateSaary:
### What type of PR is this?

This PR should fix the validate PR check by avoiding erroring out due to no results from a grep operation.

### DIFF
--- a/hack/investigation-test-coverage.sh
+++ b/hack/investigation-test-coverage.sh
@@ -11,7 +11,7 @@ BASE_SHA=$(git ls-remote "https://github.com/openshift/configuration-anomaly-det
 PR_SHA=$(git rev-parse HEAD)
 
 # Obtaining instances of files added in the PR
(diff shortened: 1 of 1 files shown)


hack/investigation-test-coverage.sh
```diff
@@ -11,7 +11,7 @@ BASE_SHA=$(git ls-remote "https://github.com/openshift/configuration-anomaly-det
 PR_SHA=$(git rev-parse HEAD)
 
 # Obtaining instances of files added in the PR
-diff_files=$(git diff --name-status "$BASE_SHA" "$PR_SHA" | grep '^A' | awk '{print $2}')
+diff_files=$(git diff --name-status "$BASE_SHA" "$PR_SHA" | (grep '^A' || true;) | awk '{print $2}')
```
bergmannf, line 14:
```suggestion
diff_files=$(git diff --name-status "$BASE_SHA" "$PR_SHA" | { grep '^A' || true; } | awk '{print $2}')
```

PR #492, opened 2025-06-23
Remove not used to debug option

petrkotas:
### What type of PR is this?

Removes not used DEBUG options

Removes unused code.

- E2E testing is desired for actioning investigations. See README for more info on investigation graduation process.

#### Test coverage checks
- [ ] Added tests

- [ ] Ran unit tests locally

### DIFF
--- a/cadctl/cmd/investigate/investigate.go
+++ b/cadctl/cmd/investigate/investigate.go
@@ -19,7 +19,6 @@ package investigate
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -111,7 +110,7 @@ func run(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	ocmClient, err := GetOCMClient()
+	ocmClient, err := ocm.New()
 	if err != nil {
 		return fmt.Errorf("could not initialize ocm client: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


cadctl/cmd/investigate/investigate.go
```diff
@@ -193,18 +195,7 @@ func handleCADFailure(err error, resources *investigation.Resources, pdClient *p
 
 // GetOCMClient will retrieve the OcmClient from the 'ocm' package
 func GetOCMClient() (*ocm.SdkClient, error) {
```
typeid, line 197: Do we even need this function? Can we fully remove it?

cadctl/cmd/investigate/investigate.go
```diff
@@ -111,7 +110,7 @@ func run(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	ocmClient, err := GetOCMClient()
+	ocmClient, err := ocm.New("")
```
bergmannf, line 113: Wouldn't it make sense to remove the string argument to `New` given that we aren't using it anymore?

petrkotas: It would, but I would rather do it in followup, when I clean up the client section. I prefer to keep this as is to make the changes limited.

PR #511, opened 2025-07-14
SREP-1011:  Fixing dependency on OCM TOKEN for CAD E2E Tests

ratnam915:
Similar to changes done to the managed node metadata operator since the OCM Token was deprecated:

https://github.com/openshift/managed-node-metadata-operator/commit/0359629ca869acb1d411c7dc22fa3b9d9f5069d8

Changes have been made to handle the same for the CAD E2E Tests


bergmannf:
MNMO also updated the osde2e-common dependency in that PR - CAD is at `github.com/openshift/osde2e-common v0.0.0-20250409072139-f917c50a4bd2` right now  while MNMO is at `github.com/openshift/osde2e-common v0.0.0-20250703192222-36c5869579b1` - I'd say let's update that as well so we don't have a slightly too old version to work.

PR #522, opened 2025-07-31
Bump go mod dependencies

MateSaary:
### What type of PR is this?

maintenance

#### Test coverage checks
- [ ] Added tests

- [ ] Ran unit tests locally


pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go
```diff
@@ -121,7 +121,7 @@ func (i *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 
 	// Summarize recommendations from investigation in PD notes, if any found
 	if len(i.recommendations) > 0 {
-		i.notes.AppendWarning(i.recommendations.summarize())
+		i.notes.AppendWarning("%s", i.recommendations.summarize())
```
typeid, line 124: Interesting change, was this proposed by a linter? I guess it makes things more safe in case the summarized string contains something like `%s`.

MateSaary: It actually seems to come from the `test-cadtl` make target, which is a simple `go test` run. If I go inline, I do see a warning (not error) which seems to stem from `go vet` specifically: `govet: printf: non-constant format string in call to ...` 

From some digging, I was able to find the exact release note (so this is caused by moving fro Go 1.23 to 1.24): https://tip.golang.org/doc/go1.24#vet

I also saved the error incase any questions came up for it, here it is :slightly_smiling_face: : 
```
# [github.com/openshift/configuration-anomaly-detection/pkg/investigations/machinehealthcheckunterminatedshortcircuitsre]
pkg/investigations/machinehealthcheckunterminatedshortcircuitsre/machinehealthcheckunterminatedshortcircuitsre.go:124:25: non-constant format string in call to (*github.com/openshift/configuration-anomaly-detection/pkg/notewriter.NoteWriter).AppendWarning
```

PR #541, opened 2025-08-11
Refactor investigation interface

RaphaelBut:
Replaces the function ShouldInvestigateAlert with a simple get method for the alert title, simplifying the alert mapping in registry.go

CCAM stops to adhere to the investigation interface, as it is not an "alert" investigation and more similiar to the precheck example. This does not change any logic.

### What type of PR is this?

refactor

Prepare for new test cases in the e2e tests

#### Test coverage checks
- [ ] Added tests
- [X] This PR may not need unit tests

- [X] Ran unit tests locally
- [X] Validated the changes in a cluster

Very slowly vibed with claude


pkg/investigations/registry.go
```diff
@@ -30,9 +30,19 @@ var availableInvestigations = []investigation.Investigation{
...
-		if inv.ShouldInvestigateAlert(title) && (experimental || !inv.IsExperimental()) {
+		if strings.Contains(title, inv.AlertTitle()) && (experimental || !inv.IsExperimental()) {
 			return inv
 		}
 	}
 	return nil
 }
+
+// GetAvailableInvestigationsTitles returns a string array with the alert titles of all available investigations.
+func GetAvailableInvestigationsTitles() []string {
```
bergmannf: This seems to be dead code?
I don't see this being used anywhere.

RaphaelBut: Good catch, I added that to facilitate the new e2e test I want to add in https://github.com/openshift/configuration-anomaly-detection/pull/542
Arguably it should be moved to the other PR, but if okay for you, I will keep it in this one 

 https://github.com/openshift/configuration-anomaly-detection/pull/542/files#diff-b41f8eaba3f873a581795129bfc035cd80d855da0564a2016c07aa0e3a16c439R88

PR #606, opened 2025-11-04
Revert "Fix: nilptr deref in ccam.go"

RaphaelBut:
Reverts openshift/configuration-anomaly-detection#597

WithAwsClient includes WithCluster https://github.com/openshift/configuration-anomaly-detection/blob/main/pkg/investigations/investigation/investigation.go#L111


MateSaary (approved):
/lgtm

PR #618, opened 2025-11-11
[SREP-2432] Add executor module

bergmannf:
### What type of PR is this?

feature

This PR sets the groundwork to split the investigation and execution of resulting action into two separate modules.
Instead of having investigations do everything, they should just report correct actions based on findings which can then be performed at a later stage.
At this  later stage it could also be decided if all or only some should be done / be done in parallel.

This is a very WIP PR and as an example it's moving the CHGM investigation to the new architecture.

#### Test coverage checks
- [ ] Added tests

- [ ] Ran unit tests locally

### DIFF
--- a/pkg/investigations/investigation/investigation.go
+++ b/pkg/investigations/investigation/investigation.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/client-go/rest"
 
 	"github.com/openshift/configuration-anomaly-detection/pkg/aws"
+	"github.com/openshift/configuration-anomaly-detection/pkg/backplane"
 	k8sclient "github.com/openshift/configuration-anomaly-detection/pkg/k8s"
 	"github.com/openshift/configuration-anomaly-detection/pkg/logging"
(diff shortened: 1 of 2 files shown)


pkg/investigations/investigation/investigation.go
```diff
@@ -20,6 +21,10 @@ type InvestigationStep struct {
 }
 
 type InvestigationResult struct {
+	// NEW: Actions to execute via reporter (modern approach)
+	Actions []types.Action
+
+	// EXISTING: Legacy fields (deprecated, maintained for backwards compatibility)
 	LimitedSupportSet  InvestigationStep
 	ServiceLogPrepared InvestigationStep
```
RaphaelBut: thought: do we keep something equivalent to servicelog prepared? 

I think we could formalize the "informing" state of an investigation pretty easily after this change. 
If the investigation is marked as informing, the actions are not executed but only posted to pd?  
And then we wont need the servicelog perpared anymore

PR #620, opened 2025-11-17
SREP-1881 add oc to cad

rolandmkunkel:
### What type of PR is this?

feature

https://issues.redhat.com/browse/SREP-1881
This adds the OC binary to the CAD image and a wrapper to make oc available from go code. It also refactors how we create and manage connections to customer clusters. Previously, the pkg/k8sclient package was responsible for creating backplane remediations and the Kubernetes rest.Config. This created tight coupling, relied on global state (backplaneURL), and made the system difficult to test and extend.

This change decouples these concerns by moving the responsibility for resource creation and lifecycle management into the investigation.ResourceBuilder. The builder is now the single source for creating a rest.Config which can then be shared by any client that needs it (e.g., k8s.Client, oc.Client).

### DIFF
--- a/pkg/investigations/investigation/errors.go
+++ b/pkg/investigations/investigation/errors.go
@@ -27,10 +27,32 @@ type AWSClientError struct {
 	Err       error
 }
 
-func (a *AWSClientError) Unwrap() error { return a.Err }
+func (e AWSClientError) Unwrap() error { return e.Err }
 
 func (e AWSClientError) Error() string {
 	return fmt.Sprintf("could not retrieve aws credentials for %s: %s", e.ClusterID, e.Err.Error())
+}
+
+type RestConfigError struct {
+	ClusterID string
+	Err       error
+}
+
+func (e RestConfigError) Unwrap() error { return e.Err }
+
+func (e RestConfigError) Error() string {
+	return fmt.Sprintf("could not create rest config for %s: %s", e.ClusterID, e.Err.Error())
+}
+
+type OCClientError struct {
+	ClusterID string
+	Err       error
+}
+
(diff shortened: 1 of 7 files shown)


pkg/investigations/investigation/errors.go
```diff
@@ -33,6 +33,28 @@ func (e AWSClientError) Error() string {
...
+}
+
+type OCClientError struct {
+	ClusterID string
+	Err       error
+}
+
+func (e *OCClientError) Unwrap() error { return e.Err }
+
+func (e OCClientError) Error() string {
```
joshbranham: nit: use pointer receivers for all functions
```suggestion
func (e * OCClientError) Error() string {
```

pkg/investigations/investigation/investigation.go
```diff
@@ -72,17 +81,21 @@ type Resources struct {
 	ClusterDeployment *hivev1.ClusterDeployment
 	AwsClient         aws.Client
 	BpClient          backplane.Client
+	RestConfig        *Config
```
joshbranham: This is one of the cases where a package alias I think makes sense. We have lots of "config" things here and hard to know what this is referring to as is

pkg/oc/oc.go
```diff
@@ -0,0 +1,154 @@
...
+	Clean() error
+}
+
+type clientImpl struct {
+	kubeConfigFile        string
+	cleanupKubeConfigFile func() error
+}
+
+// NewClient creates a new OC Client with the given rest.Config.
+func New(ctx context.Context, config *rest.Config) (Client, error) {
```
joshbranham: Func name is `New` not `NewClient`

PR #629, opened 2025-12-02
SREP-2648: Add organization-based escalation policy routing

rafael-azevedo:
### What type of PR is this?

feature

This PR implements organization-based escalation policy routing for CAD alerts. When alerts fire for clusters that belong to specific organizations (e.g., sovereign customers with dedicated support requirements), CAD now automatically reassigns those incidents to organization-specific escalation policies before running investigations. This ensures alerts are routed to the appropriate dedicated on-call teams rather than the general SRE pool.

Changes:
  - Exported MoveToEscalationPolicy() method in PagerDuty client for interceptor use
  - Added GetOrganizationID() helper to OCM client to check cluster organization membership
  - Implemented org-to-policy mapping logic in interceptor that loads from CAD_ORG_POLICY_MAPPING env var
  - Added test fixture with example mapping structure
  - Reassignment happens before investigation check, ensuring all subsequent escalations occur within the
  correct policy

  The mapping will be managed in a private repo as a Kubernetes secret and injected as an env var. The
  interceptor now requires OCM credentials (CAD_OCM_CLIENT_ID, CAD_OCM_CLIENT_SECRET, CAD_OCM_URL) which
  need to be added to the deployment.

  If reassignment fails, the alert continues processing but a warning note is added to the PagerDuty
  incident alerting SRE to manually route.

- [ ] Ran unit tests locally

TODO:
   - Finish local testing with mock PagerDuty webhooks and OCM data
  - Create sample mapping in staging and test end-to-end
  - Document deployment configuration for private repo (OCM creds, secret structure, env var setup)
  - Add documentation for onboarding new organizations to dedicated flow

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,38 @@ Grafana dashboard configmaps are stored in the [Dashboards](./dashboards/) direc
 
 - `CAD_EXPERIMENTAL_ENABLED`: enables experimental investigations when set to `true`, see mapping.go
 
+- `CAD_ORG_POLICY_MAPPING`: JSON configuration for organization-based escalation policy routing. When configured, the interceptor automatically reassigns PagerDuty incidents for clusters belonging to specific organizations to dedicated escalation policies. This enables organization-specific on-call rotations.
+
+  Example configuration:
+  ```json
+  {
+    "organizations": [
+      {
+        "name": "Customer Alpha",
+        "org_ids": ["org-id-1", "org-id-2"],
+        "escalation_policy": "P123ABC"
+      },
+      {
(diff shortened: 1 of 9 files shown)


bergmannf:
I would love for the expected format of the environment variable to be documented with a simple example (highlighting all fields required).
(Also as this is expected to be JSON I think putting it in a config-file would be nicer, but CAD currently doesn't have one, so I would not expect this PR to introduce one - adding a // TODO or // FIXME would be nice though so we won't forget about it when we add it)

PR #644, opened 2025-12-18
Update documentation for escalation policy requirements

rafael-azevedo:
- Clarify that CAD_ORG_POLICY_MAPPING is set as ENV variable in deployment via app-interface saas file (not Kubernetes Secret)
- Document that escalation policies must have at least 2 levels (CAD escalates to level 2)
- Apply same requirement to CAD_SILENT_POLICY


joshbranham (commented):
/lgtm

PR #653, opened 2026-01-23
[SREP-3107] Split investigation logic into controllers.

bergmannf:
This allows adding new CLI commands to distinguish between a PD triggered run and a manually triggered run.
### What type of PR is this?

Feature

The PD run will instantiate a PagerDutyController that will pass a PD client to the investigations.

The manual run will use a ManualController and *not* provide a PD client as there might be no alert associated with it in that case.

#### Test coverage checks
- [ ] Added tests

- [X] Ran unit tests locally

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -319,7 +319,6 @@ func (c *investigationRunner) runInvestigation(ctx context.Context, clusterId st
 
 	logging.Infof("Starting investigation for %s", inv.Name())
 	result, attempts, err := runInvestigationWithRetry(inv, builder)
-	// result, err = inv.Run(builder)
 	if err != nil {
 		return fmt.Errorf("investigation failed after %d attempt(s): %w", attempts, err)
 	}
--- a/pkg/controller/manual.go
+++ b/pkg/controller/manual.go
@@ -5,9 +5,26 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/openshift/configuration-anomaly-detection/pkg/investigations"
+	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/investigation"
 )
+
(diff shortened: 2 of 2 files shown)


pkg/controller/manual.go
```diff
@@ -0,0 +1,33 @@
...
+type ManualController struct {
+	config CommonConfig
+	manual ManualConfig
+	investigationRunner
+}
+
+func (c *ManualController) Investigate(ctx context.Context) error {
+	experimentalEnabledVar := os.Getenv("CAD_EXPERIMENTAL_ENABLED")
+	experimentalEnabled, _ := strconv.ParseBool(experimentalEnabledVar)
+	alertInvestigation := investigations.GetInvestigationByName(c.manual.InvestigationName, experimentalEnabled)
```
rolandmkunkel: I think we should tweak some of the names, like the CHGM one, otherwise this will be somewhat ugly to call from CLI

bergmannf: We kind of can not change the names, because of:
```
	// Please note that when adding an investigation the name and the directory currently need to be the same,
	// so that backplane-api can fetch the metadata.yaml
```
But I can add a shortname field and update the logic how we can pass the argument.

pkg/controller/controller.go
```diff
@@ -0,0 +1,485 @@
...
+				logger:       logger,
+				dependencies: deps,
+			},
+		}, nil
+	}
+
+	return nil, fmt.Errorf("no valid controller configuration provided")
+}
+
+func (c *investigationRunner) runInvestigation(ctx context.Context, clusterId string, inv investigation.Investigation, pdClient *pagerduty.SdkClient) error {
```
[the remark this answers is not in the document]
bergmannf: In theory we should handle cancellation (especially for the manual flow), but don't right now.
But I guess we should keep it.

PR #763, opened 2026-04-09
Fix e2e test yaml

bergmannf:
### What type of PR is this?

bug

Revert previous fixed to use the correct way to pass the ROUTING_KEY

#### Test coverage checks
- [ ] Added tests

- [ ] Ran unit tests locally


ritmun:
Could you run a bp-update and commit a clean e2e template? It should load any custom secrets.

bergmannf:
Replaced by #764

PR #776, opened 2026-04-21
[SREP-4025] Configuration file for CAD to allow filtering investigation run

bergmannf:
### What type of PR is this?

feature

#### Test coverage checks
- [X] Added tests

- [X] Ran unit tests locally

## Summary by CodeRabbit

* **New Features**
  * Added YAML-based investigation filter configuration system with operators (`in`, `notin`, `matches`, `sample`) to control which investigations run for specific clusters and organizations.
  * Introduced CLI flags `--config` and `--with-filtering` to manage investigation filtering and configuration paths.
  * AI-assisted investigation is now governed by filter configuration instead of allowlists.

* **Documentation**
  * Added example filter configuration YAML documenting schema and usage patterns.


pkg/config/filter.go
```diff
@@ -0,0 +1,379 @@
...
+
+// evaluateLeaf evaluates a single leaf node against the FilterContext.
+func (n *FilterNode) evaluateLeaf(ctx *types.FilterContext) (bool, error) {
+	switch n.Operator {
+	case OperatorSample:
+		rate, err := strconv.ParseFloat(n.Values[0], 64)
+		if err != nil {
+			return false, fmt.Errorf("sample: invalid rate %q: %w", n.Values[0], err)
+		}
+		return rand.Float64() < rate, nil //nolint:gosec // not security-sensitive, used for traffic sampling
```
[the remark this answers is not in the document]
bergmannf: sample is only called once now in the newest code so it can stay actually random and not lead to being called multiple times.
the interceptor no longer evaluates the filter.

docs/investigation-filter-config.example.yaml
```diff
@@ -0,0 +1,188 @@
...
+  #         operator: in
+  #         values: ["cluster-id-1", "cluster-id-2"]
+  #       - field: OrganizationID
+  #         operator: in
+  #         values: ["org-id-1", "org-id-2"]
+
+  # Only run the mustgather investigation on AWS clusters that are in "ready" state.
+  # Uses AND logic: both conditions must pass.
+  - investigation: mustgather
+    exclude:
```
rolandmkunkel: The key "exclude" threw me a bit off. Seeing this could be interpreted as "exclude this investigation when this condition matches"

bergmannf: You are right 'exclude' actually indicates the exact opposite of what happens.
I'll change it to 'when' - that's what Tekton uses to indicate 'only run when' which should hopefully be clear enough.

pkg/config/config.go
```diff
@@ -0,0 +1,278 @@
...
+
+// LoadConfig reads and parses the investigation filter configuration.
+// If pathOverride is non-empty, it is used as the config file path.
+// Otherwise, the path is read from the CAD_INVESTIGATION_CONFIG_PATH environment variable.
+// If neither is set, falls back to the legacy CAD_AI_AGENT_CONFIG JSON env var for
+// backwards compatibility.
+// Returns nil (no config) if no source is available.
+// The validInvestigations parameter is the list of known investigation names used to
+// validate that each filter references a real investigation.
+func LoadConfig(pathOverride string, validInvestigations []string) (*Config, error) {
```
rolandmkunkel: nit: Should we name this LoadAndValidateConfig so that its clear why we are passing the valid investigations here as well?

pkg/config/filter.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+	if ctx == nil {
+		return true, "no filter context (manual mode)", nil
+	}
+	return f.Filter.evaluate(ctx)
+}
+
+// Keys returns all field names referenced by leaf nodes in the filter tree.
+// Used to determine which FilterContext fields need to be populated.
+func (f InvestigationFilter) Keys() []string {
```
rolandmkunkel, line 116: This could also be a pointer receiver (go standard)

pkg/config/config.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	seen := make(map[string]bool)
+
+	for i, f := range c.Filters {
+		if f.Investigation == "" {
+			return fmt.Errorf("filters[%d]: investigation name must not be empty", i)
+		}
+
+		if seen[f.Investigation] {
```
rolandmkunkel, line 225: nit: I would slightly prefer:
`if _, ok := seen[f.Investigation]; ok {`

rolandmkunkel (approved):
only some nits. great work

PR #798, opened 2026-04-29
SREP-3858: Reimplement cluster-health-check managed script as a CAD action

MateSaary:
### What type of PR is this?

feat

Implements a cluster-health-check action in CAD that replicates the functionality of the managed-scripts health/cluster-health-check (https://github.com/openshift/managed-scripts/tree/main/scripts/health/cluster-health-check).

- The firing alerts check uses pods/exec in openshift-monitoring to query the Alertmanager API via wget http://localhost:9093/api/v2/alerts inside the alertmanager pod. The managed-scripts version uses a route + OAuth proxy approach, but that seems to fail in-practice.

#### Test coverage checks
- [x] Added tests

- [x] Ran unit tests locally
- [x] Validated the changes in a cluster

## Summary by CodeRabbit

* **New Features**
  * Added an experimental cluster health check investigation to diagnose API server health, operators/pods/nodes, CSRs, MachineConfigPools, alerts, events, capacity/utilization and cluster version/EOL status.
  * Includes metadata/RBAC to allow the check to access required cluster resources.

* **Bug Fixes**
  * Manual investigation short name now recognizes the "cluster-health-check" alias.

* **Tests**
  * Added comprehensive unit tests for the new investigation.

* **Documentation**
  * Added README documenting the cluster health check and deviations from existing implementations.

### DIFF
--- a/pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
+++ b/pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
@@ -1,0 +1,1168 @@
+// Package clusterhealthcheck implements a CAD investigation that runs
+// comprehensive health checks against an OpenShift cluster, replicating the
+// functionality of the managed-scripts health/cluster-health-check action.
+package clusterhealthcheck
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"math"
+	"net/http"
+	"sort"
+	"strings"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	mcfgv1 "github.com/openshift/api/machineconfiguration/v1"
+
+	"github.com/openshift/configuration-anomaly-detection/pkg/executor"
+	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/investigation"
+	k8sclient "github.com/openshift/configuration-anomaly-detection/pkg/k8s"
+	"github.com/openshift/configuration-anomaly-detection/pkg/logging"
+	"github.com/openshift/configuration-anomaly-detection/pkg/notewriter"
+	"github.com/openshift/configuration-anomaly-detection/pkg/types"
+
+	certsv1 "k8s.io/api/certificates/v1"
+	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type alertsFetcher interface {
+	fetchFiringAlerts(ctx context.Context, k8sClient k8sclient.Client, restConfig *rest.Config) ([]firingAlert, error)
+}
+
+type etcdHealthChecker interface {
+	checkEtcdHealth(ctx context.Context, k8sClient k8sclient.Client, restConfig *rest.Config, namespace string) (string, error)
+}
+
+type apiHealthChecker interface {
+	checkHealth(ctx context.Context, restConfig *rest.Config) (apiHealthResult, error)
+}
+
+type eolChecker interface {
+	checkEOL(ctx context.Context, channel string, currentMinor int) (latestMinor int, err error)
+}
+
+type apiHealthResult struct {
+	healthz string
+	livez   string
+	readyz  string
+}
+
+type firingAlert struct {
+	Name     string
+	Severity string
+	State    string
+	Summary  string
+}
+
+type Investigation struct {
+	alertsFetcher    alertsFetcher
+	etcdChecker      etcdHealthChecker
+	apiHealthChecker apiHealthChecker
+	eolChecker       eolChecker
+}
+
+func (i *Investigation) Run(rb investigation.ResourceBuilder) (investigation.InvestigationResult, error) {
+	ctx := context.Background()
+	result := investigation.InvestigationResult{}
+
+	r, err := rb.WithCluster().WithK8sClient().Build()
+	if err != nil {
+		if msg, ok := investigation.ClusterAccessErrorMessage(err); ok {
+			logging.Warnf("Cluster access error for cluster-health-check: %v", err)
+			result.Actions = []types.Action{
+				executor.Note(msg),
+			}
+			return result, nil
+		}
+		return result, investigation.WrapInfrastructure(err, "failed to build resources for cluster-health-check")
+	}
+
+	notes := notewriter.New(i.Name(), logging.RawLogger)
+
+	if r.IsHCP {
+		notes.AppendWarning("This is an HCP cluster - some checks may differ (control plane runs on management cluster)")
+	}
+
+	// run all 12 health checks, collecting results.
+	i.checkClusterOperators(ctx, r.K8sClient, notes)
+	i.checkAPIServerHealth(ctx, r, notes)
+	i.checkEtcdStatus(ctx, r, notes)
+	i.checkMachineConfigPools(ctx, r.K8sClient, r.IsHCP, notes)
+	i.checkPendingCSRs(ctx, r.K8sClient, notes)
+	i.checkNodeStatus(ctx, r.K8sClient, notes)
+	i.checkCapacity(ctx, r, notes)
+	i.checkFiringAlerts(ctx, r, notes)
+	i.checkClusterVersion(ctx, r.K8sClient, notes)
+	i.checkFailingPods(ctx, r.K8sClient, notes)
+	i.checkRestrictivePDBs(ctx, r.K8sClient, notes)
+	i.checkNonNormalEvents(ctx, r.K8sClient, notes)
+
+	result.Actions = executor.NoteAndReportFrom(notes, r.Cluster.ID(), i.Name())
+	return result, nil
+}
+
+// checkClusterOperators lists all ClusterOperators and reports any that are degraded, unavailable, or progressing.
+func (i *Investigation) checkClusterOperators(ctx context.Context, k8sClient k8sclient.Client, notes *notewriter.NoteWriter) {
+	coList := &configv1.ClusterOperatorList{}
+	if err := k8sClient.List(ctx, coList); err != nil {
+		notes.AppendWarning("Cluster Operators: failed to list - %v", err)
+		return
+	}
+
+	if len(coList.Items) == 0 {
+		notes.AppendWarning("Cluster Operators: none found")
+		return
+	}
+
+	var degraded, unavailable, progressing []string
+
+	for _, co := range coList.Items {
+		for _, cond := range co.Status.Conditions {
+			switch {
+			case cond.Type == configv1.OperatorDegraded && cond.Status == configv1.ConditionTrue:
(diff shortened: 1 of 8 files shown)


pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
```diff
@@ -0,0 +1,1012 @@
...
+
+	if len(overThreshold) > 0 {
+		notes.AppendWarning("Capacity Utilization: %d node(s) with >=80%% utilization - %s", len(overThreshold), strings.Join(overThreshold, "; "))
+	} else {
+		notes.AppendSuccess("Capacity Utilization: all nodes have less than 80%% CPU and Memory utilization")
+	}
+	notes.AppendAutomation("Capacity per-node utilization:\n  %s", strings.Join(perNodeUtil, "\n  "))
+}
+
+func formatBytes(bytes int64) string {
```
rolandmkunkel, line 657: Seems like this is not used. Was this meant for the capacity output?

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
```diff
@@ -0,0 +1,1012 @@
...
+		reason := ""
+
+		if pdb.Spec.MaxUnavailable != nil && pdb.Spec.MaxUnavailable.IntValue() == 0 {
+			isRestrictive = true
+			reason = "maxUnavailable=0"
+		}
+
+		// Check minAvailable = 100% or equal to total expected pods
+		if pdb.Spec.MinAvailable != nil {
+			if pdb.Spec.MinAvailable.Type == 1 {
```
rolandmkunkel: Can we use intstr.String here instead of the magic number?

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
```diff
@@ -0,0 +1,1012 @@
...
+		result.readyz = fmt.Sprintf("error: %v", err)
+	} else {
+		result.readyz = string(body)
+	}
+
+	return result, nil
+}
+
+// checkEtcdStatus execs etcdctl in an etcd pod to verify cluster health. Skipped on HCP clusters.
+func (i *Investigation) checkEtcdStatus(ctx context.Context, r *investigation.Resources, notes *notewriter.NoteWriter) {
```
rolandmkunkel, line 288: This is performing different checks than https://github.com/givaldolins/openshift-cluster-health-check/blob/main/oc-hc/cmd/etcdStatus.go
These may not behave the same way in some edge cases. Are we fine with this? I think that should be documented in the ticket. Same with API health check, PDB, alerts, events and EOL

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
```diff
@@ -0,0 +1,1012 @@
...
+		checker = &defaultEtcdHealthChecker{}
+	}
+
+	restConfig, err := getRestConfig(r)
+	if err != nil {
+		notes.AppendWarning("ETCD Status: unable to get REST config - %v", err)
+		return
+	}
+
+	output, err := checker.checkEtcdHealth(ctx, r.K8sClient, restConfig, "openshift-etcd")
```
rolandmkunkel, line 305: Will this return an error if  `etcdctl endpoint health --cluster` successfully returns, but with a "cluster is unhealthy"? If yes, we return success on an unhealthy cluster below

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
```diff
@@ -0,0 +1,1012 @@
...
+		if len(schedulingDisabled) > 0 {
+			notes.AppendWarning("Nodes: %d scheduling disabled - %s", len(schedulingDisabled), strings.Join(schedulingDisabled, ", "))
+		}
+		if len(conditionIssues) > 0 {
+			notes.AppendWarning("Node Conditions: %d issue(s) - %s", len(conditionIssues), strings.Join(conditionIssues, "; "))
+		}
+	}
+
+	if len(tainted) > 0 {
+		notes.AppendWarning("Node Taints: %d taint(s) found:\n  %s", len(tainted), strings.Join(tainted, "\n  "))
```
rolandmkunkel, line 512: This produces a warning on master nodes, since they always have 
```
 taints:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
```
Seems like the original implementation did that as well, though, so a NO-OP, still a bit odd

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
```diff
@@ -0,0 +1,1012 @@
...
+	for _, pdb := range pdbList.Items {
+		// (disruptionsAllowed == 0)
+		if pdb.Status.DisruptionsAllowed > 0 {
+			continue
+		}
+
+		isRestrictive := false
+		reason := ""
+
+		if pdb.Spec.MaxUnavailable != nil && pdb.Spec.MaxUnavailable.IntValue() == 0 {
```
rolandmkunkel: The original script also checks for `maxUnavail.StrVal == "0%"`
This code does actually catch it (because the conversion to int fails for 0% and returns 0), but I think that should be explicitly checked.

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
```diff
@@ -0,0 +1,1012 @@
...
+
+	if len(issues) == 0 {
+		notes.AppendSuccess("Cluster Version: %s", version)
+	} else {
+		notes.AppendWarning("Cluster Version: %s - %s", version, strings.Join(issues, "; "))
+	}
+}
+
+func checkEOL(cv *configv1.ClusterVersion) string {
+	desired := cv.Status.Desired.Version
```
rolandmkunkel: Isn't desired set by customers to trigger an update? In this case, we will only produce an EOL warning if the upgrade failed. The original script checked for the latest available version

pkg/investigations/clusterhealthcheck/clusterhealthcheck_test.go
```diff
@@ -0,0 +1,1392 @@
...
+	}
+
+	r := &investigation.Resources{
+		K8sClient: clientImpl{newFakeClient()},
+		IsHCP:     false,
+	}
+	inv.checkEtcdStatus(context.Background(), r, notes)
+
+	output := notes.String()
+	// Fake client won't have rest config, so error is from GetRestConfig, not from etcdChecker
```
rolandmkunkel: If this just tests that the missing rest config is handled gracefully, the test name is misleading

pkg/investigations/clusterhealthcheck/clusterhealthcheck_test.go
```diff
@@ -0,0 +1,1392 @@
...
+	r := &investigation.Resources{
+		K8sClient: clientImpl{newFakeClient()},
+		IsHCP:     false,
+	}
+	inv.checkEtcdStatus(context.Background(), r, notes)
+
+	output := notes.String()
+	// fake client doesn't have rest config, so it will warn
+	if !strings.Contains(output, "ETCD Status") {
+		t.Errorf("expected ETCD Status in output, got:\n%s", output)
```
rolandmkunkel: This seems to be missing assertions on correct behavior

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
MateSaary: Good point, definitely makes sense to at the very least document. As for the precise differences;
- etcd; this version would use `etcdctl` whereas the managedscript did it via `curl`. This should be more accurate, as it verifies the member can participate in quorum, not just that the process is alive. Though with 1 caveat, i.e. the MS one gave per-pod result. 
- Alerts: should net the same result, just using exec instead of route+oauth
- PDB: CAD should be checking more here
- Events: managedscript did `== "warning"`, whereas here we do `!= normal`... I guess we can expect this to less brittle at worst

Given changes from your comments, API-healthcheck and EOL should align now with original.

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
MateSaary: Should `etcdctl endpoint health --cluster` be returning successfully with "cluster is unhealthy"? From what I understand we should expect it to return non-zero in such a case

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
MateSaary: My guess is the original did it for verbosity to show all taints, question is whether we want to show that here as well :slightly_smiling_face:

MateSaary:
Thanks for the review :slightly_smiling_face: I went ahead and addressed the comments and made a few general improvements, let me know if there's anything residual to address.

pkg/investigations/clusterhealthcheck/clusterhealthcheck_test.go
```diff
@@ -0,0 +1,1548 @@
...
+	if !strings.Contains(output, "skipped") {
+		t.Errorf("expected skipped message for HCP, got:\n%s", output)
+	}
+}
+
+func TestCheckEtcdStatus_Error(t *testing.T) {
+	notes := newTestNotes()
+
+	inv := &Investigation{
+		etcdChecker: &mockEtcdHealthChecker{err: fmt.Errorf("etcd cluster is unhealthy")},
```
rolandmkunkel, line 1437: This mock error is never hit because the rest config is missing, so this is now the same test as TestCheckEtcdStatus_NoRestConfig. I think this should have a rest config

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
rolandmkunkel: In  that case, ExecInPod would return "", err in which case checkEtcdHealth would return "", err and throw away the output, meaning we can't distinguish between "check failed" and "cluster unhealthy".

We could always include the output even in error case and do something like

```
if err != nil { 
      if output != "" {
          notes.AppendWarning("ETCD Status:\n%s", output)                                                                                                                                                                                                                                                           
      } else {                                                                                                                                                                                                                                                                                                      
          notes.AppendWarning("ETCD Status: check failed - %v", err)                                                                                                                                                                                                                                                
      }                                                                                                                                                                                                                                                                                                             
      return                                                                                                                                                                                                                                                                                                        
  }          
```

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
rolandmkunkel: Should we document this in a readme?

MateSaary: Created a README

pkg/investigations/clusterhealthcheck/clusterhealthcheck.go
```diff
@@ -0,0 +1,1168 @@
...
+		checker = &defaultEtcdHealthChecker{}
+	}
+
+	restConfig, err := getRestConfig(r)
+	if err != nil {
+		notes.AppendWarning("ETCD Status: unable to get REST config - %v", err)
+		return
+	}
+
+	output, err := checker.checkEtcdHealth(ctx, r.K8sClient, restConfig, "openshift-etcd")
```
rolandmkunkel, line 305: checkEtcdHealth and ExecInPod both also discard the output, so they would need to return it, too. The test only works because the mock returns both which the real implementation does not

PR #809, opened 2026-05-05
Add metrics tracking for manually triggered investigations

petrkotas:
### What type of PR is this?

feature
jira: https://redhat.atlassian.net/browse/SREP-4672

Introduces two new Prometheus CounterVec metrics to track manual investigation lifecycle:

- cad_investigate_manual_started_total: Tracks when manual investigations are initiated via 'cadctl run' command, with labels for investigation name and dry-run mode

- cad_investigate_manual_completed_total: Tracks investigation outcomes with status labels (success, error, filtered, precheck_stopped, stop_investigations, ccam_stopped) and dry-run mode

The metrics are only incremented for manual investigations (when pdClient is nil) and are pushed to the Prometheus Pushgateway alongside existing metrics. This enables monitoring of manual investigation usage, success rates, and operational efficiency.

All completion paths in runInvestigation() are instrumented to ensure accurate tracking across success, error, and early-exit scenarios.

#### Test coverage checks
- [ ] Added tests

- [ ] Ran unit tests locally

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive metrics for manual investigations: records when manual runs start and when they complete, capturing outcomes (error, precheck stopped, stopped, filtered, CCAM stopped, success, no findings) and whether runs are dry-run.

* **Bug Fixes**
  * Improved and clarified error message for failures during PagerDuty title update action.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -263,6 +263,14 @@ func NewController(opts ControllerOptions, deps *Dependencies) (Controller, erro
 	return nil, fmt.Errorf("no valid controller configuration provided")
 }
 
+// formatBool converts boolean to string for metrics labels
+func formatBool(b bool) string {
+	if b {
+		return "true"
+	}
+	return "false"
+}
+
 func (c *investigationRunner) runInvestigation(ctx context.Context, clusterId string, inv investigation.Investigation, pdClient *pagerduty.SdkClient, filterCtx *types.FilterContext, params map[string]string) error {
 	metrics.Inc(metrics.Alerts, inv.Name())
 
@@ -299,6 +307,8 @@ func (c *investigationRunner) runInvestigation(ctx context.Context, clusterId st
 			}
 		}
 		if err != nil {
+			// Record error completion for manual investigations before handling failure
+			c.recordManualCompletion(inv.Name(), pdClient, "error")
 			handleCADFailure(err, builder, pdClient)
 		}
 	}()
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
rolandmkunkel, line 369: Error is redefined in block scope here, so its a different variable then in the defer block, meaning no metric would be recorded here, as the outer err would be nil. Same for the other locations where err := is used

pkg/controller/controller.go
```diff
@@ -263,6 +263,14 @@ func NewController(opts ControllerOptions, deps *Dependencies) (Controller, erro
 	return nil, fmt.Errorf("no valid controller configuration provided")
 }
 
+// formatBool converts boolean to string for metrics labels
+func formatBool(b bool) string {
```
rolandmkunkel: We can just use strconv.FormatBool(b) instead

pkg/controller/controller.go
```diff
@@ -313,16 +323,19 @@ func (c *investigationRunner) runInvestigation(ctx context.Context, clusterId st
 			return fmt.Errorf("failed to execute precheck actions: %w", err)
 		}
 		// We stop if the precheck returns any action this mean we do not want to run anything else.
+		c.recordManualCompletion(inv.Name(), pdClient, "precheck_stopped")
```
rolandmkunkel, line 318: There are more status fields used than what the ticket describes, and no "no findings" status. The ticket also asks for documentation update and a Grafana board. I assume these come with a later PR?

pkg/controller/controller.go
petrkotas: done

pkg/controller/controller.go
petrkotas: I added no findings, but the documentation and grafana board will come once some metrics are collected.

rolandmkunkel (approved):

PR #812, opened 2026-05-11
Update boilerplate

abyrne55:
Sync with latest boilerplate (61dbfdf)

## Summary by CodeRabbit

* **Chores**
  * Updated CI build root image version to image-v8.3.6
  * Enabled automated Docker dependency update checks on a weekly schedule
  * Updated team ownership configuration (alias adjustments)


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
```
RaphaelBut, line 1: Thanks for the PR! 
I did not realize the dependabot config is from boilerplate when i moved this logic to konflux's renovate. 

Looks like boilerplate now has both a renovate config and dependabot, which should result in duplicate PRs
https://github.com/openshift/boilerplate/blob/master/.github/renovate.json
https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/golang-osd-operator/dependabot.yml


https://github.com/openshift/configuration-anomaly-detection/blob/main/.github/renovate.json
Currently we have ^ config extending the one from boilerplate, which should make dependabot obsolete. If it does work as intended, we should probably move it to boilerplate and remove dependabot. ( it would require boilerplate consumers to be on konflux though ) 

We may want to hold this PR for a bit and figure out whats a good course of action. ( to fix cad, we can probably remove https://github.com/openshift/boilerplate/blob/master/boilerplate/openshift/osd-container-image/dependabot.yml from boilerplate as cad was the only consumer last time i checked :D

joshbranham: I think it's because this repo uses `boilerplate/openshift/osd-container-image` vs the operator one. Either way, +1 for using renovate with Konflux and removing the boilerplate config imo.

PR #818, opened 2026-05-15
Fix renovate indirect deps config

RaphaelBut:
### What type of PR is this?

bug

The Mintmaker preset (konflux-ci/mintmaker) enables all indirect Go
module updates via gomod.packageRules. 
https://github.com/konflux-ci/mintmaker/blob/main/config/renovate/renovate.json#L614

Our top-level packageRules disable rule for major indirect updates was being overridden because manager-level packageRules are applied after top-level ones.

Also remove the non-functional matchDepTypes: ["!indirect"] from the
major updates manual review rule — matchDepTypes does not support
negation, so that filter was silently matching nothing.

Also fixes the path for the boilerplate image to avoid updating and pinning it, see https://github.com/openshift/configuration-anomaly-detection/pull/816

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -9,6 +9,20 @@
     "gomod",
     "dockerfile"
   ],
+  "gomod": {
+    "packageRules": [
+      {
+        "description": "Disable major updates for indirect Go dependencies - they update when their parent dependency updates",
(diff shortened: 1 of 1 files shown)


MateSaary (approved):
/lgtm

PR #823, opened 2026-05-25
SREP-4911: Fix premature automation notes and distinguish uncertain infra cluster detection

MateSaary:
### What type of PR is this?

bug

#### Test coverage checks
- [x] Added tests

- [ ] Ran unit tests locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved infrastructure-cluster handling during API failures, including correct PagerDuty wording and escalation behavior.
  * Service log and limited-support actions now surface posting errors and add follow-up notes only after success.
  * Investigation notes now consistently use warning-style, customer-focused text (with automation wording removed).
* **Documentation**
  * Refreshed investigation-writing guidelines to reflect the updated note patterns.
* **Tests**
  * Added coverage for infra-cluster “uncertain” vs “detected” note content.

### DIFF
--- a/pkg/investigations/cpd/cpd.go
+++ b/pkg/investigations/cpd/cpd.go
@@ -100,11 +100,6 @@ func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 				notes.AppendWarning("subnet %s does not have a default route to 0.0.0.0/0", subnet)
 				byovpcRoutingSL := newBYOVPCRoutingSL(docLink)
 
-				// XXX: metrics.Inc(metrics.ServicelogSent, investigationName)
-				result.ServiceLogSent = investigation.InvestigationStep{Performed: true, Labels: nil}
-
-				notes.AppendAutomation("Sent SL: '%s'", byovpcRoutingSL.Summary)
-
 				result.Actions = append(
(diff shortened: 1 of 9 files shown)


pkg/investigations/cpd/cpd.go
```diff
@@ -103,8 +103,6 @@ func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 				// XXX: metrics.Inc(metrics.ServicelogSent, investigationName)
 				result.ServiceLogSent = investigation.InvestigationStep{Performed: true, Labels: nil}
```
rolandmkunkel: This has a similar issue as the service logs. The metric is increased even if no service log is sent afterwards.

MateSaary: Good catch, thanks. Changed it to `ServiceLogPrepared` :slightly_smiling_face:

rolandmkunkel: The fix has caused ServiceLogSent to never be sent. Ideally we would emit the metric in the executor. Although I'm happy to remove it if we don't actually care about it

MateSaary: I've removed the premature `ServiceLogSent` metric as well, let me know if I'm understanding your comment correctly.

PR #841, opened 2026-06-12
fix: use value comparison for route CIDR in IsSubnetPrivate

RaphaelBut:
[ROSAENG-59855](https://redhat.atlassian.net/browse/ROSAENG-59855)

IsSubnetPrivate compared route.DestinationCidrBlock (a *string) against awsv2.String("0.0.0.0/0") (a new *string) using ==, which compares pointers and never matches. This meant the IGW route was never detected, falling back to MapPublicIpOnLaunch which masked the bug in most real clusters but fails when a public subnet has MapPublicIpOnLaunch=false.

### What type of PR is this?

bug

[ROSAENG-59855]: https://redhat.atlassian.net/browse/ROSAENG-59855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -385,7 +385,7 @@ func (c *SdkClient) IsSubnetPrivate(subnet string) (bool, error) {
 	for _, route := range rtb.Routes {
 		// GatewayId can contain an internet gateway *or* a virtual private gateway:
 		// "The ID of an internet gateway or virtual private gateway attached to your VPC."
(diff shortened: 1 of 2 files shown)


petrkotas (approved):
The code is ok /lgtm

PR #849, opened 2026-06-19
ROSAENG-14171: add probe history and LB security group checks to console-ErrorBudgetBurn investigation

MateSaary:
### What type of PR is this?

feature

Adds two new informational checks to the consoleerrorbudgetburn investigation, automating manual SOP troubleshooting steps (sections 3.3 and 3.6-3.7 of console-ErrorBudgetBurn.md):
Probe History (classic only):
- Queries in-cluster Prometheus for 2 hours of historical probe_success range data via exec into the cluster-monitoring-operator pod
- Classifies the failure pattern as persistent, intermittent, or already recovered
- Uses regex match on the probe_url label (probe_url=~".*<hostname>.*") to handle varying label formats across RMO versions (the label may include scheme and path, e.g. https://console-openshift-console.apps.cluster.example.com/health)
- Skipped on HCP clusters where probe_success is forwarded to RHOBS and not available in in-cluster Prometheus
Load Balancer Security Group checks (AWS Classic clusters only):
- CLB path: retrieves CLB security groups and verifies TCP 443 inbound is allowed from 0.0.0.0/0 or machine CIDR
- NLB path: warns if the NLB has security groups attached (unusual per SOP), retrieves target instance security groups, and verifies the NodePort range (30000-32767) is allowed inbound
- Extends the AWS Client interface with GetCLBSecurityGroupIDs, GetSecurityGroupRules, GetInstanceSecurityGroupIDs
- Extends FindNLBByDNSName to also return NLB security groups
Both checks are informational-only — no automated actions (no service logs, no silencing).

#### Test coverage checks
- [x] Added tests

- [x] Ran unit tests locally

## Summary by CodeRabbit

* **New Features**
  * Enhanced console error budget burn investigation by validating load balancer and instance security groups against required TCP access (including NodePort range).
  * Added optional Prometheus-based probe history analysis to classify persistent vs. intermittent failures for classic clusters.
* **Bug Fixes**
  * Improved health decision logic to account for security group rules when determining allowed inbound traffic.
* **Tests**
  * Expanded unit and integration test coverage for security group rule evaluation and probe history scenarios.

### DIFF
--- a/pkg/aws/aws.go
+++ b/pkg/aws/aws.go
@@ -103,10 +103,12 @@ type Client interface {
 	FindHostedZone(ctx context.Context, dnsName string, private bool) (string, error)
 	HasResourceRecordSet(ctx context.Context, hostedZoneID, recordName, recordType string) (bool, error)
 	GetVpcDhcpConfiguration(ctx context.Context, infraID string) ([]string, error)
-	FindNLBByDNSName(ctx context.Context, dnsName string) (arn string, name string, err error)
-	FindCLBByDNSName(ctx context.Context, dnsName string) (name string, err error)
+	FindNLBByDNSName(ctx context.Context, dnsName string) (arn string, name string, securityGroups []string, err error)
+	FindCLBByDNSName(ctx context.Context, dnsName string) (name string, securityGroups []string, err error)
 	GetNLBTargetHealth(ctx context.Context, lbARN string) ([]NLBTargetHealth, error)
 	GetCLBInstanceHealth(ctx context.Context, lbName string) ([]CLBInstanceHealth, error)
+	GetSecurityGroupRules(ctx context.Context, sgIDs []string) ([]ec2v2types.SecurityGroup, error)
+	GetInstanceSecurityGroupIDs(ctx context.Context, instanceIDs []string) ([]string, error)
 }
 
 type SdkClient struct {
@@ -728,18 +730,18 @@ func (c *SdkClient) GetVpcDhcpConfiguration(ctx context.Context, infraID string)
 	return servers, nil
 }
 
-func (c *SdkClient) FindNLBByDNSName(ctx context.Context, dnsName string) (string, string, error) {
+func (c *SdkClient) FindNLBByDNSName(ctx context.Context, dnsName string) (string, string, []string, error) {
 	var marker *string
 	for {
 		out, err := c.Elbv2Client.DescribeLoadBalancers(ctx, &elbv2.DescribeLoadBalancersInput{
 			Marker: marker,
 		})
 		if err != nil {
-			return "", "", fmt.Errorf("failed to describe NLBs: %w", err)
+			return "", "", nil, fmt.Errorf("failed to describe NLBs: %w", err)
 		}
 		for _, lb := range out.LoadBalancers {
 			if lb.Type == elbv2types.LoadBalancerTypeEnumNetwork && awsv2.ToString(lb.DNSName) == dnsName {
-				return awsv2.ToString(lb.LoadBalancerArn), awsv2.ToString(lb.LoadBalancerName), nil
+				return awsv2.ToString(lb.LoadBalancerArn), awsv2.ToString(lb.LoadBalancerName), lb.SecurityGroups, nil
 			}
 		}
 		if out.NextMarker == nil {
@@ -747,21 +749,21 @@ func (c *SdkClient) FindNLBByDNSName(ctx context.Context, dnsName string) (strin
 		}
 		marker = out.NextMarker
 	}
-	return "", "", nil
+	return "", "", nil, nil
 }
 
-func (c *SdkClient) FindCLBByDNSName(ctx context.Context, dnsName string) (string, error) {
+func (c *SdkClient) FindCLBByDNSName(ctx context.Context, dnsName string) (string, []string, error) {
 	var marker *string
 	for {
 		out, err := c.ElbClient.DescribeLoadBalancers(ctx, &elbv1.DescribeLoadBalancersInput{
 			Marker: marker,
 		})
 		if err != nil {
-			return "", fmt.Errorf("failed to describe CLBs: %w", err)
+			return "", nil, fmt.Errorf("failed to describe CLBs: %w", err)
 		}
 		for _, lb := range out.LoadBalancerDescriptions {
 			if awsv2.ToString(lb.DNSName) == dnsName {
-				return awsv2.ToString(lb.LoadBalancerName), nil
+				return awsv2.ToString(lb.LoadBalancerName), lb.SecurityGroups, nil
 			}
 		}
 		if out.NextMarker == nil {
@@ -769,7 +771,7 @@ func (c *SdkClient) FindCLBByDNSName(ctx context.Context, dnsName string) (strin
 		}
 		marker = out.NextMarker
 	}
-	return "", nil
+	return "", nil, nil
 }
 
 func (c *SdkClient) GetNLBTargetHealth(ctx context.Context, lbARN string) ([]NLBTargetHealth, error) {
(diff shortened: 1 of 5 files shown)


pkg/aws/aws.go
```diff
@@ -103,10 +103,13 @@ type Client interface {
 	FindHostedZone(ctx context.Context, dnsName string, private bool) (string, error)
 	HasResourceRecordSet(ctx context.Context, hostedZoneID, recordName, recordType string) (bool, error)
 	GetVpcDhcpConfiguration(ctx context.Context, infraID string) ([]string, error)
-	FindNLBByDNSName(ctx context.Context, dnsName string) (arn string, name string, err error)
+	FindNLBByDNSName(ctx context.Context, dnsName string) (arn string, name string, securityGroups []string, err error)
 	FindCLBByDNSName(ctx context.Context, dnsName string) (name string, err error)
 	GetNLBTargetHealth(ctx context.Context, lbARN string) ([]NLBTargetHealth, error)
 	GetCLBInstanceHealth(ctx context.Context, lbName string) ([]CLBInstanceHealth, error)
+	GetCLBSecurityGroupIDs(ctx context.Context, lbName string) ([]string, error)
```
rolandmkunkel: Could we just return the security groups in FindCLBByDNSName as we do in FindNLBByDNSName? ie lb.SecurityGroups in FindCLBByDNSName

pkg/investigations/consoleerrorbudgetburn/consoleerrorbudgetburn.go
```diff
@@ -466,6 +523,50 @@ func (i *Investigation) checkNLBHealth(ctx context.Context, r *investigation.Res
 	}
 
 	reportNLBTargetHealth(name, targets, notes)
+
```
rolandmkunkel, line 530: consider extracting the infra node SG check into its own function. the function can be split into 2 parts sharing only the targets slice.

pkg/investigations/consoleerrorbudgetburn/consoleerrorbudgetburn.go
```diff
@@ -466,6 +523,50 @@ func (i *Investigation) checkNLBHealth(ctx context.Context, r *investigation.Res
...
+	}
+	if len(instanceIDs) == 0 {
+		return
+	}
+	instanceSGs, err := r.AwsClient.GetInstanceSecurityGroupIDs(ctx, instanceIDs)
+	if err != nil {
+		notes.AppendWarning("LB Security: failed to get infra node security groups — %v", err)
+		return
+	}
+	sgSet := make(map[string]struct{})
```
rolandmkunkel: GetInstanceSecurityGroupIDs returns a map[string][]string (instance -> SGs), but the caller immediately flattens it into a deduplicated set. Could we have the method return a flat []string of unique SG IDs? Alternatively, we can put a comment if we plan on using that later on

pkg/investigations/consoleerrorbudgetburn/consoleerrorbudgetburn.go
```diff
@@ -1118,6 +1241,113 @@ func (i *Investigation) checkBlackboxProbe(ctx context.Context, r *investigation
...
+		notes.AppendSuccess("Probe History: skipped on HCP — probe_success is forwarded to RHOBS, not available in in-cluster Prometheus")
+	}
+}
+
+// defaultProbeHistoryChecker implements probeHistoryChecker by exec'ing curl
+// into a cluster-monitoring-operator pod to query the in-cluster Prometheus.
+type defaultProbeHistoryChecker struct{}
+
+func (d *defaultProbeHistoryChecker) queryProbeHistory(ctx context.Context, k8sClient k8sclient.Client, restConfig *rest.Config, consoleURL string) (string, error) {
+	podList := &corev1.PodList{}
```
rolandmkunkel: This block until line 1278 appears multiple times in the code: here, lines 894-910, and very similar version from line 1090. Can we make this a helper function?

pkg/investigations/consoleerrorbudgetburn/consoleerrorbudgetburn.go
```diff
@@ -1118,6 +1241,113 @@ func (i *Investigation) checkBlackboxProbe(ctx context.Context, r *investigation
...
+			cmoPod = &podList.Items[idx]
+			break
+		}
+	}
+	if cmoPod == nil {
+		return "", fmt.Errorf("no running cluster-monitoring-operator pod found in openshift-monitoring")
+	}
+
+	// Build the Prometheus range query URL.
+	hostname := strings.TrimPrefix(consoleURL, "https://")
```
rolandmkunkel: Consider using

```suggestion
	parsedURL, err := url.Parse(consoleURL)
	if err != nil {
		return "", fmt.Errorf("failed to parse console URL %q: %w", consoleURL, err)
	}
	hostname := parsedURL.Hostname()
```

pkg/investigations/consoleerrorbudgetburn/consoleerrorbudgetburn.go
```diff
@@ -1118,6 +1241,113 @@ func (i *Investigation) checkBlackboxProbe(ctx context.Context, r *investigation
...
+				failures++
+			}
+		}
+	}
+
+	switch failures {
+	case 0:
+		notes.AppendSuccess("Probe History: probe has been succeeding for all %d data points in the past 2 hours — failure may be intermittent or already resolved", total)
+	case total:
+		notes.AppendWarning("Probe History: probe has been consistently failing — all %d data points in the past 2 hours show failure", total)
```
rolandmkunkel: How was the 2h time window chosen? Does it match the lookback window for the alert? If it is lower than 2h, this will hardly ever report something as consistently failing, because if some issue came up 1h ago, and the alert fires after 1h, this will still have 1h of successful calls in the 2h window and report it as intermittently failing, even if it has been consistently failing for 1h. Alternatively, analyzing the tail of the data would distinguish recent-onset from truly intermittent failure.

pkg/investigations/consoleerrorbudgetburn/consoleerrorbudgetburn.go
```diff
@@ -1118,6 +1241,113 @@ func (i *Investigation) checkBlackboxProbe(ctx context.Context, r *investigation
...
+	if cmoPod == nil {
+		return "", fmt.Errorf("no running cluster-monitoring-operator pod found in openshift-monitoring")
+	}
+
+	// Build the Prometheus range query URL.
+	hostname := strings.TrimPrefix(consoleURL, "https://")
+	hostname = strings.TrimPrefix(hostname, "http://")
+	now := time.Now()
+	params := url.Values{}
+	params.Set("query", fmt.Sprintf(`probe_success{probe_url=~".*%s.*"}`, hostname))
```
rolandmkunkel, line 1265: consider escaping the hostname this to make it more robust against code injection

pkg/investigations/consoleerrorbudgetburn/consoleerrorbudgetburn.go
MateSaary: You're right, 2h is not matched to a specific alert tier (since the alert has multiple burn rates, thus more nuanced than to pick a specific window), but I didn't consider the potential for falsely reporting persistence vs intermittent fails. 
I think the cleanest solution here is to revert to a 1 hour lookback (matching to the `CRITICAL` severity), and to explicitly label it as such to make it clear it's a "recent history snapshot" in the case of longer burn windows.

pkg/investigations/consoleerrorbudgetburn/consoleerrorbudgetburn.go
```diff
@@ -964,27 +1087,14 @@ func (i *Investigation) checkNodeUtilization(ctx context.Context, r *investigati
...
-		return "", fmt.Errorf("no running blackbox-exporter pod found in openshift-route-monitor-operator")
+	bbPod, err := findRunningPod(ctx, k8sClient, "openshift-route-monitor-operator",
+		map[string]string{"app": "blackbox-exporter"},
+		"blackbox-exporter")
+	if err != nil {
+		return "", err
 	}
 
-	probeURL := fmt.Sprintf("http://localhost:9115/probe?target=%s&module=http_2xx&debug=true", consoleURL)
+	probeURL := fmt.Sprintf("http://localhost:9115/probe?target=%s&module=ttp_2xx&debug=true", consoleURL)
```
rolandmkunkel: typo in "ttp_2xx"

PR #859, opened 2026-06-26
Multi-investigation support; deprecate registry-based configuration.

gvnnn:
Adds support for executing multiple investigations for a given alert; The chain of investigations is defined via a configuration file.

### What type of PR is this?

feature

This PR adds multi-investigation support to CAD through the mapping of pipelines of investigations to a given alert.

Do not merge yet

#### Test coverage checks
- [ ] Added tests

- [ ] Ran unit tests locally

## Summary by CodeRabbit

* **New Features**
  * Added alert-title–driven, ordered investigation chains with `when` support at both the alert level and each investigation entry.
  * Added configurable AI-assisted behavior when no alert chain matches.
* **Bug Fixes**
  * Improved consistent alert-chain routing and escalation behavior across PagerDuty and manual flows.
  * Updated must-gather and service log/analysis reporting to use metrics for more reliable instrumentation.
* **Documentation**
  * Updated configuration docs/examples and e2e fixtures to the new `alerts`/`investigations` schema.
* **Tests**
  * Migrated and expanded unit and e2e coverage for chain parsing and selection.
* **Chores**
  * Interceptor now validates configuration at startup and fails fast; removed the extra payload encoding from interceptor “continue” responses.

### DIFF
--- a/docs/investigation-config.md
+++ b/docs/investigation-config.md
@@ -1,6 +1,6 @@
-# Investigation Filter Configuration
+# Investigation Configuration
 
-CAD supports a YAML-based configuration file that controls which investigations run and optionally configures the AI agent. This allows you to restrict investigations to specific clusters, cloud providers, organizations, or other attributes — without code changes.
+CAD is configured via a YAML-based configuration file that controls which investigations run and optionally configures the AI agent. This allows you to restrict investigations to specific alerts, clusters, cloud providers, organizations, or other attributes — without code changes.
 
 ## Enabling the config
 
@@ -10,15 +10,15 @@ Set the `CAD_INVESTIGATION_CONFIG_PATH` environment variable to the path of your
 export CAD_INVESTIGATION_CONFIG_PATH=/path/to/config.yaml
 ```
 
-If the variable is not set, no filtering is applied and all investigations run unconditionally.
+If the variable is not set, the program exits.
 
 ## How filtering works
 
-Each investigation can have an optional **filter tree** — a boolean expression evaluated against the current alert and cluster context. If the filter passes, the investigation runs; if it fails, it is skipped.
+Each alert configuration can have an optional **filter tree** — a boolean expression evaluated against the current alert and cluster context. If the filter passes, the alert is investigated by CAD by running its configured set of investigations; Each investigation can, in turn, have its own filter tree.
 
-Investigations that have **no entry** in the config always run. An alert entry with no `when` key also always runs.
+Investigations that have **no `when` entry** in the config always run for a given alert.
 
-The one exception is `aiassisted`: it only runs if it has an entry in the config. Without an entry, the AI investigation is entirely disabled.
+The `aiassisted` investigation can run in two ways: explicitly listed in an alert's investigation chain, or as an automatic fallback when no alert title matches (or when the alert-level filter rejects). The fallback path triggers whenever the `ai_agent` section is configured, even without an explicit `aiassisted` entry in `alerts`.
 
 ## Filter tree structure
 
@@ -90,7 +90,7 @@ Only run an investigation on AWS clusters in `ready` state:
 
 ```yaml
 alerts:
-  - investigation: mustgather
+  - alert_title: "MustGather"
     when:
       and:
         - field: CloudProvider
@@ -99,47 +99,62 @@ alerts:
         - field: ClusterState
           operator: in
           values: ["ready"]
+    investigations:
+      - precheck
+      - mustgather
 ```
 
 Exclude a specific cluster:
 
 ```yaml
 alerts:
-  - investigation: chgm
+  - alert_title: "has gone missing"
     when:
       field: ClusterID
       operator: notin
       values: ["2pr3e91qrgdje312keq8denphqs70tlr"]
+    investigations:
+      - precheck
+      - ccam
+      - chgm
 ```
 
-Sample 10% of internal (`@redhat.com`) traffic, always run for external customers:
+Sample 10% of internal (`@redhat.com`) traffic for a specific investigation step, always run for external customers:
 
 ```yaml
 alerts:
-  - investigation: clustermonitoringerrorbudgetburn
-    when:
-      or:
-        - field: OwnerEmail
-          operator: notmatches
-          values: [".*@redhat\\.com$"]
-        - operator: sample
-          values: ["0.10"]
+  - alert_title: "ClusterMonitoringErrorBudgetBurnSRE"
+    investigations:
+      - precheck
+      - ccam
+      - name: clustermonitoringerrorbudgetburn
+        when:
+          or:
+            - field: OwnerEmail
+              operator: notmatches
+              values: [".*@redhat\\.com$"]
+            - operator: sample
+              values: ["0.10"]
 ```
 
 ## AI agent configuration
 
-When using the `aiassisted` investigation, the `ai_agent` section must be present:
+When using the `aiassisted` investigation, the `ai_agent` section must be present and all required fields must be set:
 
 ```yaml
 ai_agent:
-  runtime_arn: "arn:aws:bedrock:us-east-1:123456789012:agent-runtime/EXAMPLE"
-  user_id: "cad-agent"
+  runtime_arn: "arn:aws:bedrock-agentcore:us-east-1:123456789012:runtime/agent-abc123"
+  user_id: "cad-service-account"
   region: "us-east-1"
-  timeout_seconds: 900   # optional, defaults to 900
+  invoker_role_arn: "arn:aws:iam::123456789012:role/agent-invoker"
+  timeout_seconds: 900              # optional, defaults to 900
+  version: "v1.0.0"                 # optional, audit trail only
+  ops_sop_version: "v2.0.0"         # optional, audit trail only
+  rosa_plugins_version: "v3.0.0"    # optional, audit trail only
 ```
 
-The `aiassisted` investigation must also have an entry in `alerts`. Without it, AI investigation is disabled even if `ai_agent` is configured.
+When the `ai_agent` section is configured, `aiassisted` also acts as a fallback: if no alert title matches the incoming incident (or the matched alert's `when` filter rejects), CAD automatically runs `precheck` followed by `aiassisted`. This fallback does not require an explicit `aiassisted` entry in `alerts`.
 
 ## Full reference
 
-See [`docs/investigation-filter-config.example.yaml`](investigation-filter-config.example.yaml) for a fully commented example covering all operators, field types, and composition patterns.
+See [`docs/investigation-config.example.yaml`](investigation-config.example.yaml) for a fully commented example covering all operators, field types, and composition patterns.
--- a/docs/investigation-filter-config.example.yaml
+++ b/docs/investigation-filter-config.example.yaml
@@ -1,16 +1,22 @@
-# Investigation Filter Configuration
+# Investigation Chain Configuration
 #
-# This file controls which investigations are allowed to run based on
-# cluster and alert attributes. It is loaded from the file path specified
-# in the CAD_INVESTIGATION_CONFIG_PATH environment variable.
+# This file defines which investigations run for each alert, and
+# optionally filters them based on cluster and alert attributes. It is
+# loaded from the file path specified in the CAD_INVESTIGATION_CONFIG_PATH
+# environment variable.
 #
-# If the environment variable is not set, no filtering is applied and all
-# investigations run unconditionally.
+# If the environment variable is not set, no config is loaded and alerts
+# without a matching chain are escalated to SRE.
 #
 # Structure:
-#   filters:             # Top-level list of per-investigation filters.
-#     - investigation:   # Investigation name (must match a registered name, see below).
-#       when:          # Root of a recursive filter tree. Optional (nil = always run).
+#   investigations:          # List of alert → chain mappings.
+#     - alert_title:         # Substring matched against the PagerDuty incident title.
+#       experimental: false  # Optional. If true, chain only matches when CAD_EXPERIMENTAL_ENABLED=true.
+#       when:                # Optional chain-level filter. If it blocks, the entire chain is skipped
+#                            # and the alert is escalated to PagerDuty.
+#       chain:               # Ordered list of investigations to run.
+#                            # Each entry is either a bare string (investigation name) or an object
+#                            # with `name` and optional `when` filter.
 #
 # Filter Tree:
 #   A filter node is either a branch (AND/OR) or a leaf (comparison/sampling).
@@ -30,15 +36,15 @@
 #     operator: sample   # Probabilistic sampling — no field required.
 #     values: ["0.10"]   # Single value: probability between 0 and 1.
 #
-# Behavior:
-#   - Investigations WITHOUT a filter entry here always run (no restrictions).
-#   - A filter entry with no `when` tree (just the investigation name) always runs.
-#   - An empty filters list (filters: []) means no filtering — all investigations run.
-#   - A node cannot have both `and` and `or`, or be both a branch and a leaf.
-#   - The `aiassisted` investigation is special: it only runs when it has an entry
-#     in this config. Without an entry, AI investigation is disabled entirely.
+# Filtering levels:
+#   - Chain-level `when`: evaluated before any chain entry runs. If blocked, the
+#     entire chain is skipped and the alert is escalated to PagerDuty with a note.
+#   - Entry-level `when`: evaluated per step. If blocked, that step is skipped and
+#     the chain continues with the next entry.
 #
 # Valid investigation names:
+#   - precheck
+#   - ccam
 #   - aiassisted
 #   - Cluster Has Gone Missing (CHGM)
 #   - ClusterProvisioningDelay
@@ -50,6 +56,8 @@
 #   - restartcontrolplane
 #   - cannotretrieveupdatessre
 #   - mustgather
+#   - ocmagentresponsefailure
+#   - describenodes
 #
 # Valid input fields (from FilterContext):
 #
@@ -78,7 +86,7 @@
 # Notes:
 #   - The HCP field is a boolean internally but is compared as a string
 #     ("true" / "false") in filter values.
-#   - Each investigation name may only appear once in the filters list.
+#   - Each alert_title must be unique across all investigation entries.
 #   - Not all FilterContext fields are guaranteed to be populated in every
 #     context. For example, PagerDuty fields are empty when running via the
 #     manual CLI. An empty field will not match any "in" value (except an
@@ -88,7 +96,21 @@
 #
 # This section configures the AgentCore runtime used by the `aiassisted`
 # investigation. All fields except the version metadata are required when
-# AI investigation is enabled.
+# AI investigation is enabled (i.e., when any chain references `aiassisted`).
+#
+# AI investigation guards:
+#   1. The `ai_agent` section must be present — without it, any chain
+#      referencing `aiassisted` is rejected at config validation time.
+#   2. Every `aiassisted` chain entry must have a `when` filter — either
+#      on the chain itself or on the entry. This prevents uncontrolled AI
+#      execution. Configs without a filter are rejected at validation time.
+#   3. The `CAD_EXPERIMENTAL_ENABLED=true` env var must be set for the
+#      AI fallback path (unmatched alerts) to trigger. Chains explicitly
+#      configured for specific alerts do not require this env var unless
+#      they are marked `experimental: true`.
+#   4. At runtime, the `when` filter is evaluated against real cluster data
+#      (org ID, cluster ID, cloud provider, etc.) to decide whether AI
+#      actually runs for a given cluster.
 #
 # ai_agent:
 #   runtime_arn: "arn:aws:bedrock-agentcore:us-east-1:123456789012:runtime/rosasreai_stage-abc123"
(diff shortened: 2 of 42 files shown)


docs/investigation-filter-config.example.yaml
```diff
@@ -1,16 +1,22 @@
...
-#     - investigation:   # Investigation name (must match a registered name, see below).
-#       when:          # Root of a recursive filter tree. Optional (nil = always run).
+#   investigations:          # List of alert → chain mappings.
+#     - alert_title:         # Substring matched against the PagerDuty incident title.
+#       experimental: false  # Optional. If true, chain only matches when CAD_EXPERIMENTAL_ENABLED=true.
+#       when:                # Optional chain-level filter. If it blocks, the entire chain is skipped
+#                            # and the alert is escalated to PagerDuty.
+#       chain:               # Ordered list of investigations to run.
+#                            # Each entry is either a bare string (investigation name) or an object
+#                            # with `name` and optional `when` filter.
```
bergmannf:
```suggestion
#       chain:               # Ordered list of investigations to run.
#                            # Each entry is either a bare string (investigation name) or an object
#                            # with `name` and optional `when` filter.
```

Minor thing, but a chain isn't a set of steps - so maybe those shouldn't also be called investigations in the comment?

docs/investigation-filter-config.example.yaml
```diff
@@ -88,7 +96,21 @@
...
 # investigation. All fields except the version metadata are required when
-# AI investigation is enabled.
+# AI investigation is enabled (i.e., when any chain references `aiassisted`).
+#
+# AI investigation guards:
+#   1. The `ai_agent` section must be present — without it, any chain
+#      referencing `aiassisted` is rejected at config validation time.
+#   2. Every `aiassisted` chain entry must have a `when` filter — either
+#      on the chain itself or on the entry. This prevents uncontrolled AI
+#      execution. Configs without a filter are rejected at validation time.
```
bergmannf:
```suggestion
#   2. Every `aiassisted` chain entry must have a `when` filter — either
#      on the chain itself or on the entry. This prevents uncontrolled AI
#      execution. Configs without a filter are rejected at validation time.
```

This is no longer the case and must be updated in this PR.

docs/investigation-filter-config.example.yaml
```diff
@@ -100,85 +122,141 @@
...
-        - field: ClusterState
-          operator: in
-          values: ["ready"]
+      field: ClusterID
+      operator: notin
+      values: ["2pr3e91qrgdje312keq8denphqs70tlr"]
+    chain:
+      - precheck
+      - ccam
+      - "Cluster Has Gone Missing (CHGM)"
```
bergmannf:
```suggestion
      - "Cluster Has Gone Missing (CHGM)"
```

I don't like that some of the chain use shortened symbolic names (ccam) and this one is some extremely long string.

Either all should be long names or shortnames not a strange unclear mix of both.

docs/investigation-filter-config.example.yaml
```diff
@@ -100,85 +122,141 @@
...
+          values: ["0.10"]
 
-  # Only run the etcd investigation on non-HCP (classic) clusters.
-  - investigation: etcddatabasequotalowspace
+  # Cluster Provisioning Delay
+  - alert_title: "ClusterProvisioningDelay -"
+    chain:
+      - precheck
+      - ccam
+      - ClusterProvisioningDelay
```
bergmannf:
```suggestion
      - ClusterProvisioningDelay
```

Weird mix again - CCAM vs this.

interceptor/pkg/interceptor/pdinterceptor.go
```diff
@@ -53,11 +53,14 @@ type Organization struct {
...
-	stats *InterceptorStats
+	stats        *InterceptorStats
+	filterConfig *config.Config
+	configErr    error
 }
 
 func CreateInterceptorHandler(stats *InterceptorStats) http.Handler {
-	return &interceptorHandler{stats}
+	cfg, err := config.LoadConfig("", investigations.GetAvailableInvestigationsNames())
+	return &interceptorHandler{stats: stats, filterConfig: cfg, configErr: err}
```
bergmannf:
```suggestion
	return &interceptorHandler{stats: stats, filterConfig: cfg, configErr: err}
```

This looks strange - why is the config error not returned as an error value?

pkg/config/config.go
```diff
@@ -41,142 +35,73 @@ func (c *AIAgentConfig) GetTimeout() time.Duration {
...
+// LoadConfig reads and parses the investigation configuration.
+// If pathOverride is non-empty, it is used as the config file path.
+// Otherwise, the path is read from the CAD_INVESTIGATION_CONFIG_PATH environment variable.
+// Returns an error if no config path is provided or the file cannot be read.
+// The validInvestigations parameter is the list of known investigation names used to
+// validate that each chain entry references a real investigation.
+func LoadConfig(pathOverride string, validInvestigations []string) (*Config, error) {
+	path := pathOverride
+	if path == "" {
+		path = os.Getenv(ConfigEnvVar)
```
bergmannf:
```suggestion
		path = os.Getenv(ConfigEnvVar)
```

Nesting an os.Getenv lookup here is not that clean - we should move all lookups to the outermost layer and pass them in.

pkg/config/config.go
```diff
@@ -195,6 +120,23 @@ func ParseConfig(data []byte, validInvestigations []string) (*Config, error) {
 	return &cfg, nil
 }
 
+// GetChain returns the first InvestigationConfig whose AlertTitle is contained in the given alert title.
```
bergmannf:
```suggestion
// GetChain returns the first InvestigationConfig whose AlertTitle is contained in the given alert title.
```

Instead of returning the first, should the parsing code detect if an alerttitle is duplicated and error out?

pkg/config/config.go
```diff
@@ -222,56 +164,65 @@ func (c *Config) Validate(validInvestigations []string) error {
...
-			return &c.Filters[i]
+			if entry.Name == "aiassisted" {
+				hasAIAssisted = true
+				// aiassisted must always be gated by a filter to prevent uncontrolled
+				// AI execution. Either a chain-level or entry-level when clause satisfies this.
+				if ic.When == nil && entry.When == nil {
+					return fmt.Errorf(
+						"investigations[%d].chain[%d]: aiassisted requires a 'when' filter "+
+							"(on the chain or entry level) to control execution", i, j)
+				}
```
bergmannf:
```suggestion
				// aiassisted must always be gated by a filter to prevent uncontrolled
				// AI execution. Either a chain-level or entry-level when clause satisfies this.
				if ic.When == nil && entry.When == nil {
					return fmt.Errorf(
						"investigations[%d].chain[%d]: aiassisted requires a 'when' filter "+
							"(on the chain or entry level) to control execution", i, j)
				}
```

No longer the case must probably be updated/rebased.

pkg/controller/controller.go
```diff
@@ -13,11 +13,8 @@ import (
 	"github.com/openshift/configuration-anomaly-detection/pkg/config"
 	"github.com/openshift/configuration-anomaly-detection/pkg/executor"
 	"github.com/openshift/configuration-anomaly-detection/pkg/investigations"
-	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/ccam"
-	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/chgm"
-	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/expiredcertificates"
```
bergmannf, line 20:
```suggestion
	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/expiredcertificates"
```

This seems to have been silently dropped everywhere and must be put back.

pkg/controller/controller.go
```diff
@@ -266,155 +261,162 @@ func NewController(opts ControllerOptions, deps *Dependencies) (Controller, erro
...
+			logging.Errorf("Chain-level filter error for %q: %v", chainConfig.AlertTitle, filterErr)
+			return nil
+		}
+		if !pass {
+			logging.Infof("Chain %q filtered out: %s", chainConfig.AlertTitle, reason)
+			if pdClient != nil {
+				if escErr := pdClient.EscalateIncidentWithNote(fmt.Sprintf("🤖 Investigation chain %q was filtered: %s. Escalating to SRE. 🤖", chainConfig.AlertTitle, reason)); escErr != nil {
+					logging.Errorf("Failed to escalate filtered chain: %v", escErr)
+				}
+			}
```
bergmannf:
```suggestion
			if pdClient != nil {
				if escErr := pdClient.EscalateIncidentWithNote(fmt.Sprintf("🤖 Investigation chain %q was filtered: %s. Escalating to SRE. 🤖", chainConfig.AlertTitle, reason)); escErr != nil {
					logging.Errorf("Failed to escalate filtered chain: %v", escErr)
				}
			}
```

We have a PagerDutyController and a ManualController - I think instead of doing these != nil checks, we should move this into a polymorphic dispatch that is just a noop on a non-PD controller.

pkg/controller/controller.go
```diff
@@ -266,155 +261,162 @@ func NewController(opts ControllerOptions, deps *Dependencies) (Controller, erro
...
-	ccamInvestigation := ccam.CloudCredentialsCheck{}
-	result, err = ccamInvestigation.Run(builder)
-	if err != nil {
-		return err
-	}
-	// FIXME: Once all migrations are converted this can be removed.
-	updateMetrics(inv.Name(), &result)
+		// Create a fresh aiassisted instance with the runtime config to avoid mutating the registry singleton.
+		if _, ok := inv.(*aiassisted.Investigation); ok && c.dependencies.FilterConfig != nil {
+			inv = &aiassisted.Investigation{AIConfig: c.dependencies.FilterConfig.GetAIAgentConfig()}
```
bergmannf:
```suggestion
			inv = &aiassisted.Investigation{AIConfig: c.dependencies.FilterConfig.GetAIAgentConfig()}
```

Minor nitpick and not due to this PR, but we should stop calling this 'FilterConfig' - it has become a normal Config that also contains the FilterConfig.

pkg/controller/controller.go
```diff
@@ -266,155 +261,162 @@ func NewController(opts ControllerOptions, deps *Dependencies) (Controller, erro
...
+			return nil
+		}
 	}
 
-	// Record completion for manual investigations with outcome
-	if hasFindings {
-		c.recordManualCompletion(inv.Name(), pdClient, "success")
-	} else {
-		c.recordManualCompletion(inv.Name(), pdClient, "no_findings")
+	// Post-chain: title update (PD mode only)
```
bergmannf, line 410:
```suggestion
	if hasFindings {
		c.recordManualCompletion(inv.Name(), pdClient, "success")
	} else {
		c.recordManualCompletion(inv.Name(), pdClient, "no_findings")
	// Post-chain: title update (PD mode only)
```

It seems these manual completion calls are completely gone, which would drop metrics for manual runs?

pkg/controller/manual.go
```diff
@@ -68,24 +65,39 @@ func (c *ManualController) Investigate(ctx context.Context) error {
...
+	if inv.Name() != "ccam" && inv.Name() != "precheck" {
+		chain = append(chain, config.ChainEntry{Name: "ccam"})
+	}
+	chain = append(chain, config.ChainEntry{Name: inv.Name()})
+
+	chainConfig := &config.InvestigationConfig{
+		AlertTitle: inv.Name(),
+		Name:       inv.Name(),
+		Chain:      chain,
+	}
```
bergmannf:
```suggestion
	chain := []config.ChainEntry{}
	if inv.Name() != "precheck" {
		chain = append(chain, config.ChainEntry{Name: "precheck"})
	}
	if inv.Name() != "ccam" && inv.Name() != "precheck" {
		chain = append(chain, config.ChainEntry{Name: "ccam"})
	}
	chain = append(chain, config.ChainEntry{Name: inv.Name()})

	chainConfig := &config.InvestigationConfig{
		AlertTitle: inv.Name(),
		Name:       inv.Name(),
		Chain:      chain,
	}
```

Would it be cleaner to provide a valid configuration file with all investigations defined instead of always hardcoding these two steps into every manual investigation?

This would allow a manual run to opt out of steps that might be known to fail.

bergmannf (changes requested):
I did a first pass - looks really good and also cleans up a bunch of fields that felt useless on investigations.

I think there are a few small things that changed on the main branch in the meantime that need to be taken into consideration now though.

docs/investigation-filter-config.example.yaml
gvnnn: agreed, and I've modified the config syntax slightly:

* `investigations` -> `alerts` (it's really a mapping between alerts and their related investigations
* `chain` -> `investigations`

docs/investigation-filter-config.example.yaml
gvnnn: Updated to reflect upstream: aiassisted as a fallback for alerts that have no configured investigations.

docs/investigation-filter-config.example.yaml
gvnnn: This is done, I've updated CHGM and CPD to use short names as well

docs/investigation-filter-config.example.yaml
gvnnn: See above

interceptor/pkg/interceptor/pdinterceptor.go
gvnnn: ~This should do: https://github.com/gvnnn/configuration-anomaly-detection/commit/96b0eac4b4f9a42a56b8121a3dceb6e63e4817ee~

EDIT: I replied to the wrong comment. I've fixed this and the main code for interceptor now checks for the returned error and exists if it is not nil.

pkg/controller/controller.go
gvnnn: The rationale here is that investigations are configured in the config file, so the controller does no longer need to inject them. This also allows skipping a given certification based on the filtering criteria (case on point: the expired certificate investigation is supposed to be run on classic-only for the time being).

pkg/controller/controller.go
gvnnn: This should do: https://github.com/gvnnn/configuration-anomaly-detection/commit/96b0eac4b4f9a42a56b8121a3dceb6e63e4817ee

pkg/controller/controller.go
gvnnn: This PR needs frequent rebases already as it is, if it's ok with you I would address this in a separate change :)

pkg/controller/controller.go
gvnnn: Yep, I need to address this.

pkg/controller/manual.go
gvnnn: This matches what upstream does in the shared [runInvestigation](https://github.com/openshift/configuration-anomaly-detection/blob/aa2c9c1bc3e3ed8c0d4f7c73548776070ea71775/pkg/controller/controller.go#L338) path. I think providing a configuration file to the manual runs would work fine, but it essentially changes how the interface was thought: instead of passing an investigation's short name you would need a config file instead.

I would leave this consideration for another change

pkg/controller/controller.go
gvnnn: done

pkg/controller/controller.go
gvnnn: it was actually just that bit that needed updating. I've renamed it

pkg/controller/controller.go
```diff
@@ -266,162 +268,175 @@ func NewController(opts ControllerOptions, deps *Dependencies) (Controller, erro
...
+// name matches the investigation's metadata.yaml RBAC definition.
+func (c *investigationRunner) runChain(
+	ctx context.Context,
+	clusterId string,
+	alertConfig *config.AlertConfig,
+	filterCtx *types.FilterContext,
+	params map[string]string,
+) (err error) {
+	if len(alertConfig.Investigations) > 0 {
+		metrics.Inc(metrics.Alerts, alertConfig.AlertTitle)
```
rolandmkunkel: This changes the metric label. Before it was the name of the investigation, now it is the alert title. That's arguably a required change, but worth noting that there might be inconsistent data during the rollout.

pkg/controller/notifier.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (n *pdIncidentNotifier) EscalateWithNote(note string) error {
+	return n.client.EscalateIncidentWithNote(note)
+}
+
+func (n *pdIncidentNotifier) AttachToBuilder(builder investigation.ResourceBuilder) {
+	builder.WithPdClient(n.client)
+}
+
+func (n *pdIncidentNotifier) IsActive() bool {
```
rolandmkunkel: IsActive() doesn't convey what callers are actually checking. Both callsites use it to ask "is there a PagerDuty client behind this?" and not whether the notifier is healthy or enabled. Consider renaming to HasPagerDuty() so the intent is immediately clear

interceptor/pkg/interceptor/pdinterceptor.go
```diff
@@ -193,45 +197,33 @@ func (pdi *interceptorHandler) process(ctx context.Context, r *triggersv1.Interc
...
-}
-
-// continueWithEncodedPayload returns a Continue response with the webhook payload
-// base64-encoded as an extension. The TriggerBinding references this extension so
-// the payload reaches the Tekton task without shell metacharacter issues.
-func continueWithEncodedPayload(body string) *triggersv1.InterceptorResponse {
-	return &triggersv1.InterceptorResponse{
-		Continue: true,
-		Extensions: map[string]interface{}{
-			"payload_base64": base64.StdEncoding.EncodeToString([]byte(body)),
```
rolandmkunkel, line 241: Does removing the payload not break the pipeline? How are the contents of the alert passed now?

pkg/controller/controller.go
```diff
@@ -266,162 +268,175 @@ func NewController(opts ControllerOptions, deps *Dependencies) (Controller, erro
...
-	if err != nil {
-		return err
-	}
-	if len(result.Actions) > 0 {
-		if err = c.executeActions(builder, &result, "precheck"); err != nil {
-			return fmt.Errorf("failed to execute precheck actions: %w", err)
+		pass, reason, filterErr := alertConfig.ShouldRun(filterCtx)
+		if filterErr != nil {
+			logging.Errorf("Alert-level filter error for %q: %v", alertConfig.AlertTitle, filterErr)
+			return nil
```
rolandmkunkel: Do we intentionally drop these errors? That would mean a failure to perform the filtering due to a typo or similar would silently drop the alert without a metric or escalation to SRE. This also happens in a few more locations

pkg/config/config.go
```diff
@@ -222,56 +164,58 @@ func (c *Config) Validate(validInvestigations []string) error {
...
-	}
-	for i := range c.Filters {
-		if c.Filters[i].Investigation == investigationName {
-			return &c.Filters[i]
+			if entry.Name == "aiassisted" {
+				hasAIAssisted = true
+			}
+
+			// Validate entry-level when clause
+			if entry.When != nil {
```
rolandmkunkel, line 193: The docs say this field must be present. Should be enforced here

docs/investigation-config.md
```diff
@@ -89,7 +89,7 @@ Note: Not all fields are guaranteed to be populated in every context. PagerDuty
 Only run an investigation on AWS clusters in `ready` state:
 
 ```yaml
-filters:
+alerts:
```
rolandmkunkel, line 92: These examples don't match the new format of Alertconfig anymore

pkg/controller/pagerduty.go
```diff
@@ -36,70 +32,68 @@ func (c *PagerDutyController) Investigate(ctx context.Context) error {
...
-	// OCM fields will be populated inside runInvestigation after precheck.
-	filterCtx := &types.FilterContext{
-		AlertName:   alertInvestigation.AlertTitle(),
-		AlertTitle:  c.pdClient.GetTitle(),
-		ServiceName: c.pdClient.GetServiceName(),
+	// AI fallback: no title match, or title matched but when-clause filtered
+	if cfg != nil && cfg.AIAgent != nil {
+		alertConfig = &config.AlertConfig{
+			AlertTitle: "aiassisted-fallback",
+			Investigations: []config.InvestigationEntry{
```
rolandmkunkel, line 66: we might want to validate this inline config as well so that it doesn't break when something changes

pkg/controller/manual.go
```diff
@@ -68,24 +65,38 @@ func (c *ManualController) Investigate(ctx context.Context) error {
...
 	// Track manual investigation start
 	dryRun := strconv.FormatBool(c.dryRun)
-	metrics.Inc(metrics.ManualInvestigationStarted, alertInvestigation.Name(), dryRun)
+	metrics.Inc(metrics.ManualInvestigationStarted, inv.Name(), dryRun)
 
 	// For AI investigations, create a new instance with the runtime config from the global config.
-	if _, ok := alertInvestigation.(*aiassisted.Investigation); ok {
-		alertInvestigation = &aiassisted.Investigation{
-			AIConfig: c.dependencies.FilterConfig.GetAIAgentConfig(),
+	if _, ok := inv.(*aiassisted.Investigation); ok {
```
rolandmkunkel: This is done again in runChain in controller.go, so that should not be needed. We also only use inv.Name()

pkg/controller/manual.go
```diff
@@ -68,24 +65,38 @@ func (c *ManualController) Investigate(ctx context.Context) error {
...
+		inv = &aiassisted.Investigation{
+			AIConfig: c.dependencies.Cfg.GetAIAgentConfig(),
 		}
 	}
 
+	invEntries := []config.InvestigationEntry{}
+	if inv.Name() != "precheck" {
+		invEntries = append(invEntries, config.InvestigationEntry{Name: "precheck"})
+	}
+	if inv.Name() != "ccam" && inv.Name() != "precheck" {
```
rolandmkunkel, line 73: I think the expiredcertificates check should also get a similar treatment here (it might be, that these changes predate the existence of expiredcertificates)

PR #867, opened 2026-07-02
Human readable output

wviqas-rh:
### What type of PR is this?

Adds human-readable markdown formatting for Cora investigation reports. CAD now parses Cora's JSON output and formats it into a clean markdown before creating cluster reports, so SREs see formatted text instead of raw JSON blobs in PD.

The formatter extracts key fields (cluster ID, alert name, root cause summary, confidence, action steps, escalation decision) and handles edge cases like null commands and empty steps. Planning on adding error handling if the entire markdown parsing fails and no markdown can be outputted, in which case the JSON blob will just be sent out instead of the markdown to ensure the SRE still receives something. 

#### Test coverage checks
- [x] Added tests

- [x] Ran unit tests locally

Jira: https://redhat.atlassian.net/browse/ROSAENG-60533

## Summary by CodeRabbit

* **New Features**
  * AI-assisted investigation results are rendered as a readable Markdown report with cluster/alert details, summary, uppercased confidence, reasoning, evidence, action steps, and an escalation decision.
  * Action steps show optional fenced bash commands when available, or “No action steps available” when none exist.
* **Bug Fixes**
  * Improved streaming handling with more robust JSON extraction/unwrapping and graceful warning fallbacks when parsing fails, while reducing verbose AI payload logging.
* **Tests**
  * Added/extended coverage for report formatting and JSON extraction/unmarshalling into the updated result shape.

Jira: https://redhat.atlassian.net/browse/ROSAENG-60533


pkg/investigations/aiassisted/aiassisted.go
```diff
@@ -199,16 +198,26 @@ func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 	}
 
 	logging.Info("🤖 AI investigation complete")
-	logging.Infof("AI Output:\n%s", aiResponse.String())
+
+	// Parse JSON response from Cora
+	var investigationResult CoraInvestigationResult
+	if err := json.Unmarshal([]byte(aiResponse.String()), &investigationResult); err != nil {
```
BATMAN-JD: I am wondering how this part works with the headers that include special characters the way the code sits right now. 
The aiResponse string contains the metadata headers followed by the JSON. When json.Unmarshal tries to parse aiResponse.String(), it will see:

🤖 AI Investigation Results 🤖
Session ID: cad-...
Runtime: arn:...

{"investigation_id":"inv-123",...}
 I am curious if in your testing you created something that injects the sample json output and it was not more e2e that included these headers as part of aiResponse. If so, I believe the json unmarshal will fail

dakotalongRH: To add to this, in Cora's main.py line 251, it yields the disclaimer saying "This investigation is being performed by an AI agent..." before it yields the json payload on line 367.  Both are yielded by cora and will be accumulated by the scanner loop into the same builder.

So even after separating the header into its own builder (what coderabbit caught), the raw streamed content would still contain the AI disclaimer text + json so the unmarshall might fail.

Maybe you could try json.Unmarshal on each accumulated line individually and use the first one that succeeds?

pkg/investigations/aiassisted/aiassisted.go
```diff
@@ -199,16 +198,26 @@ func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
...
+		notes.AppendWarning("Failed to parse Cora JSON response: %v", err)
+		result.Actions = executor.NoteAndReportFrom(notes, clusterID, c.Name())
+		return result, nil
+	}
```
dakotalongRH: If json parsing fails here, we add a PD note "Failed to parse Cora JSON response" with no investigation content.

Prior to this we sent the aiResponse.String() directly as the backplane report data so we always had a fallback. Maybe we can use the old behavior when parsing fails?

pkg/investigations/aiassisted/formatter.go
```diff
@@ -0,0 +1,57 @@
...
+			fmt.Fprintf(&sb, "%d. %s\n", i+1, step.Action)
+			if step.Command != nil && *step.Command != "" {
+				fmt.Fprintf(&sb, "   ````bash\n   %s\n   ````\n", *step.Command)
+			}
+		}
+		sb.WriteString("\n")
+	}
+
+	// Escalation Decision
+	sb.WriteString("## Escalation Decision\n\n")
```
dakotalongRH, line 49: CodeRabbit flagged this in an earlier review - we define the "EscalationReason" on the struct cora_types and populate but don't use it here. When NeedsEscalation is true, it would be nice to see what the escalation reason is.

pkg/investigations/aiassisted/aiassisted.go
BATMAN-JD: I like that idea. Keep in mind as a Cora team that means we need to make sure we only pass back one valid json object, not sure why we would have more, but something to remember.

BATMAN-JD (commented):
## Code Review

### Critical: SSE double-encoding breaks JSON parsing

The Cora agent's BedrockAgentCore SDK double-encodes the JSON response — `_convert_to_sse()` calls `json.dumps()` on the already-JSON string from `model_dump_json()`, producing:

```
data: "{\n  \"investigation_id\": ...}"
```

The Go code strips `data: ` but not the surrounding quotes, so `json.Unmarshal` receives a JSON **string literal**, not a JSON object, and always fails. The formatted report will never be produced from real Cora output — every invocation hits the fallback warning path.

The `invoke-agent.sh` in rosa-cora-agent already works around this:
```bash
echo "$line" | sed 's/^"\(.*\)"$/\1/' | sed 's/\\n/\n/g' | sed 's/\\t/\t/g'
```

The Go code needs equivalent quote-stripping and unescaping after removing the `data: ` prefix.

### Other findings

1. **Scanner error falls through to Unmarshal** (aiassisted.go ~L198): If `scanner.Err()` is non-nil, the code logs a warning but doesn't return — it falls through to `json.Unmarshal` on potentially partial data. Should return early on scanner error.

2. **NeedsEscalation is cosmetic only** (formatter.go ~L49): The incident is always escalated unconditionally at L105, before the AI response arrives. The report text saying "No escalation needed" could confuse SREs since escalation already happened.

3. **Session metadata removed from report** (aiassisted.go ~L180): Session ID, runtime ARN, and agent/plugin versions were moved from the backplane report to `logging.Infof` only. Pipeline logs are ephemeral — this metadata should stay in the report for auditability.

4. **AI output not logged on success path** (aiassisted.go ~L210): The old `logging.Infof("AI Output:\n%s", ...)` was removed. If the backplane report upload fails, the AI investigation content is unrecoverable.

5. **Unbounded raw output in PD note** (aiassisted.go ~L205): On JSON parse failure, `AppendWarning` embeds the full raw AI response with no size guard. PD notes have server-side limits — a large response could cause the API call to fail.

6. **4-backtick code fence** (formatter.go L42): `````bash` is valid CommonMark but not universally supported. Triple backticks would work identically here (no nested fences) and are safer.

eth1030:
@coderabbitai review

eth1030:
Local testing output for test incident: https://redhat.pagerduty.com/incidents/Q1QWTF52RRC95C
📒Report Details for Report 3GolvqunYmt3E8hWzx6HWswGlHh created at 2026-07-21T15:00:35Z

**Cluster ID**: 2rmi86iil0hechn94os5392q6153qvl4
**Alert Name**: FallbackTestAlert

## Summary

Investigation of FallbackTestAlert for cluster 2rmi86iil0hechn94os5392q6153qvl4 (eth72126) reveals a healthy ROSA cluster in ready state. All 7 nodes are Ready, all 33 cluster operators are Available/Not Degraded, monitoring stack is operational with both Prometheus and Alertmanager pods running successfully. The FallbackTestAlert appears to be a test or placeholder alert with no actual underlying issue detected in the cluster infrastructure, networking, or monitoring components.

**Confidence**: HIGH

**Reasoning**: The cluster shows consistent indicators of healthy operation: (1) All nodes report Ready status with normal resource utilization (CPU 3-11%, Memory 18-34%), (2) All cluster operators including monitoring are Available=True, Degraded=False, Progressing=False, (3) Monitoring stack pods are all Running with 0 restarts in the past 45+ minutes, (4) Persistent volumes for Prometheus and Alertmanager are properly bound, (5) No errors found in recent cluster-monitoring-operator logs, only routine update tasks, (6) Cluster version 4.22.4 is stable and fully deployed. The alert name "FallbackTestAlert" suggests this is a synthetic test alert rather than a real production incident. No corresponding PrometheusRule or actual firing condition was identified in the monitoring namespace logs.

## Evidence

1. Node status: All 7 nodes (3 control-plane, 2 worker, 2 infra) in Ready state with healthy resource metrics
2. Cluster operators: 33/33 operators Available, 0 Degraded, monitoring operator status "Successfully rolled out the stack"
3. Monitoring pods: prometheus-k8s-0/1 and alertmanager-main-0/1 all Running with 6/6 containers ready
4. Storage: All 4 PVCs for monitoring (2x 100Gi prometheus, 2x 10Gi alertmanager) bound to gp3 volumes
5. Cluster metadata: ROSA cluster eth72126 in "ready" state, deployed ~85 minutes ago, expires 2026-07-23
6. No errors in alertmanager/prometheus logs when grepping for "fallback"
7. cluster-monitoring-operator logs show only routine "UpdatingAlertmanager" tasks completing successfully

## Action Steps

1. Verify if FallbackTestAlert is an expected test alert by querying the alerting system configuration or consulting with the team that created it
2. If this is a test alert that should not fire in production, remove or disable the corresponding PrometheusRule
   ```bash
   oc get prometheusrules -A -o yaml | grep -B5 -A10 FallbackTestAlert
   ```
3. If alert routing is the issue, verify alertmanager configuration has proper receivers configured
   ```bash
   oc get secret -n openshift-monitoring alertmanager-main -o jsonpath='{.data.alertmanager\.yaml}' | base64 -d
   ```
4. Monitor the cluster for any actual issues - current state shows all systems healthy and operational
   ```bash
   oc get clusteroperators && oc get nodes && oc get pods -n openshift-monitoring
   ```

## Escalation Decision

✅ No escalation needed

PR #869, opened 2026-07-06
Remove filtered investigation note when escalating to SRE

eth1030:
### What type of PR is this?

(feature/bug/documentation/other)

🤖 Investigation %s was filtered out by configuration. Escalating to SRE. 🤖 is noisy in PagerDuty. The only investigations that escalate in this way are filtered investigations (aiassisted and mustgather) and this information is not valuable to SRE on call.

#### Test coverage checks
- [ ] Added tests

- [ ] Ran unit tests locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated incident escalation handling for filtered investigations so alerts are sent without an extra note.
  * Kept the existing filtering behavior and error logging intact if escalation fails.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -609,7 +609,7 @@ func (c *investigationRunner) evaluateFilter(invName string, filterCtx *types.Fi
 	}
 
 	if pdClient != nil {
-		if escErr := pdClient.EscalateIncidentWithNote(fmt.Sprintf("🤖 Investigation %s was filtered out by configuration. Escalating to SRE. 🤖", invName)); escErr != nil {
+		if escErr := pdClient.EscalateIncident(); escErr != nil {
 			logging.Errorf("Failed to escalate filtered investigation: %v", escErr)
 		}
 	}


geowa4 (approved):
/lgtm

BATMAN-JD (approved):

pkg/controller/controller.go
```diff
@@ -609,7 +609,7 @@ func (c *investigationRunner) evaluateFilter(invName string, filterCtx *types.Fi
 	}
 
 	if pdClient != nil {
-		if escErr := pdClient.EscalateIncidentWithNote(fmt.Sprintf("🤖 Investigation %s was filtered out by configuration. Escalating to SRE. 🤖", invName)); escErr != nil {
+		if escErr := pdClient.EscalateIncident(); escErr != nil {
```
joshbranham, line 612: Should we log that we escalated, so if we are wondering how to tell if CAD actually did something (even if nothing) we can cross reference that in logs?

geowa4: That feels like it should be something the EscalateIncident function does, no?

geowa4: I could see something ahead of this to log that we're doing it. Feels _slightly_ verbose but I'm not paying for that extra line of log storage

rolandmkunkel:
The information that an investigation was filtered is useful in CreateMustGather investigations. I would instead prefer to change the SOC and let the evaluateFilter function no longer do the escalation, but instead do that on the call site, but only for other investigations than aiassisted, since its the only one that works differently.

PR #871, opened 2026-07-07
Add BATMAN-JD and eth1030 as approvers

BATMAN-JD:
## Summary
- Add BATMAN-JD and eth1030 to the approvers list in OWNERS file


BATMAN-JD:
accidentally added my original GH username on first pass. Force pushed to update. Just clarification if you read the messages and see jdowni000

PR #873, opened 2026-07-07
Populate investigation_payload with Prometheus labels for Cora

BATMAN-JD:
## Summary
Populates the previously empty `investigation_payload` (`"{}"`) with structured alert data from PagerDuty, enabling Cora to target specific resources instead of scanning all namespaces.

Per review feedback, PD wire-format parsing lives in `pkg/pagerduty` as shared helpers. The `aiassisted` package consumes typed data from the PD client instead of directly accessing raw SDK alert bodies.

## Changes

**`pkg/pagerduty/types.go`**
- `FiringAlert` struct — matches `{{ .Alerts.Firing | toJson }}` (labels, annotations, timestamps)
- `FiringAlertsResult` — bundles parsed alerts, raw fallback strings, and partial-failure flag
- `AlertCustomDetails` — typed struct for known PD custom detail fields (replaces `map[string]interface{}`)

**`pkg/pagerduty/pagerduty.go`**
- `GetAlertContext(incidentID)` on `Client` interface — single API call returns both `AlertCustomDetails` (from first alert) and `*FiringAlertsResult` (from all alerts)
- `ParseFiringJSON()` — exported function, deserializes `firing_json` into `[]FiringAlert`
- `getRawAlertDetails()` — private helper, loops all PD alerts and extracts `Body["details"]`

**`pkg/pagerduty/pagerduty_test.go`**
- `TestParseFiringJSON` — 5 subtests: single alert, empty string, invalid JSON, empty array, multiple alerts

**`pkg/investigations/aiassisted/aiassisted.go`**
- Call site uses `r.PdClient.GetAlertContext()` (through interface) instead of directly accessing alert bodies

**`pkg/investigations/aiassisted/alert_parser.go`**
- `buildInvestigationPayload()` — consumes `*pagerduty.FiringAlertsResult` and `*pagerduty.AlertCustomDetails`, shapes the AI-specific payload

**`pkg/investigations/aiassisted/alert_parser_test.go`**
- `TestPayloadStructure` — firing alerts + typed details, validates payload shape
- `TestPayloadMultiplePDAlerts` — 3 alerts from multiple PD alerts merged
- `TestPayloadPartialFailure` — mix of parsed alerts and raw fallbacks
- `TestPayloadFallbackRawJSON` — total parse failure, raw string preserved
- `TestPayloadWithRealisticPDJSON` — realistic PD alert body with all detail fields
- `TestPayloadNilInputs` — graceful degradation with nil inputs

## Example payload sent to Cora

```json
{
  "investigation_id": "Q3OFA2AM9FYDAB",
  "investigation_payload": {
    "incident_url": "https://redhat.pagerduty.com/incidents/Q3OFA2AM9FYDAB",
    "firing_alerts": [
      {
        "labels": {"alertname": "KubePersistentVolumeFillingUp", "namespace": "openshift-monitoring", "severity": "critical", "persistentvolumeclaim": "prometheus-data-prometheus-k8s-0"},
        "annotations": {"summary": "...", "runbook_url": "https://..."},
        "startsAt": "2026-07-14T10:00:00Z"
      },
      {
        "labels": {"alertname": "KubePersistentVolumeFillingUp", "namespace": "openshift-monitoring", "severity": "critical", "persistentvolumeclaim": "prometheus-data-prometheus-k8s-1"},
        "annotations": {"summary": "...", "runbook_url": "https://..."},
        "startsAt": "2026-07-14T10:05:00Z"
      }
    ],
    "alert_details": {"link": "https://...", "num_firing": "2", "region": "us-west-2"}
  },
  "alert_name": "KubePersistentVolumeFillingUp CRITICAL (2)",
  "cluster_id": "d9045cbe-..."
}
```

## Dependencies

- Requires `firing_json` field from configure-alertmanager-operator — merged in https://github.com/openshift/configure-alertmanager-operator/pull/588, pending promotion

## Related

- Cora prompt update to use `firing_alerts`: https://github.com/openshift-online/rosa-cora-agent/pull/58 (DRAFT)

cc @clcollins @AlexSmithGH

### DIFF
--- a/pkg/investigations/aiassisted/aiassisted.go
+++ b/pkg/investigations/aiassisted/aiassisted.go
@@ -137,6 +137,8 @@ func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 		return result, nil
 	}
 
+	logging.Infof("Payload: %s", string(payloadJSON))
+
 	// Get AI client (handles role assumption and client creation)
 	// Use incident ID as session identifier for audit trail
 	agentClient, err := aws.GetAIClient(ctx, aiConfig.InvokerRoleArn, aiConfig.Region, incidentID)
@@ -153,7 +155,6 @@ func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 
 	// Log AI invocation
 	logging.Infof("🤖 Invoking AI agent for incident %s", incidentID)
-	logging.Infof("Payload: %s", string(payloadJSON))
 
 	// Request streaming response format
 	acceptHeader := "text/event-stream"
--- a/pkg/investigations/aiassisted/aiassisted.go
+++ b/pkg/investigations/aiassisted/aiassisted.go
@@ -112,14 +112,12 @@ func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 	incidentID := pdClient.GetIncidentID()
 	alertName := pdClient.GetTitle()
 
-	// Extract alert details from PagerDuty
-	alertDetails, err := extractAlertDetails(pdClient)
+	alertDetails, firingResult, err := r.PdClient.GetAlertContext(incidentID)
 	if err != nil {
-		logging.Warnf("Failed to extract alert details: %v", err)
+		logging.Warnf("Failed to extract alert context: %v", err)
 	}
 
-	// Build investigation payload with alert context
-	payloadData := buildInvestigationPayload(alertDetails, pdClient.GetIncidentRef())
+	payloadData := buildInvestigationPayload(firingResult, &alertDetails, pdClient.GetIncidentRef())
 
 	investigationData := &InvestigationPayload{
 		InvestigationID:      incidentID,
--- a/pkg/investigations/aiassisted/aiassisted.go
+++ b/pkg/investigations/aiassisted/aiassisted.go
@@ -118,6 +118,9 @@ func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 	}
 
 	payloadData := buildInvestigationPayload(firingResult, &alertDetails, pdClient.GetIncidentRef())
+	if err != nil {
+		payloadData["alert_context_error"] = err.Error()
+	}
 
 	investigationData := &InvestigationPayload{
 		InvestigationID:      incidentID,
--- a/pkg/investigations/aiassisted/aiassisted.go
+++ b/pkg/investigations/aiassisted/aiassisted.go
@@ -26,10 +26,10 @@ type Investigation struct {
 
 // InvestigationPayload represents the payload sent to the AgentCore agent
 type InvestigationPayload struct {
-	InvestigationID      string `json:"investigation_id"`
-	InvestigationPayload string `json:"investigation_payload"` // TODO: Implement - should contain alert details/context
-	AlertName            string `json:"alert_name"`
-	ClusterID            string `json:"cluster_id"`
+	InvestigationID      string                 `json:"investigation_id"`
+	InvestigationPayload map[string]interface{} `json:"investigation_payload"` // Alert details and context
+	AlertName            string                 `json:"alert_name"`
+	ClusterID            string                 `json:"cluster_id"`
 }
 
 // generateSessionID generates a unique session ID for this investigation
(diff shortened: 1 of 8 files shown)


pkg/investigations/aiassisted/aiassisted.go
```diff
@@ -43,6 +53,132 @@ func generateSessionID(incidentID string) string {
...
+		// Add approved custom details using allow-list for security
+		// Only explicitly approved fields are included to prevent accidental data leakage
+		filteredDetails := make(map[string]interface{})
+		allowedFields := []string{
+			"link",         // Runbook URL
+			"num_firing",   // Number of firing alerts
+			"num_resolved", // Number of resolved alerts
+			"ocm_link",     // OCM console link
+			"region",       // AWS region
+			"dashboard",    // Grafana dashboard link (for monitoring alerts)
```
geowa4: Can Cora do anything with these or is this in preparation of a near term change?

BATMAN-JD: not necessarily I just didn't want to leave any information out initially. I am for sure open to the idea of pruning this down though

pkg/investigations/aiassisted/label_parser_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Parse Prometheus labels
+	if firing, ok := customDetails["firing"].(string); ok && firing != "" {
+		prometheusLabels := parsePrometheusLabels(firing)
+		if len(prometheusLabels) > 0 {
+			payloadData["labels"] = prometheusLabels
+		}
+	}
+
+	// Add filtered custom details (excluding redundant fields)
```
dakotalongRH: The filtering logic in this test uses a deny-list, but didn't we switch to an allow-list in buildInvestigationPayload?  So the test will always pass since the input data doesn't have any fields that would expose the difference since it's testing against its own logic, not the actual function.

For example, if there was a field in PD like "suggested_next_steps" in the custom details, the deny-list would let it through but the allow-list in the actual function would correctly block it.

I see CodeRabbit brought this up for the actual function but it didn't run again to check the test case

BATMAN-JD: You are correct. Good catch. Will update

pkg/investigations/aiassisted/aiassisted.go
```diff
@@ -43,6 +53,132 @@ func generateSessionID(incidentID string) string {
 	return fmt.Sprintf("cad-%s-%d-%s", incidentID, timestamp, randomHex)
 }
 
+// parsePrometheusLabels extracts structured labels from the Prometheus "firing" text blob
```
eth1030: could we move the new filtering functions into a separate label_parser.go file? the file is getting long with the metadata extraction, payload building, and parsing logic all mixed together. Separating parsing concerns would improve readability.

BATMAN-JD: For sure can look into this.

pkg/investigations/aiassisted/alert_parser.go
```diff
@@ -0,0 +1,104 @@
...
+	alerts, err := pdClient.GetAlertsForIncident(incidentID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get alerts for incident: %w", err)
+	}
+
+	if alerts == nil || len(*alerts) == 0 {
+		return nil, fmt.Errorf("no alerts found for incident %s", incidentID)
+	}
+
+	alert := (*alerts)[0]
```
AlexSmithGH: _Technically_ incidents can have N number of alerts associated. We might need some smarter filtering here in the future, but for now I think this is okay.

BATMAN-JD: I am going to look into this. I want to see what something like this could look like, this way if there are a bunch of alerts, at least Cora agent has an idea about all the problems and info.

pkg/investigations/aiassisted/aiassisted.go
```diff
@@ -112,10 +112,16 @@ func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 	incidentID := pdClient.GetIncidentID()
 	alertName := pdClient.GetTitle()
 
-	// Build investigation payload using typed structure
+	alertDetails, firingResult, err := r.PdClient.GetAlertContext(incidentID)
+	if err != nil {
+		logging.Warnf("Failed to extract alert context: %v", err)
+	}
+
+	payloadData := buildInvestigationPayload(firingResult, &alertDetails, pdClient.GetIncidentRef())
```
eth1030, line 120: Potential nil-pointer if GetAlertContext error is partial aiassisted.go:115-120 — When GetAlertContext errors, firingResult is nil and alertDetails is zero-value. Both are handled correctly by buildInvestigationPayload. However, the code continues past the error without any early return or fallback marker in the payload. This is fine for "best effort" but worth calling out — a consumer of investigation_payload won't know that alert context extraction failed entirely vs. the alert just having no firing data. Consider adding a key like "alert_context_error": true to the payload on error, so downstream (Cora) knows context is missing vs. empty.
Could do something like:
  ```
    if err != nil {
          payloadData["alert_context_error"] = err.Error()
    }
  ```

BATMAN-JD:
Found a bug while testing against a live prod alert (ConfigureAlertmanagerOperatorOfflineSRE on Q1HBFZ945EL5DK) — CAO #588 is promoted and firing_json is live, but PagerDuty deserializes JSON bodies into native types. firing_json arrives as []interface{} not a string, so our .(string) type assertion was silently skipping it. Fixed in latest push — handles both types now. Also fixed stringFromDetails for numeric fields like num_firing which PD returns as numbers, not strings.

BATMAN-JD:
Container test exited with code 137, reason OOMKilled will try again later

petrkotas (commented):
Hey @BATMAN-JD thank you for the PR. What is the relevant jira task please? I am curious what is the requirement. I believe this can be done in a way that all other investigations can benefit from it. e.g. doing it in two phases
- Introduce the pagerduty parsing
- Use the parsed data in investigation, one of which is cora

PR #875, opened 2026-07-08
Silence etcdDatabaseQuotaLowSpace alerts after CAD investigation

dakotalongRH:
Most etcdDatabaseQuotaLowSpace alerts resolve with compaction/defrag and don't need SRE attention. Replace Escalate with Silence on success paths.

https://redhat.atlassian.net/browse/ROSAENG-61489

### What type of PR is this?

feature

SRE is being unnecessarily pinged on primary on call for etcdDatabaseQuotaLowSpace alerts. CAD handles the investigation already and creates the cluster report. Once that's done, we can just move it to silent test.

#### Test coverage checks
- [ ] Added tests

- [ ] Ran unit tests locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Warning-level etcd quota alerts are now silenced after the investigation completes.
  * Critical alerts still escalate, with guidance tailored to the critical case.
  * HCP-specific investigations apply the same warning-versus-critical alert handling.
* **Tests**
  * Added unit coverage for severity/title detection and verifies the expected PagerDuty action sequence for both warning and critical paths.

### DIFF
--- a/pkg/investigations/etcddatabasequotalowspace/etcddatabasequotalowspace.go
+++ b/pkg/investigations/etcddatabasequotalowspace/etcddatabasequotalowspace.go
@@ -215,10 +215,12 @@ func (i *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 	result.Actions = append(
 		executor.NoteAndReportFrom(r.Notes, r.Cluster.ID(), i.Name()),
 		backplaneReportAction,
-		resolveOrEscalate(r.PdClient,
-			"etcd warning alert - analysis complete, silencing as compaction/defrag usually resolves this",
-			"etcd critical alert - analysis complete, see report for details"),
 	)
+	if isWarningAlert(r.PdClient) {
+		result.Actions = append(result.Actions, executor.Silence("etcd warning alert - investigation and analysis complete, see report for details"))
+	} else {
+		result.Actions = append(result.Actions, executor.Escalate("etcd critical alert - analysis complete, see report for details"))
+	}
 
 	return result, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/investigations/etcddatabasequotalowspace/etcddatabasequotalowspace.go
```diff
@@ -216,7 +216,7 @@ func (i *Investigation) Run(rb investigation.ResourceBuilder) (investigation.Inv
 	result.Actions = append(
 		executor.NoteAndReportFrom(r.Notes, r.Cluster.ID(), i.Name()),
 		backplaneReportAction, // write a second report here, as this contains the formatted results
-		executor.Escalate("etcd analysis complete - see report for details"),
+		executor.Silence("etcd analysis complete - compaction/defrag usually resolves this, silencing alert"),
```
joshbranham: We don't want to silence all etcd quota alerts, just the `warning` ones.

rolandmkunkel:
The silence message says "compaction/defrag usually resolves this," but CAD doesn't actually perform compaction or defrag. It only takes a snapshot, analyzes it, and reports results. If SRE doesn't get paged for this, who performr the defragmentation/compaction?

pkg/investigations/etcddatabasequotalowspace/etcddatabasequotalowspace.go
```diff
@@ -373,6 +376,24 @@ func (i *Investigation) IsExperimental() bool {
...
+// are still escalated to SRE.
+func isWarningAlert(pdClient pagerduty.Client) bool {
+	if pdClient == nil {
+		return false
+	}
+	return strings.Contains(strings.ToUpper(pdClient.GetTitle()), "WARNING")
+}
+
+// resolveOrEscalate returns a Silence action for warning alerts or an Escalate action for critical alerts.
+func resolveOrEscalate(pdClient pagerduty.Client, silenceReason, escalateReason string) types.Action {
```
AlexSmithGH: total nit you can ignore, but this should probably be named something more accurate to what it's doing. We're not marking anything as resolved, we're just silencing it, but it's still an alert. 

```suggestion
func silenceOrEscalate(pdClient pagerduty.Client, silenceReason, escalateReason string) types.Action {
```

dakotalongRH: Moved it to inline and removed the func

AlexSmithGH (commented):
/lgtm

PR #909, opened 2026-08-13
Append firing alerts to must-gather PD notes

dakotalongRH:
Extract the Alertmanager query logic from clusterhealthcheck into a shared utility package and reuse it in the mustgather investigation. After uploading a must-gather to SFTP, CAD now queries the cluster's Alertmanager for currently firing alerts and appends them to the PagerDuty note, giving SREs immediate context on what was happening at the time of collection.

https://redhat.atlassian.net/browse/ROSAENG-65338

### What type of PR is this?

(feature/bug/documentation/other)

#### Test coverage checks
- [ ] Added tests

- [ ] Ran unit tests locally

## Summary by CodeRabbit

* **New Features**
  * Must-gather reports now include relevant, currently firing Alertmanager alerts in PagerDuty notes.
  * Added support for retrieving and formatting alert details, including severity, state, and summary.

* **Bug Fixes**
  * Informational and metadata-only alerts are excluded from investigation reports.
  * Alert retrieval failures no longer block must-gather completion.

* **Tests**
  * Added coverage for alert parsing, filtering, ordering, and severity handling.

### DIFF
--- a/pkg/aws/mock/aws.go
+++ b/pkg/aws/mock/aws.go
@@ -31,6 +31,7 @@ import (
 type MockEC2API struct {
 	ctrl     *gomock.Controller
 	recorder *MockEC2APIMockRecorder
+	isgomock struct{}
 }
 
 // MockEC2APIMockRecorder is the mock recorder for MockEC2API.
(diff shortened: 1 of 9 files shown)


pkg/aws/mock/aws.go
```diff
@@ -31,7 +31,6 @@ import (
 type MockEC2API struct {
 	ctrl     *gomock.Controller
 	recorder *MockEC2APIMockRecorder
-	isgomock struct{}
```
bergmannf, line 34: Not sure why this PR removes all the `isgomock` blocks.
When I run mockgen locally it doesn't - I'm using:
```
❯ mockgen -version
v0.6.0
```
From https://github.com/uber-go/mock - maybe you are still using the old mock library?

dakotalongRH: Good catch, my local mockgen was on v0.4.0 which stripped the isgomock struct fields added in v0.6.0.  Upgraded and regenerated all mocks with the correct version.

PR #927, opened 2026-08-31
[ROSAENG-61687] Add clusteroperatordownhcp investigation for HCP ClusterOperatorDown

TheUndeadKing:
### What type of PR is this?

feature

Adds a new `clusteroperatordownhcp` investigation that handles the `ClusterOperatorDown` alert on HCP clusters, where a data plane ClusterOperator degrades because a worker node's backing EC2 instance was stopped (a customer action — SRE never stops HCP data plane instances).

Decision logic:
- **Stopped AWSMachine** in the HCP namespace (read from the management cluster): record CloudTrail stop attribution in the PD note, send the "Worker node(s) stopped, action required" service log, and silence.
- **No stopped AWSMachine** (running/none): note + escalate to a human.
- **Non-HCP cluster**: note + escalate to a human — never silenced and never allowed to fall through to the generic AI fallback. The alert is intentionally mapped with no alert-level `when` so it always reaches the investigation; the HCP guard lives inside the investigation.

AWSMachines are read from the HCP namespace on the **management cluster** via `WithManagementK8sClient()`; the CloudTrail lookup uses the HCP's own (customer) AWS account where the data plane instances actually live. `metadata.yaml` grants `get`/`list` on `infrastructure.cluster.x-k8s.io/awsmachines` in the management cluster namespace.

Also migrates all ServiceLog severities across investigations to the new HCC naming (`Major` → `Important`; `Critical` unchanged) for consistency.

#### Test coverage checks
- [X] Added tests

- [X] Ran unit tests locally

## Summary by CodeRabbit

* **New Features**
  * Added HCP-specific handling for `ClusterOperatorDown` alerts.
  * Detects stopped data-plane nodes, records available stop-event details, creates a worker-nodes-stopped service log, and silences the alert when appropriate.
  * Escalates non-HCP clusters and cases where no stopped node is found.

* **Improvements**
  * Added `ClusterOperatorDown` to the example investigation configuration.
  * Updated selected service-log severities from “Major” to “Important” for clearer prioritization.

### DIFF
--- a/docs/investigation-config.example.yaml
+++ b/docs/investigation-config.example.yaml
@@ -174,6 +174,16 @@ alerts:
   #           - operator: sample
   #             values: ["0.10"]
 
+  # Cluster Operator Down (HCP)
+  # No alert-level `when`: the alert must always reach the investigation so a
+  # non-HCP occurrence escalates to a human rather than being filtered into the
+  # AI fallback. The HCP guard lives inside the investigation itself.
+  - alert_title: "ClusterOperatorDown"
+    name: clusteroperatordownhcp
+    investigations:
+      - precheck
+      - clusteroperatordownhcp
+
   # Insights Operator Down
   - alert_title: "InsightsOperatorDown"
     name: insightsoperatordown
--- a/pkg/investigations/clusteroperatordownhcp/clusteroperatordownhcp.go
+++ b/pkg/investigations/clusteroperatordownhcp/clusteroperatordownhcp.go
@@ -1,0 +1,193 @@
+// Package clusteroperatordownhcp investigates the ClusterOperatorDown alert for
+// HCP clusters, where a data plane ClusterOperator is degraded because the
+// worker node's backing EC2 instance was stopped.
+package clusteroperatordownhcp
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	ec2v2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/openshift/configuration-anomaly-detection/pkg/aws"
+	"github.com/openshift/configuration-anomaly-detection/pkg/executor"
+	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/investigation"
+	"github.com/openshift/configuration-anomaly-detection/pkg/ocm"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Investigation struct{}
+
+func (c *Investigation) Name() string {
+	return "clusteroperatordownhcp"
+}
+
+// awsMachineStopped is the value of an AWSMachine's status.instanceState when
+// its backing EC2 instance has been stopped.
+const awsMachineStopped = "stopped"
+
+func (c *Investigation) Run(rb investigation.ResourceBuilder) (investigation.InvestigationResult, error) {
+	result := investigation.InvestigationResult{}
+	ctx := context.Background()
+
+	// WithAwsClient authenticates to the HCP's own (customer) AWS account, which
+	// is where the data plane EC2 instances live — not the management cluster's.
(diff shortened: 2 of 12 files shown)


docs/investigation-config.example.yaml
```diff
@@ -174,6 +174,16 @@ alerts:
   #           - operator: sample
   #             values: ["0.10"]
 
+  # Cluster Operator Down (HCP)
+  # No alert-level `when`: the alert must always reach the investigation so a
+  # non-HCP occurrence escalates to a human rather than being filtered into the
+  # AI fallback. The HCP guard lives inside the investigation itself.
+  - alert_title: "ClusterOperatorDown"
```
rolandmkunkel, line 181: There needs to be a PR to app-interface to add this to the staging and production configs

pkg/investigations/clusteroperatordownhcp/clusteroperatordownhcp.go
```diff
@@ -0,0 +1,193 @@
...
+
+	events, err := awsCli.PollInstanceStopEventsFor([]ec2v2types.Instance{inst}, 5)
+	if err != nil || len(events) == 0 {
+		return "CloudTrail stop event not found (instance may have been stopped more than 2h ago)."
+	}
+
+	stopEvent := events[0]
+	stoppedBy := "unknown"
+	if stopEvent.Username != nil {
+		stoppedBy = *stopEvent.Username
```
rolandmkunkel, line 173: The username is PII, I would rather use stopEvent.AccessKeyId instead, which lets SREs trace the identity without embedding PII directly to be more defensive with that

pkg/investigations/clusteroperatordownhcp/clusteroperatordownhcp.go
```diff
@@ -0,0 +1,193 @@
...
+		r.Notes.AppendWarning("No stopped AWSMachine found in %s - the degraded operator is not caused by a stopped worker instance", r.HCPNamespace)
+		result.Actions = append(
+			executor.NoteAndReportFrom(r.Notes, r.Cluster.ID(), c.Name()),
+			executor.Escalate("No stopped worker instance found - manual investigation required"),
+		)
+		return result, nil
+	}
+
+	// On HCP, SRE never stops data plane instances, so a stopped instance is a
+	// customer action by definition. CloudTrail attribution is recorded in the
```
rolandmkunkel, line 73: Is that assumption confirmed? Are there no possible edge cases such as when instances are replaced or clusters are transitioning between states? We do send a service log based on that without SRE interaction

docs/investigation-config.example.yaml
```diff
@@ -174,6 +174,16 @@ alerts:
   #           - operator: sample
   #             values: ["0.10"]
 
+  # Cluster Operator Down (HCP)
```
rolandmkunkel, line 177: The new investigation should also be added to the list on the top

pkg/investigations/cpd/cpd.go
```diff
@@ -23,7 +23,7 @@ func newBYOVPCRoutingSL(docLink string) *ocm.ServiceLog {
 	}
 
 	return &ocm.ServiceLog{
-		Severity:     "Major",
+		Severity:     "Important",
```
rolandmkunkel, line 26: These don't seem to be directly related to the ticket. Is that a breaking or cosmetic change?

pkg/investigations/clusteroperatordownhcp/clusteroperatordownhcp.go
```diff
@@ -0,0 +1,193 @@
...
+	// is where the data plane EC2 instances live — not the management cluster's.
+	r, err := rb.WithCluster().WithManagementK8sClient().WithAwsClient().WithNotes().Build()
+	if err != nil {
+		return result, err
+	}
+
+	// This alert is only actionable on HCP. On classic clusters the AWSMachine /
+	// management-cluster model below does not exist, so hand it to a human rather
+	// than letting it fall through to the generic AI fallback.
+	if !r.IsHCP {
```
rolandmkunkel, line 45: We should check if its an HCP cluster before building the AWS client, otherwise this will incur costs for a client we don't need

PR #938, opened 2026-09-08
ROSAENG-66516: fix CAD double-escalating PagerDuty incidents on AI fallback

BATMAN-JD:
## Summary

CAD could escalate the same PagerDuty incident twice when it took the AI fallback path: `aiassisted` always escalates the incident itself whenever it actually runs, but `PagerDutyController.Investigate` also issued a second, generic escalation unconditionally afterward.

Rather than special-case the one reported path, this centralizes escalation de-duplication at the source:

- `trackingPDClient` (`pkg/controller/escalation_tracker.go`) wraps the real `pagerduty.Client` and ensures `EscalateIncident`/`EscalateIncidentWithNote` only ever reach the PagerDuty API once per incident, no matter which caller gets there first (an investigation's direct call, an investigation's action via the executor, or the controller's own fallback/failure-handling logic). A single instance is shared across all of those paths.
- `EscalateIncidentWithNote` degrades to a plain note (instead of no-op) once already escalated, so note content isn't lost.
- This let several per-call-site double-escalation guards (`handleCADFailure`, `escalateDocumentationMismatch`, `Investigate`'s fallback) be simplified to unconditional calls, trusting the client-level guard.

## Bugs found and fixed along the way

- `aiassisted.go` type-asserted its PagerDuty client to the concrete `*pagerduty.SdkClient` type. Since `trackingPDClient` is a different concrete type wrapping that interface, this assertion would always fail once wired up — silently disabling the entire AI-assisted investigation in production. Fixed by widening the `pagerduty.Client` interface (`GetIncidentID`, plus `GetIncidentRef`/`GetServiceName` needed to make `PagerDutyController` mockable at all) and switching to a nil check + direct interface calls.
- Replaced an initial `atomic.Bool`/CompareAndSwap-based tracker with a plain `sync.Mutex`, removing a cross-package assumption (that PagerDuty actions never execute concurrently) that the CAS approach silently depended on.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `golangci-lint run` all clean
- [x] `go test ./...` passes (including `-race` on the affected packages)
- [x] New regression tests:
  - `pkg/controller/escalation_tracker_test.go` — de-duplication behavior of `trackingPDClient`, including the note-degradation and failed-then-retried-escalation cases
  - `pkg/controller/pagerduty_test.go` — `Investigate()`-level tests asserting `EscalateIncident` fires exactly once
  - `pkg/investigations/aiassisted/aiassisted_test.go` — regression test for the `PdClient` interface-vs-concrete-type bug above

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented PagerDuty incidents from being escalated more than once, even when multiple investigation or fallback paths trigger escalation.
  - Preserved context from repeated escalation requests by adding subsequent notes to the incident instead of issuing another escalation.
  - Improved escalation handling when no matching alert or configuration is available.
- **Reliability**
  - PagerDuty integrations now work consistently across supported client implementations without relying on a specific client type.

### DIFF
--- a/pkg/controller/pagerduty.go
+++ b/pkg/controller/pagerduty.go
@@ -83,10 +83,9 @@ func (c *PagerDutyController) Investigate(ctx context.Context) error {
 
 	// Nothing above escalated this incident yet (e.g. the AI-fallback chain
 	// was skipped entirely, or precheck/aiassisted didn't escalate on their
-	// own): issue exactly one generic escalation now.
-	if c.notifier.HasEscalated() {
-		return nil
-	}
+	// own): issue a generic escalation now. If something upstream already
+	// escalated, notifier.Escalate() is a safe no-op (trackingPDClient
+	// de-duplicates at the source) rather than a real second escalation.
 	if escErr := c.notifier.Escalate(); escErr != nil {
 		return fmt.Errorf("could not escalate unsupported alert: %w", escErr)
 	}
@@ -101,6 +100,8 @@ func escalateDocumentationMismatch(docErr *ocm.DocumentationMismatchError, resou
 		message = resources.Notes.String()
 	}
 
+	// If this incident was already escalated (e.g. by aiassisted), trackingPDClient
+	// degrades this to a plain note instead of a real second escalation.
 	if err := notifier.EscalateWithNote(message); err != nil {
 		logging.Errorf("Failed to escalate documentation mismatch notes to PagerDuty: %v", err)
 		return
--- a/pkg/controller/pagerduty.go
+++ b/pkg/controller/pagerduty.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 
 	"github.com/openshift/configuration-anomaly-detection/pkg/config"
+	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/aiassisted"
 	"github.com/openshift/configuration-anomaly-detection/pkg/investigations/investigation"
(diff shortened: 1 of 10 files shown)


pkg/controller/pagerduty.go
```diff
@@ -80,7 +81,12 @@ func (c *PagerDutyController) Investigate(ctx context.Context) error {
 		}
 	}
 
-	if escErr := c.pdClient.EscalateIncident(); escErr != nil {
+	// Nothing above escalated this incident yet (e.g. the AI-fallback chain
+	// was skipped entirely, or precheck/aiassisted didn't escalate on their
+	// own): issue a generic escalation now. If something upstream already
+	// escalated, notifier.Escalate() is a safe no-op (trackingPDClient
+	// de-duplicates at the source) rather than a real second escalation.
+	if escErr := c.notifier.Escalate(); escErr != nil {
```
gvnnn, line 89: Looks good to me, other than I believe at this point we could have a silenced alert:

precheck (chained before aiassisted) finds an uninstalling cluster -> silences the alert -> the alert gets escalated again. Perhaps the wrapper could dedupe silences as well? This is preexisting anyways, so it does not necessarily have to be addressed as part of this PR

eth1030: +1 all else lgtm, can lgtm if you want to address this in another PR

